PR #3, opened 2019-01-16
Windows 32bit build

mjp41:
This PR fixes multiple issues with making 32bit compilation work on Windows. 

It adds 32 bit Windows builds to the CI. 

It does not address 32bit for other platforms.

### DIFF
--- a/src/mem/alloc.h
+++ b/src/mem/alloc.h
@@ -521,9 +521,7 @@ namespace snmalloc
         this->size += sizeclass_to_size(sizeclass);
 
         Remote* r = (Remote*)p;
-        r->set_sizeclass(sizeclass);
-        assert(r->sizeclass() == sizeclass);
-        r->set_target_id(target_id);
+        r->set_sizeclass_and_target_id(target_id, sizeclass);
         assert(r->sizeclass() == sizeclass);
         assert(r->target_id() == target_id);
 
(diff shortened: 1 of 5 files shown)


src/mem/alloc.h
```diff
@@ -38,7 +38,10 @@ namespace snmalloc
     PMMediumslab = 2
   };
 
-  using SuperslabPagemap = Pagemap<SUPERSLAB_BITS, uint8_t, 0>;
+  using SuperslabPagemap = std::conditional_t<
+    bits::is64(),
```
davidchisnall: It would be nice to have a new interface here to get the valid address space size, rather than `8*sizeof(void*)`.  That would let us handle cases like SGX, where we may want to use a 32-bit address range with a 64-bit ABI, or CHERI.  Factoring this out early means that we won't need to do a load of cleanup later.

src/mem/pagemap.h
```diff
@@ -224,4 +228,44 @@ namespace snmalloc
       } while (length > 0);
     }
   };
+
+  template<size_t GRANULARITY_BITS, typename T>
+  class FlatPagemap
```
davidchisnall: Missing doc comment.

src/mem/alloc.h
mjp41: So we use https://github.com/Microsoft/snmalloc/blob/master/src/ds/bits.h#L76 to get the addressable range, which for SGX we could set to 32bit, with 64bit pointers.  This conditional_t is not ideal, but I don't know what features we want to enable here.

We may also want to use the FlatPagemap even for 64bit with 48bit address space as it is just 16MiB.  Perhaps the right approach is to set a max FlatPagemap size, and then if the flatpagemap is under that size, we use it, and if not we use the tree.

PR #4, opened 2019-01-17
Merge changes required for using snmalloc in FreeBSD libc

davidchisnall:
With these changes, 32- and 64-bit userland builds, static and dynamically linked 64-bit versions are tested (32-bit ones are not, but hopefully work!).

### DIFF
--- a/src/override/malloc.cc
+++ b/src/override/malloc.cc
@@ -4,18 +4,62 @@
 
 using namespace snmalloc;
 
+#ifndef SNMALLOC_EXPORT
+#  define SNMALLOC_EXPORT
+#endif
+
 #ifndef SNMALLOC_NAME_MANGLE
 #  define SNMALLOC_NAME_MANGLE(a) a
 #endif
 
+namespace
+{
+  /**
+   * RAII wrapper around an `Alloc`.  This class gets an allocator from the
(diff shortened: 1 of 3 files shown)


src/override/malloc.cc
```diff
@@ -4,34 +4,78 @@
...
 
+namespace
+{
+  /**
+   * RAII wrapper around an `Alloc`.  This class gets an allocator from the
+   * global pool and wraps it so that `Alloc` methods can be called
+   * directly via the `->` operator on this class.  When this object is
+   * destroyed, it returns the allocator to the global pool.
+   */
+  struct slow_allocator
```
mjp41: Do you need to explicitly delete the copy constructor?

src/override/malloc.cc
```diff
@@ -4,34 +4,78 @@
 
 using namespace snmalloc;
 
+#ifndef SNMALLOC_EXPORT
+#  define SNMALLOC_EXPORT
```
mjp41: Should we make this `extern "C"` by default, and remove the `extern "C"` block?

src/override/malloc.cc
mjp41: I wonder if this class should be in a header file as it seems useful in more scenarios.

PR #7, opened 2019-01-18
Fixes remote and pagemap

mjp41:
This PR makes the pagemap have OS_PAGE_SIZE alignment, and fixes the surrounding API.

Additionally, it fixes the encoding of sizeclasses into the top bits of remote allocation, to use the bottom bits, and shift the pointers up.  This allows a right signed shift to preserve the sign extended address space.

### DIFF
--- a/src/mem/largealloc.h
+++ b/src/mem/largealloc.h
@@ -68,7 +68,8 @@ namespace snmalloc
      * Primitive allocator for structure that are required before
      * the allocator can be running.
      ***/
-    void* alloc_chunk(size_t size, size_t alignment = 64)
+    template<size_t alignment = 64>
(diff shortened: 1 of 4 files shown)


src/mem/largealloc.h
```diff
@@ -61,7 +68,7 @@ namespace snmalloc
      * Primitive allocator for structure that are required before
      * the allocator can be running.
      ***/
-    void* alloc_chunk(size_t size)
+    void* alloc_chunk(size_t size, size_t alignment = 64)
```
davidchisnall: Do we ever not know the alignment statically?  If not, then we might want to make alignment a template parameter.

mjp41: Good point. Will fix.

PR #10, opened 2019-01-24
Hardening allocator

mjp41:
General hardening of parts of the allocator:
* Prevent memory corruption in various places becoming a further gadget.
* Ensure we only access a valid meta slab index
* Remove some use of raw pointers to metaslabs.

### DIFF
--- a/src/mem/metaslab.h
+++ b/src/mem/metaslab.h
@@ -92,12 +92,9 @@ namespace snmalloc
       head = (uint16_t)~0;
     }
 
-    void zero()
+    void init()
     {
-      head = 0;
       used = 0;
-      link = 0;
-      sizeclass = 0;
       next = 0;
     }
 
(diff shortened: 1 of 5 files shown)


src/mem/metaslab.h
```diff
@@ -91,6 +92,15 @@ namespace snmalloc
...
     }
 
+    void zero()
+    {
+      head = 0;
+      used = 0;
+      link = 0;
+      sizeclass = 0;
+      next = 0;
+    }
```
davidchisnall: It would be nice to have a `static_assert` on `sizeof(metaslab)` here so that if someone adds a field they don't silently break this.

src/mem/metaslab.h
```diff
@@ -92,12 +92,9 @@ namespace snmalloc
       head = (uint16_t)~0;
     }
 
-    void zero()
+    void init()
```
davidchisnall: Can we have a comment explaining why this isn't the constructor?

PR #12, opened 2019-02-02
Avoid unnecessary allocation from realloc

Theodus:


Theodus:
Also, quick question: Should `calloc` ever result in a call to `madvise(... MADV_DONTNEED)` on the region being allocated? This seems wrong to me.

davidchisnall:
I'm not 100% convinced by this change.  `realloc` is often used in a pattern where you allocate something bigger than you'll need, figure out how much you actually need, and then `realloc` to shrink the allocation.  With `malloc` implementations that subdivide allocations, this lets the allocator reuse the rest of the allocation.  It's not clear to me what the correct behaviour should be with snmalloc: we'll waste memory if we don't copy, but we'll also incur extra overhead.

On the `calloc` / `madvise` question, it might make sense on Linux.  Linux's `MADV_DONTNEED` is misnamed and actually removes the physical pages and replaces them with CoW mappings of a zero page.  Whether this hurts performance more or less than calling `memset` is not clear.  We'll only hit this code path on allocations that are greater than 4KB, so we'll end up blowing away a chunk of cache if we `memset`.  On the other hand, it's a system call that acquires a bunch of VM subsystem locks (probably does some IPIs), then the same thing on page fault a few instructions later when the allocation is touched.  We should possibly consider always `memset`ing the first page and only `madvise`ing later ones.

mjp41:
We could only copy if the size would lead to a different sizeclass?

PR #13, opened 2019-02-05
Simplification around remote allocation

mjp41:
This removes storing the sizeclass in the remote allocation.  On the most common paths, we touch the cache lines that contain the size class, so there is no noticeable perf improve from storing this in the block.

### DIFF
--- a/src/mem/alloc.h
+++ b/src/mem/alloc.h
@@ -522,6 +522,15 @@ namespace snmalloc
       size_t size = 0;
       RemoteList list[REMOTE_SLOTS];
 
+      inline size_t get_slot(size_t id, size_t r)
+      {
+        constexpr size_t allocator_size =
+          sizeof(Allocator<MemoryProvider, PageMap, IsQueueInline>);
+        constexpr size_t initial_shift =
+          bits::next_pow2_bits_const(allocator_size);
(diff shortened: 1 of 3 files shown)


mjp41:
CC @plietar

src/mem/alloc.h
```diff
@@ -522,16 +522,24 @@ namespace snmalloc
       size_t size = 0;
       RemoteList list[REMOTE_SLOTS];
 
+      inline size_t get_slot(size_t id, size_t r)
```
davidchisnall: Please add a doc comment explaining what a slot is.

src/mem/remoteallocator.h
```diff
@@ -10,80 +10,26 @@ namespace snmalloc
...
-      {
-        return ((value & 1) == 1) ? 0 : possible_sizeclass;
-      }
+      return allocator_id;
     }
   };
 
-  static_assert(
-    (offsetof(Remote, possible_sizeclass)) <= MIN_ALLOC_SIZE,
-    "Need to be able to cast any small alloc to Remote");
```
plietar: Keep this assertion, but use `sizeof(Remote)` rather than `offsetof(...)`

PR #19, opened 2019-02-12
Make internal symbols hidden.

plietar:
Only the malloc/free/... interface is exported.

This allows all of C++'s weak ODR symbols to be resolved at
compile-time, rather than staying weak and resolved at load-time.

Since internal calls don't need to go through the PLT anymore, we get a
nice speedup, at least on small objects.

image


plietar:
Was hoping closing the old one and making a new one would make the CLA bot happy. Didn't work.

PR #20, opened 2019-02-13
Add malloc tests

Theodus:
Added some tests for the malloc shim


src/override/malloc.cc
```diff
@@ -40,10 +40,10 @@ extern "C"
   {
     bool overflow = false;
     size_t sz = bits::umul(size, nmemb, overflow);
-    if (overflow)
+    if (overflow || ((nmemb | size) == (size_t)-1))
```
mjp41: I guess you are required to fail sensibly, if you are given negative numbers?  Or just -1.  Does this cover all the cases that are specified?

src/override/malloc.cc
```diff
@@ -135,7 +135,12 @@ extern "C"
       return nullptr;
     }
 
-    uint8_t sc = size_to_sizeclass((std::max)(size, alignment));
+    size = (std::max)(size, alignment);
+    if (size >= SUPERSLAB_SIZE)
+    {
```
mjp41: Why is this necessary?  Doesn't the subsequent test sc>=NUM_SIZECLASSES cover this?  Is this an optimisation, or am I missing something in the logic.

src/test/func/malloc/malloc.cc
```diff
@@ -0,0 +1,127 @@
...
+  while (__builtin_popcount(n) < 2)
+    n = r.next();
+
+  return n;
+}
+
+void check_result(size_t size, void* p, int err, bool null)
+{
+  fprintf(stderr, "%p, errno: %s\n", p, strerror(errno)); // TODO
+  assert(errno == err);
```
mjp41: As you are using assert, this will not be tested in a release build.  Is that what you want?  Should we make this if (!b) abort().  So we can test the release build too?

mjp41: Or redefine assert

src/test/func/malloc/malloc.cc
```diff
@@ -0,0 +1,127 @@
...
+  auto n = (r.next() % (max + 1)) + min;
+  while (__builtin_popcount(n) < 2)
+    n = r.next();
+
+  return n;
+}
+
+void check_result(size_t size, void* p, int err, bool null)
+{
+  fprintf(stderr, "%p, errno: %s\n", p, strerror(errno)); // TODO
```
mjp41: TODO?

src/test/func/malloc/malloc.cc
```diff
@@ -0,0 +1,127 @@
...
+  constexpr int SUCCESS = 0;
+
+  auto r = xoroshiro::p128r64((size_t)time(nullptr));
+
+  size_t n, s;
+  do
+  {
+    n = rand_pow_2(r);
+    s = rand_pow_2(r);
+  } while ((n * s) > (1 << max_pow_2));
```
mjp41: Should this loop be around the next command too?

mjp41: Also the subsequent tests that use r, do you want to do them more than once, so we get some coverage of the values?

src/test/func/malloc/malloc.cc
```diff
@@ -0,0 +1,127 @@
...
+}
+
+int main(int argc, char** argv)
+{
+  UNUSED(argc);
+  UNUSED(argv);
+
+  constexpr int SUCCESS = 0;
+
+  auto r = xoroshiro::p128r64((size_t)time(nullptr));
```
mjp41: I don't like the test being unpredictable.  This may or may not fail in CI.  This is bad for reproducibility.

I would sooner you take a seed from the command line, and run multiple instances of the test with different seeds.

If you really think having some "real" randomness is necessary, please print the seed, and allow it to be taken from the command line, so we can reproduce failures.

src/override/malloc.cc
Theodus: glibc sets ENOMEM if either nmemb or size is -1. I assume that it would be unreasonable for an application to rely on this behavior for other values, but maybe that could also be said for -1. Is there a more reasonable upper bound that we could put on the size of allocations?

src/override/malloc.cc
Theodus: It wasn't. We just need to set size as the max of size and alignment for the following allocation

src/test/func/malloc/malloc.cc
Theodus: sorry

src/test/func/malloc/malloc.cc
Theodus: The randomness wasn't necessary. The tests now run through all size classes and power of 2 alignments that may have been chosen randomly

CMakeLists.txt
```diff
@@ -74,11 +74,18 @@ subdirlist(TEST_CATEGORIES ${TESTDIR})
 foreach(SUPER_SLAB_SIZE 1;16)
   foreach(TEST_CATEGORY ${TEST_CATEGORIES})
     subdirlist(TESTS ${TESTDIR}/${TEST_CATEGORY})
+    if(MSVC)
```
mjp41: Is this required?

src/override/malloc.cc
mjp41: It seems unreasonable to even rely on it for -1 to me.  I am just concerned about putting tests on a critical path.

CMakeLists.txt
mjp41: If you include malloc.cc in the test, does it do the right thing?

mjp41: From CI it looks like it doesn't.  Leave the exception, and I can fix this later.

Theodus: The CI says no

Theodus: Should we be providing symbols like `_msize` and `_aligned_malloc`? If so, do we support all of these: https://docs.microsoft.com/en-us/cpp/c-runtime-library/data-alignment?view=vs-2017?

Theodus: ok, I'll leave this for later

Theodus:
Should largealloc be calling `notify_using` with `NoZero` here, since the comment above says the rest of the pages don't need to be zeroed?
https://github.com/Microsoft/snmalloc/blob/master/src/mem/largealloc.h#L198

mjp41:
> Should largealloc be calling notify_using with NoZero here, since the comment above says the rest of the pages don't need to be zeroed?
> https://github.com/Microsoft/snmalloc/blob/master/src/mem/largealloc.h#L198

So I think the comment hasn't been correctly updated when I introduced the PAL.  

The first page has been committed the whole time.  So definitely needs zeroing.  The rest may or may not need zeroing depending on the platform.  If we have madvised free, in the not-using, you are not guaranteed the pages are zero when they come back, as they may still contain the previous objects data, or zeros.  So this should call madvise dontneed.

If not_using does nothing, then using will have to memset.  For systems where you don't have access to a decommit like feature.

Theodus:
The assertion failure occurs with every call to `LargeAlloc::alloc<ZeroMem=YesZero>` with sizes in the range [917520, 1048560] (step size of 16), excluding every multiple of 96.

src/test/func/malloc/malloc.cc
```diff
@@ -0,0 +1,118 @@
+#include <cerrno>
+#include <malloc.h>
```
davidchisnall: This file has been deprecated for about 20 years and causes a hard error on FreeBSD, it probably should be stdlib.h.

src/test/func/malloc/malloc.cc
```diff
@@ -0,0 +1,118 @@
...
+    abort();
+
+  if (null)
+  {
+    if (p != nullptr)
+      abort();
+  }
+  else
+  {
+    if (malloc_usable_size(p) < size)
```
davidchisnall: This is a non-standard function, so we can't expect it to be in any headers.  It isn't on FreeBSD or macOS.

src/test/func/malloc/malloc.cc
```diff
@@ -0,0 +1,118 @@
...
+  void* p = nullptr;
+  errno = posix_memalign(&p, align, size);
+  check_result(size, align, p, err, null);
+}
+
+void test_memalign(size_t size, size_t align, int err, bool null)
+{
+  fprintf(stderr, "memalign(%d, %d)\n", (int)align, (int)size);
+  errno = 0;
+  void* p = memalign(align, size);
```
davidchisnall: memalign is deprecated in glibc and doesn't exist anywhere else, so we probably need to explicitly declare it if we want to test our version.

PR #21, opened 2019-02-13
Place the next pointer at a different place on every object.

plietar:
This limits the collision in cache buckets, especially for objects in the 512-8K range.

capture

This is behind a flag (CACHE_FRIENDLY_OFFSET) for the moment.


CMakeLists.txt
```diff
@@ -46,6 +48,10 @@ if(USE_SNMALLOC_STATS)
   add_definitions(-DUSE_SNMALLOC_STATS)
 endif()
 
+if(CACHE_FRIENDLY_OFFSET)
```
mjp41: Given the tests are setting this manually.  Should this just be added to the `add_shim` macro?

src/mem/sizeclass.h
```diff
@@ -6,6 +6,9 @@ namespace snmalloc
 {
   constexpr static uint16_t get_slab_offset(uint8_t sc, bool is_short);
   constexpr static size_t sizeclass_to_size(uint8_t sizeclass);
+  constexpr static size_t sizeclass_to_cache_friendly_mask(uint8_t sizeclass);
+  constexpr static size_t
+  sizeclass_to_inverse_cache_friendly_mask(uint8_t sizeclass);
```
mjp41: Is that really what clangformat gives here?

CMakeLists.txt
plietar: Done

src/mem/sizeclass.h
plietar: Done

PR #22, opened 2019-02-14
Pal zero bug alignment

mjp41:
This fixes a bug found by PR #20.

The PAL requires the values to be page aligned, but in a large alloc case they weren't.


mjp41:
@Theodus this should fix at least one bug you found.

PR #23, opened 2019-02-15
Fix comment to represent current PAL API.

mjp41:


src/mem/largealloc.h
```diff
@@ -191,10 +191,13 @@ namespace snmalloc
       {
         if ((decommit_strategy != DecommitNone) || (large_class > 0))
         {
-          // Only the first page needs to be zeroed, as this was decommitted.
+          // The first page is already in "use" for the stack element,
+          // this will need zeroing for a YesZero call.
           if (zero_mem == YesZero)
```
Theodus: Should we be marking these `zero_mem` checks as constexpr?

src/mem/largealloc.h
```diff
@@ -193,7 +193,7 @@ namespace snmalloc
         {
           // The first page is already in "use" for the stack element,
           // this will need zeroing for a YesZero call.
-          if (zero_mem == YesZero)
+          if constexpr (zero_mem == YesZero)
```
Theodus: We should probably do the same for the `if (allow_reserve == YesReserve)` check above (line 141)

PR #24, opened 2019-02-15
Onepastend

davidchisnall:
Introduce a `OnePastEnd` option for the pointer immediately after the
end of the allocation.  This simplifies some of the logic in callers,
where they wants to say 'is base + length safe to use?'.

Also restructure some of the other logic somewhat.


davidchisnall:
I'll rebase once #25 is in.

PR #25, opened 2019-02-15
Made the malloc tests run on Windows.

mjp41:


davidchisnall:
Works on FreeBSD, seems to not work on Windows...

PR #29, opened 2019-02-20
Initial cut at a lazy decommit strategy.

davidchisnall:
This does not deallocate memory until the OS tells us that we are short
on memory, then tries to decommit all of the cached chunks (except for
the first page, used for the linked lists).

Nowhere near enough testing to commit to master yet!

PR submitted to push it through CI and see what I broke.

### DIFF
--- a/src/mem/largealloc.h
+++ b/src/mem/largealloc.h
@@ -116,40 +116,6 @@ namespace snmalloc
       lazy_decommit_guard.clear();
     }
 
-  private:
-    /**
-     * Query whether the PAL given by `P` supports a specific feature.  This is
-     * used internally in templated functions where PAL can't be referenced in
-     * an `enable_if` context.
-     */
-    template<typename P, PalFeatures F>
-    constexpr static bool pal_supports()
-    {
-      return (P::pal_features & F) == F;
-    }
-    /**
-     * Wrapper that is instantiated only if the memory provider supports low
-     * memory notifications and forwards the call to the memory provider.
-     */
-    template<typename M>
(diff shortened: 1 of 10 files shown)


src/pal/pal_windows.h
```diff
@@ -19,7 +19,76 @@ namespace snmalloc
...
+     */
+    uint64_t low_memory_epoch()
+    {
+      return pressure_epoch.load(std::memory_order_acquire);
+    }
+    /**
+     * Check whether the low memory state is still in effect.  This is an
+     * expensive operation and should not be on any fast paths.
+     */
+    bool expensive_low_memory_check()
```
mjp41: This is not used?

src/pal/pal_windows.h
```diff
@@ -127,5 +196,8 @@ namespace snmalloc
 #  endif
     }
   };
+  HEADER_GLOBAL std::atomic<uint64_t> PALWindows::pressure_epoch;
```
mjp41: What does the code gen do for this?  Hopefully doesn't do thread safe initialisation.

src/mem/largealloc.h
```diff
@@ -103,12 +160,71 @@ namespace snmalloc
...
+      static std::atomic<uint64_t> counter;
+      auto c = counter++;
+      if (c % TEST_LAZY_DECOMMIT == 0)
+      {
+        lazy_decommit();
+      }
+#else
+      if constexpr (decommit_strategy == DecommitSuperLazy)
+      {
+        auto new_epoch = low_memory_epoch();
```
mjp41: Is the order of these two reads here important.

PR #33, opened 2019-04-09
Adding support for shared allocators

rschust:
Exporting the global pagemap and the default memory provider's reserve function in order to support shared allocators

### DIFF
--- a/src/mem/allocconfig.h
+++ b/src/mem/allocconfig.h
@@ -4,6 +4,34 @@
 
 namespace snmalloc
 {
+  void error(const char* const str);
+
+  /**
+   * Flags in a bitfield of optional features that a PAL may support.  These
+   * should be set in the PAL's `pal_features` static constexpr field.
+   */
+  enum PalFeatures : uint64_t
+  {
+    /**
+     * This PAL supports low memory notifications.  It must implement a
+     * `low_memory_epoch` method that returns a `uint64_t` of the number of
+     * times that a low-memory notification has been raised and an
+     * `expensive_low_memory_check()` method that returns a `bool` indicating
+     * whether low memory conditions are still in effect.
+     */
+    LowMemoryNotification = (1 << 0),
+    /**
+     * This PAL natively supports allocation with a guaranteed alignment.  If
+     * this is not supported, then we will over-allocate and round the
+     * allocation.
+     *
+     * A PAL that does supports this must expose a `request()` method that takes
+     * a size and alignment.  A PAL that does *not* support it must expose a
+     * `request()` method that takes only a size.
+     */
+    AlignedAllocation = (1 << 1)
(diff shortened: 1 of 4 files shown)


mjp41:
@davidchisnall defining the PalFeatures outside the Pal.h means we can define a Pal externally to snmalloc, by including allocconfig.h.  The current master makes this impossible.

src/override/malloc.cc
```diff
@@ -205,6 +205,17 @@ extern "C"
     return ENOENT;
   }
 
+  SNMALLOC_EXPORT void* SNMALLOC_NAME_MANGLE(get_global_pagemap)(void)
```
mjp41: I wonder if these should be conditionally compiled?  `#ifdef` As we don't need these symbols in most scenarios.
 E.g.
```C
#ifdef SNMALLOC_EXPOSE_PAGEMAP
```
and
```C
#ifdef SNMALLOC_EXPOSE_RESERVE
```

src/mem/allocconfig.h
```diff
@@ -4,6 +4,34 @@
 
 namespace snmalloc
 {
+  void error(const char* const str);
```
mjp41: Not sure why this is here?  Is this actually used? @rschust I know this was there before, but not sure we want this prototype anywhere now?

src/mem/allocconfig.h
```diff
@@ -4,6 +4,34 @@
 
 namespace snmalloc
```
mjp41: I wonder if we should perhaps have a pal_pre_include.h that contains the enums required to define a pal and nothing else.

src/mem/allocconfig.h
rschust: It is being used by alloc.h, globalalloc.h and malloc.cc, but the forward declaration should (and will) be on pal.h.

src/mem/allocconfig.h
rschust: Shall I create pal_consts.h?

src/mem/allocconfig.h
mjp41: Okay. Thanks

src/mem/allocconfig.h
mjp41: That sounds good. Thanks

src/override/malloc.cc
davidchisnall: I would also like to see this one prefixed with snmalloc.  We probably do want to support this in the libc implementation.  I had a FIXME somewhere to expose something like this, but taking the version (or possibly some encoding of the configuration) so that we can return null if there's a mismatch between the pagemap representation that the caller expects and the callee provides.  Since we're carefully sidestepping type safety here, we need to reintroduce it somehow.

PR #34, opened 2019-04-10
Pagemap access refactor

davidchisnall:
Now that we have a function for finding the pagemap that we can use across shared libraries, modify it to do some type checking and make some of the other code more resilient to importing a pagemap from a library without pulling in the symbol.

### DIFF
--- a/src/mem/alloc.h
+++ b/src/mem/alloc.h
@@ -101,7 +101,7 @@ namespace snmalloc
 
   public:
     /**
-     * Construtor.  Accesses the pagemap via the C ABI accessor and casts it to
+     * Constructor.  Accesses the pagemap via the C ABI accessor and casts it to
      * the expected type, failing in cases of ABI mismatch.
      */
     ExternalGlobalPagemap()
--- a/src/mem/alloc.h
+++ b/src/mem/alloc.h
@@ -70,7 +70,15 @@ namespace snmalloc
    */
   struct SuperslabMap
   {
+    /**
+     * Reference to the pagemap that we're using, allowing it to not be the
+     * default pagemap.
+     */
     SuperslabPagemap& pagemap;
+    /**
+     * Constructor.  Uses the global pagemap by default, but can use a
+     * different one if required.
+     */
     SuperslabMap(SuperslabPagemap& pm = global_pagemap) : pagemap(pm) {}
     /**
      * Get the pagemap entry corresponding to a specific address.
--- a/src/mem/alloc.h
+++ b/src/mem/alloc.h
@@ -138,7 +138,7 @@ namespace snmalloc
     {
       const snmalloc::PagemapConfig* c;
       external_pagemap =
-        SuperslabPagemap::cast_to_pagemap(snmalloc_get_global_pagemap(&c), c);
+        SuperslabPagemap::cast_to_pagemap(snmalloc_pagemap_global_get(&c), c);
       // FIXME: Report an error somehow in non-debug builds.
       assert(external_pagemap);
     }
--- a/src/mem/alloc.h
+++ b/src/mem/alloc.h
@@ -80,36 +80,6 @@ namespace snmalloc
   };
 
   /**
-   * Mixin used by `SuperslabMap` to access a pagemap provided by the user.
-   * This should be used by code that needs to access the canonical pagemap via
-   * some custom mechanism.
-   */
-  class CustomPagemap
-  {
-    /**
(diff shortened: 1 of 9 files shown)


src/mem/alloc.h
```diff
@@ -62,20 +62,112 @@ namespace snmalloc
...
+  class ExternalGlobalPagemap
+  {
+    /**
+     * A pointer to the pagemap.
+     */
+    SuperslabPagemap* external_pagemap;
+
+  public:
+    /**
+     * Construtor.  Accesses the pagemap via the C ABI accessor and casts it to
```
mjp41: typo

src/mem/pagemap.h
```diff
@@ -10,6 +10,36 @@ namespace snmalloc
...
   static constexpr size_t PAGEMAP_NODE_SIZE = 1ULL << PAGEMAP_NODE_BITS;
 
+  /**
+   * Structure describing the configuration of a pagemap.  When querying a
+   * pagemap from a different instantiation of snmalloc, the pagemap is exposed
+   * as a `void*`.  This structure allows the caller to check whether the
+   * pagemap is of the format that they expect.
+   */
+  struct PagemapConfig
+  {
```
mjp41: Should address bits be part of this?  We could end up exposing across a 32/64 bit boundary?  Many things would go wrong.  But perhaps this would catch it early.

src/test/func/two_alloc_types/main.cc
```diff
@@ -42,6 +47,10 @@ int main()
   oe_end = (uint8_t*)oe_base + size;
   std::cout << "Allocated region " << oe_base << " - " << oe_end << std::endl;
 
+  // Call these functions to trigger asserts if the cast-to-self doesn't work.
+  enclave_snmalloc_pagemap_global_get(nullptr);
```
mjp41: Doesn't passing nullptr, mean this will not do a check?

src/mem/alloc.h
```diff
@@ -62,20 +62,112 @@ namespace snmalloc
...
+     * the expected type, failing in cases of ABI mismatch.
+     */
+    ExternalGlobalPagemap()
+    {
+      const snmalloc::PagemapConfig* c;
+      external_pagemap =
+        SuperslabPagemap::cast_to_pagemap(snmalloc_pagemap_global_get(&c), c);
```
mjp41: This can't be used, if `snmalloc_pagemap_global_get` ignores the config.  Should we make cast able to take a null config and just succeed?

src/mem/pagemap.h
```diff
@@ -168,6 +198,31 @@ namespace snmalloc
...
+    /**
+     * Cast a `void*` to a pointer to this template instantiation, given a
+     * config describing the configuration.  Return null if the configuration
+     * passed does not correspond to this template instantiation.
+     *
+     * This intended to allow code that depends on the pagemap having a
+     * specific representation to fail gracefully.
+     */
+    static Pagemap* cast_to_pagemap(void* pm, const PagemapConfig* c)
+    {
```
mjp41: Check for null config and succeed?

src/mem/pagemap.h
davidchisnall: Good idea, added

src/mem/pagemap.h
davidchisnall: I don't see when that would be useful.  If you have a null config, it's only because you haven't asked for one. You can just do a `static_cast` if you don't want any type checking.

src/test/func/two_alloc_types/main.cc
davidchisnall: Ooops, yes.

src/mem/alloc.h
```diff
@@ -62,20 +62,82 @@ namespace snmalloc
...
+    /**
+     * Returns the pagemap.
+     */
+    SuperslabPagemap& pagemap()
+    {
+      return global_pagemap;
+    }
+  };
+
+  /**
```
rschust: It might be more consistent to ifdef this section (including ExternalGlobalPagemap) using SNMALLOC_EXPOSE_PAGEMAP

davidchisnall: If we did ifdef that, we'd want to use a different macro, because one is for the compilation unit that exports the pagemap and the other is for the compilation units that import it.  In most use cases, consumers of this function will not be in libraries that implement it.  In the FreeBSD builds, we're exporting `snmalloc_pagemap_global_get` from libc, but we're then importing it into everything else that wants to interop with it.

PR #35, opened 2019-04-12
Shared allocator cmake support

rschust:
Passing the EXPOSE_EXTERNAL_PAGEMAP and EXPOSE_EXTERNAL_RESERVE arguments to cmake will now expose the global pagemap and the default memory provider's reserve API on libsnmallocshim*.so accordingly.


CMakeLists.txt
```diff
@@ -67,6 +68,10 @@ if(USE_SBRK)
   add_definitions(-DUSE_SBRK)
 endif()
 
+if(EXPOSE_EXTERNAL_ALLOCATOR_API)
+  add_definitions(-DSNMALLOC_EXPOSE_RESERVE -DSNMALLOC_EXPOSE_PAGEMAP)
```
davidchisnall: Do we want expose reserve and expose external pagemap to be in the same define?  It might not matter for the shim library - I'm currently only exposing pagemap, when I want to reuse the Pal, I'm including the relevant `pal_{foo}.h` files and using the one for the host platform.

src/test/func/two_alloc_types/alloc1.cc
```diff
@@ -4,7 +4,9 @@
 #define USE_RESERVE_MULTIPLE 1
 #define NO_BOOTSTRAP_ALLOCATOR
 #define IS_ADDRESS_SPACE_CONSTRAINED
-#define SNMALLOC_EXPOSE_PAGEMAP
+#ifndef SNMALLOC_EXPOSE_PAGEMAP
```
davidchisnall: This shouldn't be needed.

CMakeLists.txt
davidchisnall: This shouldn't be add_definitions, we should be adding these definitions only for the shim library target, not unconditionally for everything.  We should probably also hide this option on Windows, where the shim library isn't built.

PR #36, opened 2019-04-25
General code cleaning

mjp41:
General code tidying to make auditing easier, mostly 
* Reducing uses of casts.
* Reducing standard library includes.

### DIFF
--- a/src/ds/bits.h
+++ b/src/ds/bits.h
@@ -453,13 +453,13 @@ namespace snmalloc
       }
     }
 
-    template <typename T>
+    template<typename T>
     constexpr inline T min(T t1, T t2)
     {
       return t1 < t2 ? t1 : t2;
     }
 
-    template <typename T>
+    template<typename T>
     constexpr inline T max(T t1, T t2)
     {
       return t1 > t2 ? t1 : t2;
--- a/src/ds/bits.h
+++ b/src/ds/bits.h
@@ -452,5 +452,17 @@ namespace snmalloc
         return (size_t)1 << (m_e + LOW_BITS);
       }
     }
+
+    template <typename T>
+    constexpr inline T min(T t1, T t2)
+    {
+      return t1 < t2 ? t1 : t2;
+    }
+
+    template <typename T>
+    constexpr inline T max(T t1, T t2)
+    {
+      return t1 > t2 ? t1 : t2;
+    }
   }
 }
--- a/src/ds/bits.h
+++ b/src/ds/bits.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <stddef.h>
+#include <limits>
 
 #ifdef _MSC_VER
 #  include <immintrin.h>
--- a/src/ds/bits.h
+++ b/src/ds/bits.h
@@ -453,12 +453,24 @@ namespace snmalloc
       }
     }
 
+    /**
+     * Implementation of std::min
+     *
+     * std::min is algorithm, so pulls in a lot of unneccessary code
+     * We write our own to reduce the code that potentially needs reviewing.
+     **/
     template<typename T>
     constexpr inline T min(T t1, T t2)
     {
       return t1 < t2 ? t1 : t2;
     }
 
+    /**
+     * Implementation of std::max
+     *
+     * std::max is algorithm, so pulls in a lot of unneccessary code
+     * We write our own to reduce the code that potentially needs reviewing.
+     **/
     template<typename T>
     constexpr inline T max(T t1, T t2)
     {
--- a/src/ds/bits.h
+++ b/src/ds/bits.h
@@ -1,7 +1,7 @@
 #pragma once
 
-#include <stddef.h>
 #include <limits>
+#include <stddef.h>
 
 #ifdef _MSC_VER
 #  include <immintrin.h>
--- a/src/ds/bits.h
+++ b/src/ds/bits.h
@@ -454,9 +454,9 @@ namespace snmalloc
     }
 
     /**
-     * Implementation of std::min
+     * Implementation of `std::min`
      *
-     * std::min is algorithm, so pulls in a lot of unneccessary code
+     * `std::min` is in `<algorithm>`, so pulls in a lot of unneccessary code
      * We write our own to reduce the code that potentially needs reviewing.
      **/
     template<typename T>
@@ -466,9 +466,9 @@ namespace snmalloc
     }
 
     /**
-     * Implementation of std::max
+     * Implementation of `std::max`
      *
-     * std::max is algorithm, so pulls in a lot of unneccessary code
+     * `std::max` is in `<algorithm>`, so pulls in a lot of unneccessary code
      * We write our own to reduce the code that potentially needs reviewing.
      **/
     template<typename T>
--- a/src/ds/dllist.h
+++ b/src/ds/dllist.h
@@ -6,16 +6,24 @@
 
 namespace snmalloc
 {
-  template<class T, uintptr_t terminator = 0>
+  template<class T, uintptr_t terminator_ = 0>
   class DLList
   {
   private:
-    static_assert(
-      std::is_same<decltype(((T*)0)->prev), T*>::value, "T->prev must be a T*");
-    static_assert(
-      std::is_same<decltype(((T*)0)->next), T*>::value, "T->next must be a T*");
+    // Single point to perform this cast.
+    // Would like this to be a constexpr, but reinterpret cast is not allowed in
+    // constexpr.
+    static inline T* terminator()
+    {
+      return (T*)terminator_;
+    }
 
-    T* head = (T*)terminator;
+    static_assert(
+      std::is_same<decltype(T::prev), T*>::value, "T->prev must be a T*");
+    static_assert(
+      std::is_same<decltype(T::next), T*>::value, "T->next must be a T*");
+
+    T* head = terminator();
 
   public:
     T* get_head()
@@ -27,7 +35,7 @@ namespace snmalloc
     {
       T* item = head;
 
-      if (item != (T*)terminator)
+      if (item != terminator())
         remove(item);
 
       return item;
@@ -40,9 +48,9 @@ namespace snmalloc
 #endif
 
       item->next = head;
-      item->prev = (T*)terminator;
+      item->prev = terminator();
 
-      if (head != (T*)terminator)
+      if (head != terminator())
         head->prev = item;
 
       head = item;
@@ -57,10 +65,10 @@ namespace snmalloc
       debug_check_contains(item);
 #endif
 
-      if (item->next != (T*)terminator)
+      if (item->next != terminator())
         item->next->prev = item->prev;
 
-      if (item->prev != (T*)terminator)
+      if (item->prev != terminator())
         item->prev->next = item->next;
       else
         head = item->next;
@@ -78,7 +86,7 @@ namespace snmalloc
 
       while (curr != item)
       {
-        assert(curr != (T*)terminator);
+        assert(curr != terminator());
         curr = curr->next;
       }
 #else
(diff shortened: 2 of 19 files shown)


mjp41:
@nwf,  @davidchisnall mentioned you had been doing some clean up to bring up snmalloc for Cheri.  Could you take a look at this to make sure I am not making your life harder.  Thanks

src/ds/dllist.h
```diff
@@ -6,16 +6,24 @@
 
 namespace snmalloc
 {
-  template<class T, uintptr_t terminator = 0>
+  template<class T, uintptr_t terminator_ = 0>
```
davidchisnall: Why is this getting a trailing underscore?  It looks like a typo everywhere.  Isn't the convention to use Lead Capitals (or ALL CAPITALS) for template parameters to differentiate them from other values?

src/ds/dllist.h
```diff
@@ -6,16 +6,24 @@
...
   class DLList
   {
   private:
+    // Single point to perform this cast.
+    // Would like this to be a constexpr, but reinterpret cast is not allowed in
+    // constexpr.
+    static inline T* terminator()
+    {
+      return (T*)terminator_;
```
davidchisnall: Please can we avoid new C-style casts?  I think this wants to be a `static_cast`.

src/ds/bits.h
```diff
@@ -452,5 +452,29 @@ namespace snmalloc
         return (size_t)1 << (m_e + LOW_BITS);
       }
     }
+
+    /**
+     * Implementation of std::min
+     *
+     * std::min is algorithm, so pulls in a lot of unneccessary code
```
davidchisnall: Probably means 'in `<algorithm>`'. For doc comments, we can use Markdown syntax (Doxygen will pull it out sensibly, so please also stick backticks around `std::min` and `std::max`).

src/ds/dllist.h
davidchisnall: Can we make this a type with a default constructor, rather than uintptr_t?  That would then let us use `nullptr_t`, which is default constructed as `nullptr`.  If we require the type to be implicitly convertable to `T*` (which `nullptr_t` is already) then this also removes the need for the `terminator()` function.  The following is fine:

```c++
T *a = std::nullptr_t();
```

We can avoid depending on `<cstddef>` by using `decltype(nullptr)` as the default, rather than `nullptr_t`.

src/mem/alloc.h
```diff
@@ -135,10 +133,19 @@ namespace snmalloc
     /**
      * Get the pagemap entry corresponding to a specific address.
      */
-    uint8_t get(void* p)
+    uint8_t get(uintptr_t p)
```
davidchisnall: @nwf, what do you think here?  My gut feeling is that we want this to be a non-provenance-carrying virtual address on CHERI, because we're going to derive a capability via the pagemap, so this probably wants to be `size_t` not `uintptr_t`?

src/mem/alloc.h
```diff
@@ -135,10 +133,19 @@ namespace snmalloc
...
     {
       return PagemapProvider::pagemap().get(p);
     }
+
+    /**
+     * Get the pagemap entry corresponding to a specific address.
+     */
+    uint8_t get(void* p)
+    {
+      return get((uintptr_t)p);
```
davidchisnall: `reinterpret_cast`.

src/mem/alloc.h
```diff
@@ -179,26 +186,27 @@ namespace snmalloc
       size_t size_bits = bits::next_pow2_bits(size);
       set(p, (uint8_t)size_bits);
       // Set redirect slide
-      uintptr_t ss = (uintptr_t)((size_t)p + SUPERSLAB_SIZE);
+      uintptr_t ss = (uintptr_t)p + SUPERSLAB_SIZE;
```
davidchisnall: `reinterpret_cast`

src/mem/alloc.h
```diff
@@ -179,26 +186,27 @@ namespace snmalloc
...
       for (size_t i = 0; i < size_bits - SUPERSLAB_BITS; i++)
       {
         size_t run = 1ULL << i;
         PagemapProvider::pagemap().set_range(
-          (void*)ss, (uint8_t)(64 + i + SUPERSLAB_BITS), run);
-        ss = (uintptr_t)ss + SUPERSLAB_SIZE * run;
+          ss, (uint8_t)(64 + i + SUPERSLAB_BITS), run);
```
davidchisnall: `static_cast` (not a new issue, but if you're changing this line...).

src/mem/alloc.h
```diff
@@ -179,26 +186,27 @@ namespace snmalloc
...
+        ss = ss + SUPERSLAB_SIZE * run;
       }
-      PagemapProvider::pagemap().set(p, (uint8_t)size_bits);
+      PagemapProvider::pagemap().set((uintptr_t)p, (uint8_t)size_bits);
```
davidchisnall: `reinterpret_cast`, `static_cast`.

src/mem/alloc.h
```diff
@@ -179,26 +186,27 @@ namespace snmalloc
...
     }
     /**
      * Update the pagemap to remove a large allocation, of `size` bytes from
      * address `p`.
      */
-    void clear_large_size(void* p, size_t size)
+    void clear_large_size(void* vp, size_t size)
     {
+      uintptr_t p = (uintptr_t)vp;
```
davidchisnall: `reinterpret_cast`.

We should probably add clang-tidy to our CI...

src/mem/alloc.h
```diff
@@ -209,7 +217,7 @@ namespace snmalloc
      */
     void set(void* p, uint8_t x)
     {
-      PagemapProvider::pagemap().set(p, x);
+      PagemapProvider::pagemap().set((uintptr_t)p, x);
```
davidchisnall: `reinterpret_cast`

src/mem/alloc.h
```diff
@@ -419,7 +427,7 @@ namespace snmalloc
 
       // Free memory of an unknown size. Must be called with an external
       // pointer.
-      uint8_t size = pagemap().get(p);
+      uint8_t size = pagemap().get((uintptr_t)p);
```
davidchisnall: `reinterpret_cast`

src/mem/alloc.h
```diff
@@ -462,7 +470,7 @@ namespace snmalloc
       }
 
 #  ifndef SNMALLOC_SAFE_CLIENT
-      if (size > 64 || (void*)super != p)
+      if (size > 64 || (uintptr_t)super != (uintptr_t)p)
```
davidchisnall: `reinterpret_cast`.

src/mem/alloc.h
```diff
@@ -472,13 +480,13 @@ namespace snmalloc
     }
 
     template<Boundary location = Start>
-    static void* external_pointer(void* p)
+    static uintptr_t external_uintptr(void* p)
```
davidchisnall: I don't like the name of this (what is an external uintptr?).  I think it would be a lot cleaner to define an `address_t` (which is probably `uint32_t` or `uint64_t`) and make this `external_address()`.  We can then insert explicit conversions from addresses to pointers that will run via the pagemap in the CHERI case and be `reinterpret_cast`s in the non-CHERI case.

src/mem/alloc.h
```diff
@@ -472,13 +480,13 @@ namespace snmalloc
...
     {
 #ifdef USE_MALLOC
       error("Unsupported");
       UNUSED(p);
 #else
-      uint8_t size = global_pagemap.get(p);
+      uint8_t size = global_pagemap.get((uintptr_t)p);
```
davidchisnall: `reinterpret_cast`

src/mem/alloc.h
```diff
@@ -507,33 +515,39 @@ namespace snmalloc
...
-        size = global_pagemap.get((void*)ss);
+        size = global_pagemap.get(ss);
       }
 
       if (size == 0)
       {
         if constexpr ((location == End) || (location == OnePastEnd))
           // We don't know the End, so return MAX_PTR
-          return (void*)-1;
+          return (uintptr_t)~0;
```
davidchisnall: `reinterpret_cast`, though what you probably actually mean is `uintptr_t(~0ULL)`.

src/mem/alloc.h
```diff
@@ -974,7 +987,7 @@ namespace snmalloc
       SlabLink* link = sc->get_head();
       Slab* slab;
 
-      if (link != (SlabLink*)~0)
+      if (~(uintptr_t)link != 0)
```
davidchisnall: `reinterpret_cast`

src/mem/allocconfig.h
```diff
@@ -84,6 +84,8 @@ namespace snmalloc
     ;
 
   // The remaining values are derived, not configurable.
+  static constexpr size_t POINTER_BITS =
```
davidchisnall: Unused?

src/mem/alloc.h
nwf: I agree that it probably should be something other than `uintptr_t`, but I am kind of wary of `size_t`, because that isn't required to be large enough to be an entire virtual address, merely the size of any particular object in the system.  I know segmented architectures are largely a thing of the past, but it would be nice to have a type specifically for "virtual address" even if it had to be defined in one of snmalloc's platform-specific utility headers.

If the choice is between `uintptr_t` and `size_t` specifically, I think I'd prefer the former: it's guaranteed to be big enough without being too big to comfortably pass (unlike, say, `uintmax_t`) and we can always shed the CHERI tags inside `get()` itself.

src/mem/alloc.h
```diff
@@ -480,7 +480,7 @@ namespace snmalloc
     }
 
     template<Boundary location = Start>
-    static void* external_pointer(void* p)
+    static uintptr_t external_uintptr(void* p)
```
nwf: On a different topic, but while cleaning code, `external_pointer` and `alloc_size` should shed their `static` annotations: they will, I think, need to access the pagemap via the allocator's handle to it rather than specifically using the global variable.  Doing the latter in combination with the existence of `ExternalGlobalPagemap` seems like a recipe for trouble.  The test-suite needs some obvious, compiler-directed changes to follow along.

nwf:
> @nwf, @davidchisnall mentioned you had been doing some clean up to bring up snmalloc for Cheri. Could you take a look at this to make sure I am not making your life harder. Thanks

Thanks for reaching out.  From a quick look, none of this makes my life harder.

I don't yet have the story fully fleshed out for CHERI, but as a (naive?) first pass, I was hoping to make the page map store pointers to the superslabs and, at the top of `Allocator::dealloc` and friends (those functions taking pointers as arguments) push the user capability through the pagemap to get back a capability with wider base and bounds and leave the rest of the code more or less un-modified.  Does that seem sensible?

src/mem/alloc.h
mjp41: So I am not liking the `ULL`, after getting it wrong for 32bit pointers. 

I am thinking about adding an `constexpr UINTPTR_MAX` to bits to encapsulate the few places we want -1 as a pointer in one place.

src/mem/alloc.h
mjp41: That would be good.

src/mem/alloc.h
mjp41: So it sounds like we want an address_t, that is beyond what I wanted to do for now.

src/mem/alloc.h
mjp41: There are some cases where we want this code path to be very fast, so may have to do something extra, but I think that is orthogonal. And the code is currently a bit broken abstraction wise.

I see you have a change on your fork doing this.  Please feel free to PR this.

mjp41:
@nwf I think your proposal for Cheri sounds sensible.  Please feel free to raise a PR or issue for this.  Also it would be good to think about adding code/abstraction that makes it hard for us to break compatibility without having a Cheri CI machine.

PR #38, opened 2019-04-29
Clang tidy

davidchisnall:
Add a clang-tidy file and clean up all of the C-style casts.

This also adds an explicit address type, replaces all address<->pointer casts with explicit operations, and cleans up a few `size_t` uses.

@mjp41, please check the places where else blocks have gone away - in at least one, I think the original code had a logic error, but deleting the dead code may not actually be the right fix, we may have been missing a branch condition somewhere.

@nwf, hopefully this makes some of the CHERI work easier.


src/ds/address.h
```diff
@@ -0,0 +1,42 @@
...
+   * separated into two types, one for raw addresses and one for addresses that
+   * can be cast back to pointers.
+   */
+  typedef uintptr_t address_t;
+
+  /**
+   * Perform pointer arithmetic and return the adjusted pointer.
+   */
+  template<typename T>
+  inline T* pointer_offset(T* base, size_t diff)
```
mjp41: size_t or ptrdiff_t?  I am happy with size_t.

src/ds/bits.h
```diff
@@ -212,7 +216,7 @@ namespace snmalloc
 
       for (int i = BITS - 1; i >= 0; i--)
       {
-        size_t mask = (size_t)1 << i;
+        size_t mask = static_cast<size_t>(1) << i;
```
mjp41: I wonder if we should add
``` C++
static constexpr size_t ONE = size_t(1);
```
and then use that.

src/mem/pagemap.h
```diff
@@ -78,7 +78,7 @@ namespace snmalloc
       (INDEX_LEVELS * BITS_PER_INDEX_LEVEL) + BITS_FOR_LEAF + GRANULARITY_BITS;
 
     // Value used to represent when a node is being added too
-    static constexpr uintptr_t LOCKED_ENTRY = 1;
+    static constexpr InvalidPointer<1> LOCKED_ENTRY;
```
mjp41: This is breaking in the Windows build.

src/ds/address.h
davidchisnall: Callers pass `size_t` - I made it `ptrdiff_t` initially, but got lots of warnings about sign conversions.  I'd like to make it `ptrdiff_t` eventually, but it requires cleaning up all of the callers.

src/ds/bits.h
```diff
@@ -225,7 +235,7 @@ namespace snmalloc
 
       for (int i = BITS - 1; i >= 0; i--)
```
mjp41:
```C
for (int i = BITS - 1; i >= 0; i--)
```
should be 
```C
for (size_t i = BITS - 1; i >= 0; i--)
```

src/mem/metaslab.h
```diff
@@ -102,10 +102,13 @@ namespace snmalloc
...
       size_t size = sizeclass_to_size(sizeclass);
       size_t offset = get_slab_offset(sizeclass, is_short);
+      size_t all_ones = std::numeric_limits<size_t>::is_signed ?
+        size_t(-1) :
+        std::numeric_limits<size_t>::max();
 
       size_t head_start =
-        remove_cache_friendly_offset(head & ~static_cast<size_t>(1), sizeclass);
-      size_t slab_start = offset & ~static_cast<size_t>(1);
+        remove_cache_friendly_offset(head & all_ones, sizeclass);
```
mjp41: What is this doing?
```C 
& all_ones
```
looks like a nop here.  Has something gotten broken here?

I think this use to be
```C
      size_t head_start =
        remove_cache_friendly_offset(head & ~(size_t)1, sizeclass);
      size_t slab_start = offset & ~(size_t)1;
```

Looks like it is unsetting the bottom bit. I.e. `_ & 0xffff...fffe`

davidchisnall: Ah, this was me not reading the code that I was changing properly.

src/ds/bits.h
davidchisnall: I don't know what this is trying to do, but that change make this an infinite loop.  Specifically, an infinite loop at compile time in a constexpr that's used as a template parameter...

mjp41: Sorry, there was a int -> size_t conversion complaint from this line.

PR #40, opened 2019-05-03
Fix the GCC build

mjp41:
This PR fixes the GCC build.


davidchisnall:
Looks fine, but please can you update the README to say that we can build with gcc but that some symbols will be preemptible if you do.

PR #41, opened 2019-05-07
CHERI Preparatory work

nwf:
Some portability fixes and some changes to pointer provenance throughout.  No explicit MIPS or CHERI support yet (on this branch), but perhaps subsets of this are ready to merge and others deserve a chance to receive commentary. :)


src/ds/address.h
```diff
@@ -39,4 +39,32 @@ namespace snmalloc
...
-} // namespace snmalloc
+
+  template<size_t granule, typename T = void>
+  inline T* pointer_align_down(void* p)
+  {
+    static_assert((granule & (granule - 1)) == 0);
+#if defined(__has_builtin) && __has_builtin(__builtin_align_down)
+    return reinterpret_cast<T*>(__builtin_align_down(p, granule));
+#else
+    /* XREF bits::align_down; can't use here for cyclic deps */
```
mjp41: If we move `bits::is_aligned_block` then we could move `bits.h` before `address.h`?

src/mem/alloc.h
```diff
@@ -45,6 +45,17 @@ namespace snmalloc
     PMNotOurs = 0,
     PMSuperslab = 1,
     PMMediumslab = 2
+
```
mjp41: This description is correct and useful.  Thank you. Could you remove the first person supposition :-)

src/mem/alloc.h
```diff
@@ -45,6 +45,17 @@ namespace snmalloc
...
+    /*
+     * Values SUPERSLAB_BITS (inclusive) through 64 (exclusive, I suppse)
+     * are used for log2(size) at the heads of large allocations.
+     *
+     * Values 64 (inclusive) through 128 (exclusive) are used for entries
```
mjp41: The 64-128 are used for backwards jumps towards the start of an object.  They are the largest power of two that does not jump beyond the start.  I.e.  79 would be that there are at least `2^(79 - 64)` entries before this one to find the start of the object.  This allows logarithmic access to the start of the object.

src/mem/alloc.h
```diff
@@ -480,13 +492,13 @@ namespace snmalloc
     }
 
     template<Boundary location = Start>
-    static address_t external_address(void* p)
+    address_t external_uintptr(void* p)
```
mjp41: Why did you change the name back?  Looks like a merge conflict.

src/mem/alloc.h
```diff
@@ -794,16 +805,16 @@ namespace snmalloc
...
       size_t rsize = sizeclass_to_size(sizeclass);
-      size_t end_point_correction = location == End ?
-        (end_point - 1) :
-        (location == OnePastEnd ? end_point : (end_point - rsize));
-      size_t offset_from_end =
-        (end_point - 1) - static_cast<size_t>(address_cast(p));
-      size_t end_to_end = round_by_sizeclass(rsize, offset_from_end);
-      return end_point_correction - end_to_end;
+      void *end_point_correction = location == End ?
+        ((char*)end_point - 1) :
```
mjp41: Can we use `uint8_t*` rather than `char*`.  I find this more descriptive.

src/mem/largealloc.h
```diff
@@ -145,14 +145,20 @@ namespace snmalloc
       {
         FlagLock f(lock);
 
-        auto aligned_bump = bits::align_up(bump, alignment);
-        if ((aligned_bump - bump) > remaining)
+        char* aligned_bump;
+        if constexpr (alignment != 0) {
+          aligned_bump = __builtin_align_up(bump, alignment);
```
mjp41: Should be the thing you defined in `address.h`

src/mem/largealloc.h
```diff
@@ -145,14 +145,20 @@ namespace snmalloc
...
-        auto aligned_bump = bits::align_up(bump, alignment);
-        if ((aligned_bump - bump) > remaining)
+        char* aligned_bump;
+        if constexpr (alignment != 0) {
+          aligned_bump = __builtin_align_up(bump, alignment);
+        } else {
+          aligned_bump = bump;
+        }
+
+        if ((unsigned long)(aligned_bump - bump) > remaining)
```
mjp41: No C style casts.

src/mem/largealloc.h
```diff
@@ -145,14 +145,20 @@ namespace snmalloc
...
         {
           new_block();
         }
         else
         {
-          remaining -= aligned_bump - bump;
+          remaining -= (unsigned long)(aligned_bump - bump);
```
mjp41: No C style casts please.

src/mem/mediumslab.h
```diff
@@ -41,7 +41,7 @@ namespace snmalloc
 
     static Mediumslab* get(void* p)
     {
-      return pointer_cast<Mediumslab>(address_cast(p) & SUPERSLAB_MASK);
+      return pointer_align_down<SUPERSLAB_SIZE, Mediumslab>(p);
```
mjp41: Nice

src/mem/slab.h
```diff
@@ -39,7 +39,7 @@ namespace snmalloc
 
       assert(rsize == sizeclass_to_size(meta.sizeclass));
       meta.debug_slab_invariant(is_short(), this);
-      assert(sc->get_head() == (SlabLink*)((size_t)this + meta.link));
+      assert(sc->get_head() == (SlabLink*)pointer_offset(this,meta.link));
```
mjp41: @davidchisnall why didn't clang-tidy pick this already.  Should we run clang tidy on a debug and a release build?

src/ds/bits.h
```diff
@@ -109,7 +109,9 @@ namespace snmalloc
 
     inline void pause()
     {
-#if defined(PLATFORM_IS_X86)
+#if defined(__clang__) && __has_builtin(__builtin_pause)
```
davidchisnall: Why do we need the `defined(__clang__)` check here?  If another compiler implements `__has_builtin` and `__builtin_pause` then we probably want to use that.  There's no short-circuit evaluation in the preprocessor, so this isn't guarding the `__has_builtin` check...

src/ds/bits.h
```diff
@@ -118,11 +120,13 @@ namespace snmalloc
 
     inline uint64_t tick()
     {
-#if defined(PLATFORM_IS_X86)
+#if defined(__clang__) && __has_builtin(__builtin_readcyclecounter)
```
davidchisnall: As above.

src/mem/alloc.h
```diff
@@ -516,7 +527,7 @@ namespace snmalloc
       {
         // This is a large alloc redirect.
         ss = ss - (1ULL << (size - 64));
-        size = global_pagemap.get(ss);
+        size = pagemap().get((void*)ss);
```
davidchisnall: Please don't add C-style casts (they will cause clang-tidy to be unhappy in CI).  You shouldn't need to cast to `void*` (C++ allows implicit casts to `void*`, but not implicit casts *from* `void*`).

src/mem/alloc.h
```diff
@@ -540,15 +551,15 @@ namespace snmalloc
     }
 
     template<Boundary location = Start>
-    static void* external_pointer(void* p)
+    void* external_pointer(void* p)
     {
-      return pointer_cast<void>(external_address<location>(p));
+      return pointer_cast<void>(external_uintptr<location>(p));
```
davidchisnall: Merge conflict?

src/mem/alloc.h
davidchisnall: Since we're doing addressable-unit calculations, I think `char*` is technically more correct here, though I doubt anyone will try to use snmalloc on 16-bit addressable platforms (mind you, I might have said the same thing about platfroms with 128-bit fat pointers...).

src/mem/alloc.h
```diff
@@ -794,16 +805,16 @@ namespace snmalloc
...
-        (end_point - 1) - static_cast<size_t>(address_cast(p));
-      size_t end_to_end = round_by_sizeclass(rsize, offset_from_end);
-      return end_point_correction - end_to_end;
+      void *end_point_correction = location == End ?
+        ((char*)end_point - 1) :
+        (location == OnePastEnd ? end_point : pointer_offset(end_point,-rsize));
+      ptrdiff_t offset_from_end =
+        ((char*)end_point - 1) - (char*)p;
+      size_t end_to_end = round_by_sizeclass(rsize, (size_t)offset_from_end);
+      return address_cast<char>((char*)end_point_correction - end_to_end);
```
davidchisnall: Lots of C-style casts added here.  Please can you turn them back into teh correct kinds of C++ casts?

src/mem/largealloc.h
```diff
@@ -58,7 +58,7 @@ namespace snmalloc
   class MemoryProviderStateMixin : public PAL
   {
     std::atomic_flag lock = ATOMIC_FLAG_INIT;
-    address_t bump;
+    char* bump;
```
davidchisnall: Why `char*` and not `void*`?  This means we need some explicit casts later and are doing explicit pointer arithmetic on `char*` rather than with the `pointer_offset` function, which makes auditing the code for pointer arithmetic harder.

src/mem/largealloc.h
```diff
@@ -161,16 +167,17 @@ namespace snmalloc
...
-      auto page_start = bits::align_down(address_cast(p), OS_PAGE_SIZE);
-      auto page_end = bits::align_up(address_cast(p) + size, OS_PAGE_SIZE);
+      auto page_start = pointer_align_down<OS_PAGE_SIZE, char>(p);
+      auto page_end = pointer_align_down<OS_PAGE_SIZE, char>
+          (pointer_offset(p,size));
 
-      PAL::template notify_using<NoZero>(
-        pointer_cast<void>(page_start), page_end - page_start);
+      PAL::template notify_using<NoZero>(page_start,
+        (size_t)(page_end - page_start));
```
davidchisnall: C-style cast.

src/mem/slab.h
davidchisnall: It might miss things in macro instantiations?

src/mem/alloc.h
mjp41: If we aren't a byte addressable platform, I have no idea what would happen with this code base.  

`uint8_t` is in some sense "more" correct, as it codifies an assumption. :-)

src/ds/address.h
nwf: `bits::hash` also depends on `address_cast`.

src/mem/alloc.h
nwf: 'twas, indeed; thanks for the catch.

src/mem/largealloc.h
nwf: I think this change predates the existence of `pointer_offset`; I've moved things over to `void *` and `pointer_offset`, though there's still pointer subtraction without a wrapper method.

mjp41:
I am just investigating the Windows 32bit failures.  We have been having an intermittent failure, thought this looks different.

I have not managed to duplicate this locally.

nwf:
Well, that's better, but win 32 is still sad.

src/mem/largealloc.h
```diff
@@ -145,14 +145,25 @@ namespace snmalloc
...
+        if constexpr (alignment != 0)
+        {
+          aligned_bump = pointer_align_up<alignment, uint8_t>(bump);
+        }
+        else
+        {
+          aligned_bump = static_cast<uint8_t*>(bump);
+        }
+
+        unsigned long bump_delta = static_cast<unsigned long>(
```
mjp41: Why is this `unsigned long` shouldn't this be `size_t`?  On 64 bit Windows long is only 32bit.

src/mem/largealloc.h
```diff
@@ -145,14 +145,25 @@ namespace snmalloc
       {
         FlagLock f(lock);
 
-        auto aligned_bump = bits::align_up(bump, alignment);
-        if ((aligned_bump - bump) > remaining)
+        uint8_t* aligned_bump;
```
mjp41: Could we rewrite this to scope aligned_bump inside the 
```C
if constexpr (alignment != 0) 
{ 
          uint8_t* aligned_bump = pointer_align_up<alignment, uint8_t>(bump);
          size_t bump_delta = ...
          if (bump_delta > remaining)
          {
                new_block();
          }
         else
         {
              remaining -= bump_delta;
              bump = aligned_bump;
         }
}

File: src/ds/address.h [hunk 20]
```diff
@@ -39,4 +39,31 @@ namespace snmalloc
...
+    size_t align_1 = granule - 1;
+    return reinterpret_cast<T*>(reinterpret_cast<uintptr_t>(p) & ~align_1);
+#endif
+  }
+
+  template<size_t granule, typename T = void>
+  inline T* pointer_align_up(void* p)
+  {
+    static_assert((granule & (granule - 1)) == 0);
```
mjp41: Does this work if `granule` is zero? Perhaps forbid with a static assert?

src/mem/largealloc.h
nwf: `unsigned long` was me just being bad and placating my compiler, I'm sure.  `size_t` is the right answer and I'll move to that.

src/ds/address.h
```diff
@@ -39,4 +39,33 @@ namespace snmalloc
...
+  {
+    static_assert(granule > 0);
+    static_assert((granule & (granule - 1)) == 0);
+#if defined(__has_builtin) && __has_builtin(__builtin_align_up)
+    return reinterpret_cast<T*>(__builtin_align_up(p, granule));
+#else
+    /* XREF bits::align_up; can't use here for cyclic deps */
+    size_t align_1 = granule - 1;
+    return reinterpret_cast<T*>(
+      reinterpret_cast<uintptr_t>(pointer_offset(p, granule)) & ~align_1);
```
mjp41: I think the problem with 32bit build is caused by:
```
pointer_offset(p, granule)
```
should be
```
pointer_offset(p, align_1)
```
This is causing a lot of Virtual address space to be wasted, and leading to the issue.

mjp41:
@nwf Any news on the CLA approval?  I have some changes, that might make your life much harder if we don't get your PR in soon.

nwf:
@mjp41 I'm deep in a queue for the University's legal department, so far as I know.  It's probably going to be a while.  I'll suffer rebases in the interim.

src/ds/address.h
nwf: Ah ha.  Good catch.  I'm out of the office right now but I'll push a replacement as soon as I can.

src/ds/address.h
nwf: I see that `bits::hash` is no more; where would you like `bits::is_aligned_block` instead?

mjp41: I think it makes most sense in `address.h`. Thanks

mjp41: Then you can use `bits.h` align functions when the compiler does not support `__builtin_align_up`.

mjp41:
@nwf Any news on the CLA with the University?

nwf:
I have heard nothing from Cambridge legal; I'll nudge Robert to nudge them again.  Sorry.

mjp41:
Same question again: @nwf Any news on the CLA with the University?

nwf:
So far as I know, there's still a queue of other CHERI-related questions for Cambridge legal.  I'll poll again.  Sorry.  If you'd prefer I close this PR while waiting, since I am still also refactoring and rebasing as work continues, I'm happy to reopen it later.

mjp41:
I am happy for you to leave it open, I'll just prod you once a month while it is open :)

nwf:
OK!  I got permission from Cambridge to sign the CLA.

mjp41:
That's great news.  Have other people in Cambridge contributed to what you will want to integrate?  Do they also have permission to sign the CLA?

nwf:
The CHERI+snmalloc work has been just me on the Cambridge side, so I don't think we need anyone else's CLA.

nwf:
Well it sure has been a while, but here we are at last.  Almost all of this is just nits, and so I hope fairly boring to review (sorry), and I tried to exclude anything that's overtly CHERI-specific so that all of this can be CI'd without needing CHERI infrastructure.

davidchisnall:
It seems to have broken all platforms in CI, which is nicely consistent.

We had some conversations with Robert last week about CI for CHERI.  I think, in the scope of the ISCF, we'll aim to put some joint effort into having reuseable CI parts for Azure DevOps + CHERI, but I don't yet know what that will look like: probably a weekly CI job that creates many of the things that cheribuild builds and makes them available for other CI jobs to pull in.  If we can copy the tests into a CheriBSD image and run them in QEMU, that would be ideal (well, not as good as having QEMU user mode working for CheriABI, but close).

nwf:
It's now just down to Windows, and it's not obvious to me what's broken... something to do with inserting into the small_classes[], but I don't think I touched anything there?  If someone with easy access to a Windows machine wouldn't mind bisecting, that may be informative.

src/mem/alloc.h
```diff
@@ -877,16 +1084,23 @@ namespace snmalloc
...
-        (location == OnePastEnd ? end_point : (end_point - rsize));
-      size_t offset_from_end =
-        (end_point - 1) - static_cast<size_t>(address_cast(p));
-      size_t end_to_end = round_by_sizeclass(rsize, offset_from_end);
-      return end_point_correction - end_to_end;
+
+      void* end_point_correction = location == End ?
+        (static_cast<uint8_t*>(end_point) - 1) :
+        (location == OnePastEnd ? end_point :
+                                  (static_cast<uint8_t*>(end_point) - rsize));
```
davidchisnall: Should this  (and the similar cases below) not be `pointer_offset`?

src/mem/allocconfig.h
```diff
@@ -4,6 +4,18 @@
 
 namespace snmalloc
 {
+#ifndef SNMALLOC_PAGEMAP_POINTERS
+#  define SNMALLOC_PAGEMAP_POINTERS 0
```
davidchisnall: Please document all of these magic constants.

src/mem/largealloc.h
```diff
@@ -58,7 +58,7 @@ namespace snmalloc
   class MemoryProviderStateMixin : public PAL
   {
     std::atomic_flag lock = ATOMIC_FLAG_INIT;
-    address_t bump;
+    void* bump;
```
davidchisnall: Wouldn't making this `char*` avoid a lot of the casts below?

src/mem/pagemap.h
```diff
@@ -241,7 +241,7 @@ namespace snmalloc
         (c->version != 1) || (c->is_flat_pagemap) ||
         (c->sizeof_pointer != sizeof(uintptr_t)) ||
         (c->pagemap_bits != GRANULARITY_BITS) ||
-        (c->size_of_entry != sizeof(T)) || (!std::is_integral_v<T>))
```
davidchisnall: You've deleted the is_integral check, but that probably means that we should be enforcing the type some other way.  Please extend the config struct to set `version = 2` for non-integral types and add a field at the end to specify that it's `void*` so that the checks still do the right thing and we can't mix pagemaps with incorrect compilation modes.

src/mem/pagemap.h
```diff
@@ -381,7 +379,15 @@ namespace snmalloc
      */
     size_t index_for_address(uintptr_t p)
     {
-      return reinterpret_cast<size_t>((p >> SHIFT) & (OS_PAGE_SIZE - 1));
+      size_t paddr;
+
+#if defined(__CHERI_PURE_CAPABILITY__)
+      paddr = static_cast<size_t>(static_cast<vaddr_t>(p));
```
davidchisnall: Is this incorrect in the non-CHERI case?  It seems that this should work both ways.

src/mem/pagemap.h
```diff
@@ -390,8 +396,17 @@ namespace snmalloc
     void* page_for_address(uintptr_t p)
     {
       assert((reinterpret_cast<uintptr_t>(&top) & (OS_PAGE_SIZE - 1)) == 0);
+
+      size_t paddr;
+#if defined(__CHERI_PURE_CAPABILITY__)
+      paddr = static_cast<size_t>(static_cast<vaddr_t>(p));
```
davidchisnall: As above.

src/override/malloc.cc
```diff
@@ -18,7 +18,7 @@ extern "C"
 {
   SNMALLOC_EXPORT void* SNMALLOC_NAME_MANGLE(__malloc_end_pointer)(void* ptr)
   {
-    return Alloc::external_pointer<OnePastEnd>(ptr);
+    return ThreadAlloc::get()->external_pointer<OnePastEnd>(ptr);
```
davidchisnall: This is on the fast path for some things that we have, which makes this look like it's a potentially expensive change.  Why is it needed?  A number of functions were made non-`static` above, but I'm not completely sure that I understand why.

src/test/func/malloc/malloc.cc
```diff
@@ -21,7 +21,7 @@ void check_result(size_t size, size_t align, void* p, int err, bool null)
     if (our_malloc_usable_size(p) < size)
       abort();
 
-    if (((uintptr_t)p % align) != 0)
+    if ((size_t)((uintptr_t)p % align) != 0)
```
davidchisnall: Please don't add C-style casts (and, if you're modifying code that contains them, please fix them).

src/ds/address.h
```diff
@@ -18,7 +21,7 @@ namespace snmalloc
   template<typename T>
   inline T* pointer_offset(T* base, size_t diff)
   {
-    return reinterpret_cast<T*>(reinterpret_cast<char*>(base) + diff);
+    return reinterpret_cast<T*>(reinterpret_cast<uint8_t*>(base) + diff);
```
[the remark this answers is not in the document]
nwf: That's fair enough; I was just following @mjp41's request to use `uint8_t*`, above.

src/override/malloc.cc
nwf: I think that commit predates https://github.com/microsoft/snmalloc/pull/86 and is perhaps stale, sorry.  Now that `pagemap()` is static, there's no risk of not finding the external pagemap.  I'll remove the commit from the next `push -f`.

src/mem/alloc.h
```diff
@@ -401,8 +545,11 @@ namespace snmalloc
 
       handle_message_queue();
 
-      // Free memory of a statically known size. Must be called with an
-      // external pointer.
+      if constexpr (SNMALLOC_PAGEMAP_REDERIVE)
```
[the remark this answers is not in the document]
nwf: That seems sensible.  This was here for measurement of costs, but I don't think we actually care to measure at that fine a granularity.  I'll eliminate all `SNMALLOC_PAGEMAP_REDERIVE` tests and its definition.

src/mem/pagemap.h
nwf: OK; new version forthcoming.

src/mem/alloc.h
nwf: They should!  Adjusted.

src/mem/alloc.h
```diff
@@ -565,19 +731,44 @@ namespace snmalloc
...
+          ss = ss - (1ULL << (size - 64));
+          size = pagemap().get(ss);
+        }
+
+        /*
+         * It may, for debugging, be useful to cross-check the pagemap
+         * pointers, by overriding the above use thereof.  In that case,
+         * we should arrive at the same answer.
+         */
+        if constexpr (SNMALLOC_PAGEMAP_POINTERS)
```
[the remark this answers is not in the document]
nwf: Yes; added.

mjp41:
> It's now just down to Windows, and it's not obvious to me what's broken... something to do with inserting into the small_classes[], but I don't think I touched anything there?  If someone with easy access to a Windows machine wouldn't mind bisecting, that may be informative.

The memory at the stack trace in CI is reserved but not committed. 

Bisecting gives:
  a533336a0b161d75650ed55c1129d0a5876cbe58

Haven't had time to look deeper.

mjp41:
So looking at the change the issues is 
```C++
|| (large_class > 0)
```
This is used as a guard for deallocation in `large_dealloc` for whether to call `not_using`.

I proposed you make change the else branch to:
```C++
      else
      {
        stats.superslab_pop();

        bool decommitted =
          ((decommit_strategy == DecommitSuperLazy) &&
           ((static_cast<Baseslab*>(p)->get_kind() == Decommitted))) ||
          (large_class > 0) || (decommit_strategy != DecommitNone);

        if (decommitted)
        {
          // The first page is already in "use" for the stack element,
          // this will need zeroing for a YesZero call.
            memory_provider.template zero<true>(p, OS_PAGE_SIZE);

          // Notify we are using the rest of the allocation.
          // Passing zero_mem ensures the PAL provides zeroed pages if
          // required.
          memory_provider.template notify_using<zero_mem>(
            pointer_offset(p, OS_PAGE_SIZE),
            bits::align_up(size, OS_PAGE_SIZE) - OS_PAGE_SIZE);
        }
        else
        {
          // This is a superslab that has not been decommitted.
            memory_provider.template zero<true>(
              p, bits::align_up(size, OS_PAGE_SIZE));
        }
      }
```

src/mem/alloc.h
```diff
@@ -158,36 +217,96 @@ namespace snmalloc
...
+    }
+
     /**
      * Set a pagemap entry indicating that there is a superslab at the
      * specified index.
      */
     void set_slab(Superslab* slab)
     {
-      set(slab, static_cast<size_t>(PMSuperslab));
+      if constexpr (SNMALLOC_PAGEMAP_POINTERS)
```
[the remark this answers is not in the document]
nwf: I took a stab at that.  I'm going to push up to see what CI thinks, even though I haven't addressed all your comments.  Mind taking a look again?

src/ds/address.h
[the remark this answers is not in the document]
mjp41: So `pointer_diff` and `pointer_offset` should be in the same units.  I am worried that the code assumes `char` is one `byte`, which was where my comments came from. But reading the standard sizeof(char)==1, so I think that is fine.  Let's use char*.

src/mem/alloc.h
```diff
@@ -40,20 +40,89 @@ namespace snmalloc
...
+     * allocation.  See SuperslabMap::set_large_size and external_address's
+     * handling of large reallocation redirections.
+     *
+     * Values 128 (inclusive) through 255 (inclusive) are as yet unused.
+     */
+
   };
 
+  /* Ensure that PageMapSuperslabKind values are actually disjoint */
+  static_assert(SUPERSLAB_BITS > 2, "Large allocations possibly too small");
```
mjp41: 2 or PMMediumSlab?

src/mem/alloc.h
```diff
@@ -196,18 +307,32 @@ namespace snmalloc
...
+        char* ss = reinterpret_cast<char*>(pointer_offset(p, SUPERSLAB_SIZE));
+        for (size_t i = 0; i < size_bits - SUPERSLAB_BITS; i++)
+        {
+          size_t run = 1ULL << i;
+          PagemapProvider::pagemap().set_range(
+            address_cast(ss), pointer_offset(p, 64 + i + SUPERSLAB_BITS), run);
+          ss += SUPERSLAB_SIZE * run;
+        }
+      }
+      else
```
mjp41: The two branches here, seem to be needlessly different
* `ss += ... ` versus `ss = ss + ...` 
* Different type of ss
* Use of address cast in different places. 
I think I would sooner, you push the `if` to be there twice, once around the `set` and once around the `set_range`, then it is clearer what the differences are.

src/mem/alloc.h
```diff
@@ -401,8 +557,8 @@ namespace snmalloc
 
       handle_message_queue();
 
-      // Free memory of a statically known size. Must be called with an
-      // external pointer.
+      p = pagemap().getp(p);
```
mjp41: This is on the fast path? Ensure it is erased on non-Cheri case.  Perhaps add a comment?

src/mem/alloc.h
```diff
@@ -438,8 +598,8 @@ namespace snmalloc
 #else
       handle_message_queue();
 
-      // Free memory of a dynamically known size. Must be called with an
-      // external pointer.
+      p = pagemap().getp(p);
```
mjp41: This is on the fast path? Ensure it is erased on non-Cheri case. Perhaps add a comment?

src/mem/alloc.h
```diff
@@ -158,36 +237,68 @@ namespace snmalloc
...
+     * Fetch a pointer through the pagemap.  The return value and the
+     * argument point at the same location, if the template parameter offset
+     * is true, but may have different metadata associated with them, such
+     * as CHERI bounds.  If the pagemap is storing pointers, then the
+     * template parameter offset may be false, in which case, the return
+     * value is exactly the pagemap entry, which will point at the bottom of
+     * the pagemap granule and include the argument in its bounds.
+     */
+    template<bool offset = true>
+    void* getp(void* p)
```
mjp41: Add fast path annotation, so we are guaranteed this will be erased in NON-Cheri case.

src/mem/alloc.h
```diff
@@ -469,16 +629,20 @@ namespace snmalloc
...
     {
 #ifdef USE_MALLOC
       return free(p);
 #else
 
-      // Free memory of an unknown size. Must be called with an external
-      // pointer.
       uint8_t size = pagemap().get(address_cast(p));
 
+      p = pagemap().getp(p);
```
mjp41: This is on the fast path? Ensure it is erased on non-Cheri case. Perhaps add a comment?

nwf:
I'm unhappy with the whole `SNMALLOC_PAGEMAP_POINTERS` thing, so I'm going to propose an alternative.  In the interim, I've pushed the bulk of these commits, which are not terribly specific to CHERI to their own PR (#103) and will then move the rest of this to another one, once the rest of the issues above are addressed.

mjp41:
This is so old now, I think it should be closed.

PR #46, opened 2019-05-09
CMake Header-Only Target

mjp41:
This PR creates a header only CMake target, which can be included by other projects without adding all the tests targets.


mjp41:
@nwf I am afraid this will conflict with your PR, but is easy to resolve. Sorry.

nwf:
@mjp41 No worries; thanks for the heads up.

PR #48, opened 2019-05-12
Add instructions on how to use the header-only library

ashamis:


README.md
```diff
@@ -92,6 +92,28 @@ This has a single build target `snmalloc_lib`, which includes
...
+
+Add these files to your CMake file
+```
+set(SNMALLOC_ONLY_HEADER_LIBRARY ON)
+add_subdirectory(snmalloc EXCLUDE_FROM_ALL)
+```
+
+In addition make sure you executable is compiled to support 128 bit atomic operations. This may require you to add the following to your CMake file.
+```
+target_compile_options([lib_name] PRIVATE -mcx16)
```
mjp41: So I think this should really be
```
target_link_libraries([my_applib] PUBLIC execinfo)
```
Or for a header only library you are adding snmalloc to 
```
target_link_libraries([my_applib] INTERFACE execinfo)
```
Did that not include the `-mcx16` setting on you subsequent target?

ashamis: I updated the PR

PR #49, opened 2019-05-13
Two minor performance improvements

mjp41:
* Modified TLS code to have the slow path in a separate function
* Switch some sanity checks to debug only

### DIFF
--- a/src/mem/alloc.h
+++ b/src/mem/alloc.h
@@ -872,7 +872,7 @@ namespace snmalloc
     inline void handle_message_queue()
     {
       // Inline the empty check, but not necessarily the full queue handling.
-      if (message_queue().is_empty())
+      if (likely(message_queue().is_empty()))
         return;
 
       handle_message_queue_inner();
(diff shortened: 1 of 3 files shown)


src/mem/alloc.h
```diff
@@ -872,7 +872,7 @@ namespace snmalloc
     inline void handle_message_queue()
     {
       // Inline the empty check, but not necessarily the full queue handling.
```
davidchisnall: Should we have an always inline attribute on this function?

src/mem/threadalloc.h
```diff
@@ -243,9 +252,24 @@ namespace snmalloc
...
+     */
+    static inline Alloc*& get()
+    {
+      Alloc*& per_thread = inner_get();
+
+      if (per_thread != nullptr)
+        return per_thread;
+
+      // Slow path that performs initialization
+      return inner_init();
```
davidchisnall: Do you want to put a cold attribute on this one?

Issue #52, opened 2019-05-14
slab free list not properly terminated?


nwf:
I've spent a while chasing what seems to be a bug in the slab state machine, but I'm not sure.

The case in question is that a slab has evolved, from its initial, empty state, to full (without `dealloc`s in the interim), to having one entry free and, so, on the free list and used as the inter-slab link.  In this latter state, an allocation request arrives to discover that the free list is populated, so it fetches what it believes to be the next value from the free entry and populates `meta.head` accordingly.  Unfortunately, because this entry was the slab link, that read has not produced a valid successor index for the free list, and so `Metaslab::debug_slab_invariant` correctly screams that the slab is full without satisfying `Metaslab::is_full`.

I believe the fix to have `Slab::alloc` check, when `(head & 1) == 0`, whether `meta.link` and `meta.head` are equal. If so, no read from the slab is necessary and `alloc` should immediately invoke `meta.set_full()` (possibly setting `meta.head = 1` beforehand to avoid the assert therein, or remove the assert).  If not, then it should read the successor `head` index as now.

Does that seem reasonable?  I don't fully understand, if this analysis is not wildly off the mark, why this isn't tickled in native builds, just CHERI, but tickled it is.

mjp41:
So there is a pretty hideous invariant around here. It is vaguely documented in 
> https://github.com/microsoft/snmalloc/blob/master/src/mem/metaslab.h#L48-L54

We have more explanation in the paper that we should be adding to here soon.  

If you are allocating from the slab, then it is the top of the stack.  This means that the previous pointer for the `SlabLink` and the next pointer for the freelist overlap.  By using a 64bit -1 for the previous pointer in the `SlabLink` terminator, and a 16bit -1 for the free list terminator, then it can be both at once.  Are the Cheri capability bits getting in the way here?

I have another design I have been working on in a branch, that doesn't require such a subtle invariant. I am just trying to work out the difference in performance before I PR it.

nwf:
Ah, I think I am now full of understanding.  Yes, CHERI stores the address after the metadata in memory, and `InvalidPointer<UINTPTR_MAX>` is, I suppose, an untagged 128 bit quantity consisting of 64 0 bits followed by 64 1 bits.  I'll see what I can do.

PR #53, opened 2019-05-16
Used start of a slab for link.

mjp41:


mjp41:
Build failure is not this PRs fault.  It appears the build pool is moving from gcc-7 to gcc-8? Our current CI builds with gcc-7.

mjp41:
@nwf This should help with Cheri from your comments in #52.

nwf:
LGTM.

PR #54, opened 2019-05-16
Made the statistics print atexit

mjp41:
Fixed some statistics and made them automatically print atexit.


mjp41:
This is failing CI due to something wrong with the build pool. I am going to force a merge.

PR #60, opened 2019-06-06
A few checks for easy corruptions

mjp41:
Will detect corruption caused by either

* Use-after-free
* Double-free

Neither is comprehensive.  Full temporal safety is not possible.
This just aids with debugging.

### DIFF
--- a/src/mem/metaslab.h
+++ b/src/mem/metaslab.h
@@ -92,27 +92,28 @@ namespace snmalloc
     }
 
     /// Value used to check for corruptions in a block
-    static constexpr size_t POISON = 
+    static constexpr size_t POISON =
       static_cast<size_t>(bits::is64() ? 0xDEADBEEFDEAD0000 : 0xDEAD0000);
 
     /// Store next pointer in a block. In Debug using magic value to detect some
     /// simple corruptions.
     static void store_next(void* p, uint16_t head)
-    { 
+    {
 #ifdef NDEBUG
       *static_cast<size_t*>(p) = head;
 #else
-      *static_cast<size_t*>(p) = head ^ POISON;
+      *static_cast<size_t*>(p) =
+        head ^ POISON ^ (static_cast<size_t>(head) << (bits::BITS - 16));
 #endif
     }
 
     /// Accessor function for the next pointer in a block.
-    /// In Debug checks for simple corruptions.    
+    /// In Debug checks for simple corruptions.
     static uint16_t follow_next(void* node)
-    {   
+    {
       size_t next = *static_cast<size_t*>(node);
 #ifndef NDEBUG
-      if ((next ^ POISON) > 0xFFFF)
+      if (((next ^ POISON) ^ (next << (bits::BITS - 16))) > 0xFFFF)
         error("Detected memory corruption.  Use-after-free.");
 #endif
       return static_cast<uint16_t>(next);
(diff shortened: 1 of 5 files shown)


src/mem/metaslab.h
```diff
@@ -91,6 +91,32 @@ namespace snmalloc
       return reinterpret_cast<SlabLink*>(pointer_offset(slab, link));
     }
 
+    /// Value used to check for corruptions in a block
+    static constexpr size_t POISON = 
+      static_cast<size_t>(bits::is64() ? 0xDEADBEEFDEAD0000 : 0xDEAD0000);
```
plietar: This won’t detect corruption in the low 16 bits. On low endian that’s the first two bytes of the object, which makes overwriting just those likely enough.
How about some sort of simple checksum? eg. repeat the two bytes in the remaining 6, and xor the poison.

src/mem/metaslab.h
```diff
@@ -122,8 +148,33 @@ namespace snmalloc
...
       assert(SLAB_SIZE > accounted_for);
 
-      // Walk bump-free-list-segment accounting for unused space
+      // Walk bump-free-list-segment checking for cycles.
+      // Using 
+      // https://en.wikipedia.org/wiki/Cycle_detection#Floyd's_Tortoise_and_Hare
+      // We don't expect a cycle, so worst case is only followed by a crash, so
+      // slow doesn't mater.
       uint16_t curr = head;
+      uint16_t curr_slow = head;
```
plietar: Could you extract this to a separate function

src/mem/metaslab.h
```diff
@@ -92,27 +92,28 @@ namespace snmalloc
...
-    static constexpr size_t POISON = 
+    static constexpr size_t POISON =
       static_cast<size_t>(bits::is64() ? 0xDEADBEEFDEAD0000 : 0xDEAD0000);
 
     /// Store next pointer in a block. In Debug using magic value to detect some
     /// simple corruptions.
     static void store_next(void* p, uint16_t head)
-    { 
+    {
 #ifdef NDEBUG
```
davidchisnall: It would be nice to make this a separate macro from `NDEBUG`, because some users may wish to use this in production and I may wish to debug the other code path at some point.

src/mem/metaslab.h
```diff
@@ -92,27 +92,28 @@ namespace snmalloc
...
     }
 
     /// Accessor function for the next pointer in a block.
-    /// In Debug checks for simple corruptions.    
+    /// In Debug checks for simple corruptions.
     static uint16_t follow_next(void* node)
-    {   
+    {
       size_t next = *static_cast<size_t*>(node);
 #ifndef NDEBUG
```
davidchisnall: As above.

src/mem/metaslab.h
davidchisnall: I think it would be cleaner to have an `add_poison` and a `check_and_remove_poison` method, then we could experiment with different poison mechanisms (ARM Pointer Authentication would be fun!)

mjp41: I am not sure on an API we want as this is currently restricted to 16-bit slab relative pointers.  Other mechanisms, would perhaps want different assumptions.  The code here factors out the functionality we need to provide in the free blocks. I think this is sufficient to make future experiments without much hassle.

Issue #58, opened 2019-05-20
Optimising small allocations


mjp41:
# Problem

The current design of **snmalloc** means that to perform a small allocation, we need to follow the following chain.

1. Follow `small_classes[sc]`, which points to the head of doubly linked list nodes for slabs with free space of this sizeclass.  
2. Calculate the meta-slab that is associated to this doubly linked list node. 
3. Read and modify the data in this meta-slab.

Based on profiling this may be costly.  It is certainly creating a fairly long chain of instructions to get to the allocation.

# Proposal 

We should investigate if we can keep more metadata in `small_classes[sc]`.  If we kept, a slab pointer, free list, and a used count. Then we could allocate directly from that first cache line.
This `slab` pointer to the **active block** could be independent of the `head` pointer to the `Slablink`.  

Deallocations, should go to the per slab meta data, and then free lists could be pulled from meta-slabs into the `small_classes[sc]`.  Effectively, the union of the `small_classes` data, and the meta slabs would form the actual free list. 

There are complications, the bump ptr would need handling specially.  If we separate the bump pointer and free list in the meta slab to be separate fields, then we could just pull free lists, and if we run out of space, bump allocate a page worth of free list values, to reduce the number of times we touch the metaslab on allocation.

A second complication is that we cannot determine if the active block is empty as the `used` count purely in the meta-slab is not enough.  If the meta-slab has a zero `used` we could, can check if this is the **active slab** and then move the used count from `small_classes[sc]` into the meta-slab. If it is still, zero used, then we need to return this slab, and update the **active slab**.

This issues is up-for-grabs as I don't have enough time to actually experiment with this.  If you want to have a go, please add a comment.

mjp41:
The ideas in this optimisation are inspired by [mimalloc](http://github.com/microsoft/mimalloc).  Maybe with them we can claw back some of the overhead on the purely sequential benchmarks.

mjp41:
I have prototyped this, and it has improved performance.  I will push a PR with this and many other performance fixes in the next few days.

PR #62, opened 2019-07-01
Optimise the fast path for alloc and dealloc

mjp41:
This seems to have a pretty drastic improvement in speed to heavy allocation benchmarks. It pushes `snmalloc` to have approximately the same performance as `mimalloc` on the `cfrac` benchmark.


mjp41:
This addresses #58

src/mem/sizeclass.h
```diff
@@ -4,35 +4,31 @@
...
-    return static_cast<uint8_t>(
+    return static_cast<sizeclass_t>(
       bits::to_exp_mant_const<INTERMEDIATE_BITS, MIN_ALLOC_BITS>(size));
   }
 
   constexpr static inline size_t large_sizeclass_to_size(uint8_t large_class)
   {
     return bits::one_at_bit(large_class + SUPERSLAB_BITS);
-  }
+  }
```
davidchisnall: Space.

src/mem/sizeclass.h
```diff
@@ -4,35 +4,31 @@
...
-  constexpr static inline uint8_t size_to_sizeclass_const(size_t size)
+
+  constexpr static inline sizeclass_t size_to_sizeclass_const(size_t size)
   {
     // Don't use sizeclasses that are not a multiple of the alignment.
     // For example, 24 byte allocations can be
     // problematic for some data due to alignment issues.
-    return static_cast<uint8_t>(
+    return static_cast<sizeclass_t>(
       bits::to_exp_mant_const<INTERMEDIATE_BITS, MIN_ALLOC_BITS>(size));
```
davidchisnall: Do we want an assert that this is in range for a `uint8_t`?

src/mem/sizeclasstable.h
```diff
@@ -5,31 +5,49 @@
...
 namespace snmalloc
 {
+  constexpr size_t PTR_BITS = bits::next_pow2_bits_const(sizeof(void*));
+
+  constexpr static size_t sizeclass_lookup_index(size_t s)
+  {
+    // We shift by PTR_BITS as makes code-gen for the table lookup nicer.
+    // We could shift by MIN_ALLOC_BITS, but then there is a slightly more
+    // complex sequence.
+    return (s - 1) >> PTR_BITS;
```
davidchisnall: I think I was more confused after reading this comment than before.  I guess the key point is that we can index the table by anything we like that's cheap to compute, and this particular shift makes it cheap, at the expense of a larger table?

src/mem/sizeclass.h
mjp41: Fix clangformat to actually run on builds, so this should be fixed if it passes CI now.

PR #64, opened 2019-07-05
Lazy tls

davidchisnall:
Lazily initialise the TLS pointer on slow paths, so that we can avoid a check that it's initialised on the fast paths.  We now statically set the TLS pointer to the address of a global allocator, which does not own any memory.  When we try to allocate from it, we hit a slow path and there we check if it's actually the global allocator (in which case we allocate a real allocator and store a pointer to it in TLS).

This trims one branch from the fast path in `malloc`.

### DIFF
--- a/src/mem/alloc.h
+++ b/src/mem/alloc.h
@@ -234,6 +234,11 @@ namespace snmalloc
     FastFreeLists() : small_fast_free_lists() {}
   };
 
+  ALWAYSINLINE inline void* no_replacement(void*)
+  {
+    return nullptr;
+  }
+
   /**
    * Allocator.  This class is parameterised on three template parameters.  The
    * `MemoryProvider` defines the source of memory for this allocator.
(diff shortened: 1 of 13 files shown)


src/mem/alloc.h
```diff
@@ -1317,17 +1370,36 @@ namespace snmalloc
...
 
       void* offseted = apply_cache_friendly_offset(p, sizeclass);
 
+      // Check whether this will overflow the cache first.  If we are a fake
+      // allocator, then our cache will always be full and so we will never hit
+      // this path.
+      size_t sz = sizeclass_to_size(sizeclass);
+      if ((remote.size + sz) < REMOTE_CACHE)
+      {
+        stats().remote_free(sizeclass);
```
mjp41: Add assert that we are not fake at this point!

mjp41: Perhaps put something similar in other places where we don't expect to see the fake allocator.

PR #68, opened 2019-07-09
Minor opts

mjp41:
* Move to using FlatPageMap by default on platforms, which lazily commit memory.
* Minor changes to inlining behaviour for better codegen.

### DIFF
--- a/src/mem/alloc.h
+++ b/src/mem/alloc.h
@@ -46,8 +46,7 @@ namespace snmalloc
 // Use flat map is under a single node.
 #  define SNMALLOC_MAX_FLATPAGEMAP_SIZE PAGEMAP_NODE_SIZE
 #endif
-  static constexpr bool USE_FLATPAGEMAP =
-    (Pal::pal_features & PalFeatures::LazyCommit) ||
+  static constexpr bool USE_FLATPAGEMAP = pal_supports<LazyCommit>() ||
     (SNMALLOC_MAX_FLATPAGEMAP_SIZE >=
      sizeof(FlatPagemap<SUPERSLAB_BITS, uint8_t>));
 
@@ -1209,9 +1208,7 @@ namespace snmalloc
           else if constexpr (decommit_strategy == DecommitSuperLazy)
           {
             static_assert(
-              std::remove_reference_t<decltype(
-                large_allocator.memory_provider)>::
-                template pal_supports<LowMemoryNotification>(),
+              pal_supports<LowMemoryNotification, MemoryProvider>(),
               "A lazy decommit strategy cannot be implemented on platforms "
               "without low memory notifications");
           }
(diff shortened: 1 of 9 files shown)


src/mem/alloc.h
```diff
@@ -46,8 +46,10 @@ namespace snmalloc
 // Use flat map is under a single node.
 #  define SNMALLOC_MAX_FLATPAGEMAP_SIZE PAGEMAP_NODE_SIZE
 #endif
-  static constexpr bool USE_FLATPAGEMAP = SNMALLOC_MAX_FLATPAGEMAP_SIZE >=
-    sizeof(FlatPagemap<SUPERSLAB_BITS, uint8_t>);
+  static constexpr bool USE_FLATPAGEMAP =
+    (Pal::pal_features & PalFeatures::LazyCommit) ||
```
davidchisnall: `largealloc.h` has a typesafe version of this test (`pal_supports()`).  It would be good to reuse that if possible, moving it somewhere more useful if necessary.

src/mem/alloc.h
```diff
@@ -1366,15 +1370,8 @@ namespace snmalloc
...
-#if defined(__GNUC__) && !defined(__clang__) && !defined(__OPTIMIZE__)
-    // Don't force this to be always inlined in debug builds with GCC, because
-    // it will fail and then raise an error.
-    inline
-#else
-    SNMALLOC_FAST_PATH
-#endif
-      void
-      remote_dealloc(RemoteAllocator* target, void* p, sizeclass_t sizeclass)
+    SNMALLOC_SLOW_PATH
```
davidchisnall: Please add a comment explaining why moving this to the slow path is a good idea, or in 6 months someone is likely to think it should be fast path.

src/pal/pal_consts.h
```diff
@@ -25,7 +25,13 @@ namespace snmalloc
      * a size and alignment.  A PAL that does *not* support it must expose a
      * `request()` method that takes only a size.
      */
-    AlignedAllocation = (1 << 1)
+    AlignedAllocation = (1 << 1),
+    /**
+     * This PAL natively supports lazy commit of pages. This means have large
+     * allocations and not touching them does not up memory usage.  This is
```
davidchisnall: up is not a verb!

src/pal/pal_consts.h
```diff
@@ -25,7 +25,13 @@ namespace snmalloc
...
+     * exposed in the P
```
davidchisnall: In the P?

azure-pipelines.yml
```diff
@@ -124,7 +124,7 @@ jobs:
       solution: build/snmalloc.sln
       msbuildArguments: '/m /p:Configuration=$(BuildType)'
 
-  - script: 'ctest -j 4 --interactive-debug-mode 0 --output-on-failure -C $(BuildType)'
+  - script: 'ctest -j 1 --interactive-debug-mode 0 --output-on-failure -C $(BuildType)'
```
davidchisnall: This change now causes a conflict.  I turned this down to -j2 on the Win8Compat variant but left it at -j4 elsewhere.  It seemed to make the tests happy again...

PR #71, opened 2019-07-15
Improved ThreadAlloc::get API

mjp41:
Made the API so that get always returns an initialised Alloc*.  Added
new fast path that doesn't perform checking, but can lead to very slow
behaviour if called and reused.

### DIFF
--- a/src/mem/threadalloc.h
+++ b/src/mem/threadalloc.h
@@ -127,10 +127,11 @@ namespace snmalloc
      */
     static void inner_release()
     {
-      if (get() != &GlobalPlaceHolder)
+      Alloc*& a = get_noncachable();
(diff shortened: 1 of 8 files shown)


src/mem/threadalloc.h
```diff
@@ -127,10 +127,11 @@ namespace snmalloc
      */
     static void inner_release()
     {
-      if (get() != &GlobalPlaceHolder)
+      Alloc*& a = get_noncachable();
```
davidchisnall: Should `get_noncacheable` return a reference?  It might be better to make it return a pointer and have a `get_reference` that is `private` and returns a reference to the `thread_local` thing.

PR #75, opened 2019-08-01
[NFC] Remove some code duplication in the PALs.

davidchisnall:
Pull out a generic POSIX PAL as a superclass for the Linux and generic
BSD PALs.  Now we have FreeBSD and Linux adding OS-specific behaviour,
OpenBSD as a named subclass of the generic BSD PAL that doesn't add any
behaviour.

I believe a NetBSD PAL should now be identical to the OpenBSD one -
patches welcome if anyone wants to test one!


davidchisnall:
@devnexen, this cleans up some of the code duplication between PALs.  Please can you check that it hasn't broken OpenBSD?

src/pal/pal_openbsd.h
```diff
@@ -11,10 +11,23 @@
...
   public:
-    static constexpr uint64_t pal_features = LazyCommit;
+    /**
+     * The features exported by this PAL.
+     *
+     * Currently, these are identical to the generic BSD PAL.  This field is
+     * declared explicitly to remind anyone who modifies this class that they
+     * should add any required features.
+     */
+    static constexpr uint64_t pal_features = PALBSD<PALOBSD>;
```
devnexen: Little mistake here should be `PALPOSIX<OS>::pal_features` or `PALBSD<OS>::pal_features`

davidchisnall: Just saw that while refactoring the Apple PAL.  Should be fixed now.

davidchisnall:
This refactoring removes 77 lines of code (and adds 104 lines of comments).

src/pal/pal_openbsd.h
devnexen: Just a little idea if you wish for APPLE, you can either choose calling raw vm_allocate/vm_deallocate or even keeping mmap wrapper but adding VM_MAKE_FLAGS for the file descriptor arguments ... practical to monitor anonymous pages.

PR #77, opened 2019-08-01
NetBSD support

davidchisnall:
@krytarowski, can you test this one please?

### DIFF
--- a/src/pal/pal_bsd_aligned.h
+++ b/src/pal/pal_bsd_aligned.h
@@ -1,7 +1,6 @@
 #pragma once
 
-#if defined(__FreeBSD__) && !defined(_KERNEL)
-#  include "pal_bsd.h"
+#include "pal_bsd.h"
 
 namespace snmalloc
 {
@@ -52,4 +51,3 @@ namespace snmalloc
     }
   };
 } // namespace snmalloc
-#endif


krytarowski:
```
$ make
[  1%] Linking CXX executable func-two_alloc_types-16
ld: cannot find -latomic
*** Error code 1
```

davidchisnall:
If you remove this line from the CMakeLists.txt, does it work?

```
target_link_libraries(snmalloc_lib INTERFACE atomic)
```

krytarowski:
```
ld: CMakeFiles/func-release-rounding-16.dir/src/test/func/release-rounding/rounding.cc.o: in function `std::atomic<snmalloc::ABA<snmalloc::Allocator<snmalloc::MemoryProviderStateMixin<snmalloc::PALNetBSD>, snmalloc::SuperslabMap<snmalloc::GlobalPagemap>, true, &snmalloc::lazy_replacement>, (snmalloc::Construction)0>::Linked>::compare_exchange_weak(snmalloc::ABA<snmalloc::Allocator<snmalloc::MemoryProviderStateMixin<snmalloc::PALNetBSD>, snmalloc::SuperslabMap<snmalloc::GlobalPagemap>, true, &snmalloc::lazy_replacement>, (snmalloc::Construction)0>::Linked&, snmalloc::ABA<snmalloc::Allocator<snmalloc::MemoryProviderStateMixin<snmalloc::PALNetBSD>, snmalloc::SuperslabMap<snmalloc::GlobalPagemap>, true, &snmalloc::lazy_replacement>, (snmalloc::Construction)0>::Linked, std::memory_order, std::memory_order)':
/usr/include/g++/atomic:291: undefined reference to `__atomic_compare_exchange_16'
ld: CMakeFiles/func-release-rounding-16.dir/src/test/func/release-rounding/rounding.cc.o: in function `std::atomic<snmalloc::ABA<snmalloc::Largeslab, (snmalloc::Construction)0>::Linked>::compare_exchange_weak(snmalloc::ABA<snmalloc::Largeslab, (snmalloc::Construction)0>::Linked&, snmalloc::ABA<snmalloc::Largeslab, (snmalloc::Construction)0>::Linked, std::memory_order, std::memory_order)':
/usr/include/g++/atomic:291: undefined reference to `__atomic_compare_exchange_16'
```

krytarowski:
I was told that NetBSD deliberately skips libatomic.

davidchisnall:
I'm a bit surprised that you're seeing this, if the `-mcx16` flag is being passed to the compiler.  Can you check that this is the case?  What version of clang are you using?

krytarowski:
```
$ gcc -v
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/libexec/lto-wrapper
Target: x86_64--netbsd
Configured with: /usr/src/tools/gcc/../../external/gpl3/gcc/dist/configure --target=x86_64--netbsd --enable-long-long --enable-threads --with-bugurl=http://www.NetBSD.org/Misc/send-pr.html --with-pkgversion='NetBSD nb3 20190319' --with-system-zlib --without-isl --enable-__cxa_atexit --enable-libstdcxx-time=rt --enable-libstdcxx-threads --with-diagnostics-color=auto-if-env --with-tune=nocona --with-default-libstdcxx-abi=new --with-mpc-lib=/var/obj/mknative/amd64-x86_64/usr/src/external/lgpl3/mpc/lib/libmpc --with-mpfr-lib=/var/obj/mknative/amd64-x86_64/usr/src/external/lgpl3/mpfr/lib/libmpfr --with-gmp-lib=/var/obj/mknative/amd64-x86_64/usr/src/external/lgpl3/gmp/lib/libgmp --with-mpc-include=/usr/src/external/lgpl3/mpc/dist/src --with-mpfr-include=/usr/src/external/lgpl3/mpfr/dist/src --with-gmp-include=/usr/src/external/lgpl3/gmp/lib/libgmp/arch/x86_64 --enable-tls --disable-multilib --disable-libstdcxx-pch --build=x86_64--netbsd --host=x86_64--netbsd --with-sysroot=/var/obj/mknative/amd64-x86_64/usr/src/destdir.amd64
Thread model: posix
gcc version 7.4.0 (nb3 20190319)
```

davidchisnall:
It looks as if the `-mcx16` is being ignored with gcc7 (even though it's a documented flag).  Your choice is either link libatomic or compile with clang.

krytarowski:
with clang builds until completion.

krytarowski:
But please blacklist libatomic on NetBSD.

davidchisnall:
I don't see any advantage in doing that.

- If someone is building with upstream gcc on NetBSD, they will have libatomic and the build will work if `-latomic` is passed from the build system.
- If they are building with NetBSD's packaged gcc then they will either get a linker failure from not finding libatomic or they will get a linker failure from not finding the symbols that libatomic provides.  

At least trying `-latomic` makes it work for some people and gives the rest an idea of why it's broken.  Why does NetBSD not ship libatomic?

davidchisnall:
(Does the test suite now pass on NetBSD?)

krytarowski:
The developers disagree with libatomic idea and reject it.

libatomic can be at least automatically detected with CMake.

Insisting on it, will keep NetBSD broken with both gcc and clang.

davidchisnall:
If the NetBSD developers disagree with the libatomic idea, what is their alternative?  Or do they just decide not to support C11 / C++11?  I'll happily review a patch to properly support NetBSD (which means linking whatever the NetBSD equivalent of libatomic is), but otherwise it's probably better to remove the NetBSD PAL rather than leave something that only works in non-default configurations.

krytarowski:
The alternative is to rely on hw features for 128bit atomic operations - recent x86_64 can do that. Optionally redo the code and switch/fallback to something else like futexes. We have got an ongoing work on futexes, but it will probably land NetBSD 10.

krytarowski:
C11/C++11 atomics are supported.

krytarowski:
Independently, I will try to check whether this gcc -mcx16 is a local bug or GCC issue.

davidchisnall:
We only need 128-bit atomics, but it sounds as if the compiler that NetBSD ships by default does not emit 128-bit atomics even with the `-mcx16` flag.  There are three reasonable approaches here:

- Ship libatomics.
- Ship another library that provides these symbols (e.g. put them in libc).
- Ship a compiler that implements `_Atomic` in some other way and does not depend on the `__atomic_compare_exchange_{16,n}` symbols.

Shipping a compiler that emits calls to support library functions and then not shipping the support library is the worst possible solution.  The required CMake changes on our side would be quite complex unless we just say reject trying to compile with gcc on NetBSD entirely.

krytarowski:
I propose the following approach for the time being:

 - detect libatomic, if it is available link with it
 - fallback to hw atomics and detect for x86_64 whether -mcx16 is functional, if so pick it
 - error with a verbose message to pick another compiler

This way all users on NetBSD and other platforms will be aware to check with clang.

krytarowski:
It looks like a GCC specific issue:

```
       -mcx16
           This option enables GCC to generate "CMPXCHG16B" instructions in
           64-bit code to implement compare-and-exchange operations on 16-byte
           aligned 128-bit objects.  This is useful for atomic updates of data
           structures exceeding one machine word in size.  The compiler uses
           this instruction to implement __sync Builtins.  However, for
           __atomic Builtins operating on 128-bit integers, a library call is
           always used.
```

library call -> libatomic

The same happens on Linux.

krytarowski:
It looks like for GCC users there is an option to code `LOCK CMPXCHG16B` in assembly.

davidchisnall:
We only add libatomic when compiling with gcc, so I think the current logic is correct.  If NetBSD ever ships a version of GCC that actually works, then we can revisit this, otherwise NetBSD users will get an error when they try to link with GCC but have everything work when they compile with clang.

krytarowski:
I'm working on libatomic. I'm trying to port the LLVM runtime to GCC+Clang.

https://reviews.llvm.org/D79713

Issue #82, opened 2019-08-03
Lazy TLS initialization breaks stats counting


nwf:
Bisection reveals that b8a5d7fca96fd442eb74df8db59666bd1704696c breaks the idea that stat counters don't go negative.  In particular, adding `assert(sizeclass[sc].count.current > 0);` to `sizeclass_dealloc` (and applying the fix for #79) will cause at least `func-malloc` to fail on that commit and  after.  The problem is that `small_alloc`'s call to `stats().sizeclass_alloc` happen before the check that we're on the fast path, and so apply to the stub global allocator's stat structure.  This does not appear to be a problem in `medium_alloc` or `large_alloc`, as the replacement path will have tail-called before attempting to update the stub's stats structure.

mjp41:
Thanks for reporting this.  I see you have a commit in your fork to fix this.  Have you managed to get the University to approve the CLA?  If so, can you PR it in?

mjp41:
Resolved by #84

Issue #81, opened 2019-08-03
func-fixed_region is not built correctly


nwf:
When running revions after e240dd279a0c40247abd1324604eb58be6cf620c with `-O2` and `USE_SNMALLOC_STATS` enabled and on a Linux amd64 box, `func-fixed_region-16` crashes (but `func-fixed_region-1` does not).  The tail of the output under `gdb` is

```
Allocated object 0x7fffea02b530
Allocated object 0x7fffea02b5a0
GlobalStats, DumpID, AllocatorID, Remote freed, Remote posted, Remote received, Superslab pop, Superslab push, Superslab fresh, Segments
BucketedStats, DumpID, AllocatorID, Size group, Size, Current count, Max count, Total Allocs, Current Slab bytes, Max Slab bytes, Total slab allocs, Average Slab Usage, Average wasted space

Program received signal SIGSEGV, Segmentation fault.
snmalloc::FlatPagemap<24ul, unsigned char>::get (p=140737119453312, this=<optimised out>)
    at /home/nwf20/cheri/mainline/snmalloc/src/override/../mem/pagemap.h:360
360	      return top[p >> SHIFT].load(std::memory_order_relaxed);

```

The backtrace is... well, fairly full of POSIX and C++ dragons:

```
#0  snmalloc::FlatPagemap<24ul, unsigned char>::get (p=140737119453312, this=<optimised out>)
    at /home/nwf20/cheri/mainline/snmalloc/src/override/../mem/pagemap.h:360
#1  snmalloc::SuperslabMap<snmalloc::GlobalPagemap>::get (p=140737119453312, this=<optimised out>)
    at /home/nwf20/cheri/mainline/snmalloc/src/override/../mem/alloc.h:135
#2  snmalloc::Allocator<snmalloc::MemoryProviderStateMixin<snmalloc::PALLinux>, snmalloc::SuperslabMap<snmalloc::GlobalPagemap>, true, &snmalloc::lazy_replacement>::dealloc (this=0x7fffe8004000, p=0x7fffea030080)
    at /home/nwf20/cheri/mainline/snmalloc/src/override/../mem/alloc.h:465
#3  operator delete (p=0x7fffea030080) at /home/nwf20/cheri/mainline/snmalloc/src/override/new.cc:41
#4  0x0000000000403451 in __gnu_cxx::new_allocator<char>::deallocate (__p=0x7fffea030080 "Average wasted space", 
[...]

Issue #80, opened 2019-08-03
func-malloc test leaks objects


nwf:
`func-malloc` test has always leaked objects since its inception; its
`test_realloc(our_malloc(size), (size_t)-1, ENOMEM, true);` drops the allocation on the floor.

mjp41:
Resolved by #84

Issue #79, opened 2019-08-03
debug_check_empty useless after message queue changes


nwf:
Bisecting reveals that eb4e28e8d0c9ee37096771728b90c5e364548e8b, while it does remove the special case, breaks tests like `func-memory` that expect to be able to test that there are no outstanding allocations, becuase every allocator will be nonempty, having message queue stubs in place.  The fix, I think, is to move the test for emptiness into the main loop, testing for emptiness between `alloc->message_queue().destroy()` and `alloc->init_message_queue()`; the count must be reset to zero at the top of the main loop.

mjp41:
Thanks for reporting this. I see you have a commit in your fork to fix this. Have you managed to get the University to approve the CLA? If so, can you PR it in?

nwf:
I have not, which is why I didn't open a PR.  Apparently the entire CTSRD project is stuck in legal limbo for reasons that escape me.

mjp41:
Looking at this, I think there are more issues introduced by this change because of the allocator local free lists.  I will create some tests to more properly check this. And address these issues.

mjp41:
Resolved by #84

PR #84, opened 2019-08-13
Bug fix

mjp41:
This is addressing several of the bugs reported by @nwf in #79 #80 #81 #82.


davidchisnall:
Looks as if it's made things sufficiently slow in debug build that 64-bit Windows times out on one of the tests...

mjp41:
> Looks as if it's made things sufficiently slow in debug build that 64-bit Windows times out on one of the tests...

I think it has posted a Dialog box, I am just modifying the tests to change the abort behaviour to not do that. So I can see what the error is, as it completes super quick locally for that build (although slightly different version of MSVC)

PR #88, opened 2019-08-14
CI testing help

mjp41:
This is a pretty dirty hack to catch every exception and print a stack trace for the tests.

It doesn't correctly handle issues if the crash is inside a CRT allocation, but the tests don't really do any allocations with the CRT's allocator, so should be good enough for debugging in CI.


src/test/setup.h
```diff
@@ -1,19 +1,95 @@
...
 #  include <ds/bits.h>
+#  include <iostream>
+#  include <pal/pal.h>
 #  include <signal.h>
 #  include <stdlib.h>
+// Has to come after the PAL.
+#  include <DbgHelp.h>
+#  pragma comment(lib, "dbghelp.lib")
+
+void print_stack_trace()
```
davidchisnall: Do we want to export this in the PAL as the `PAL::error()` thing, and then just call that from the tests?  That would let us produce a stack trace from any PAL that supports it, from any of the checked error conditions.

src/test/setup.h
```diff
@@ -1,19 +1,95 @@
...
+#  define CALL_LAST 0
+LONG WINAPI VectoredHandler(struct _EXCEPTION_POINTERS* ExceptionInfo)
+{
+  UNUSED(ExceptionInfo);
+
+  puts("*****UNHANDLED EXCEPTION******");
+
+  print_stack_trace();
+
   _exit(1);
```
davidchisnall: More generally, do these want to call `DefaultPAL::error()`?

src/test/setup.h
mjp41: That makes sense.  We would need to stop calling abort().  Also, assert failures call abort.  So perhaps we need to override assert as well.  I'll create an issue for this, as I think it needs more discussion and design than I have time for at the moment.

mjp41: #89

src/test/setup.h
mjp41: Yes.  This should be part of the #89 issue.

PR #92, opened 2019-08-26
darwin: override zero/non zeroed pages to tag them.

devnexen:


mjp41:
@devnexen, could you run the clangformat build target to fix the formatting of your commit, i.e. `make clangformat`, `ninja clangformat`,...

src/pal/pal_apple.h
```diff
@@ -22,6 +22,56 @@ namespace snmalloc
...
+  private:
+    /**
+     * Anonymous page tag ID
+     *
+     * Darwin platform allows to gives an ID to anonymous pages
+     * from 240 up to 255 are guaranteed to be free of usage
+     * however eventually a lower could be taken (e.g. LLVM sanitizers
+     * has 99) so we can monitor their states via vmmap for instance.
+     */
+    int pal_anon_id = 241 << 24;
```
davidchisnall: I don't see this anywhere in the [macOS `mmap` man page](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/mmap.2.html).  I'm very hesitant to commit something that depends on undocumented features of the macOS system call interface, because Apple reserves the right to change these (and does) across minor OS releases.

mjp41: Should you use the macro `VM_MAKE_TAG` here, and include the relevant header?

devnexen: It is VM_MAKE_TAG macro in fact.

devnexen: I just answer above yes I can do that instead if that please you better.

davidchisnall: I see, yes please use the `VM_MAKE_TAG` macro and update the comment.

devnexen:
> @devnexen, could you run the clangformat build target to fix the formatting of your commit, i.e. `make clangformat`, `ninja clangformat`,...

Sorry saw it just now ... doing it.

src/pal/pal_apple.h
```diff
@@ -22,6 +24,58 @@ namespace snmalloc
      * should add any required features.
      */
     static constexpr uint64_t pal_features = PALBSD::pal_features;
+
+    // OS specific function for zeroing memory with ID 241.
```
davidchisnall: Please make this a proper doc comment, don't mention the magic number, do say what the magic number represents.

src/pal/pal_apple.h
```diff
@@ -22,6 +24,58 @@ namespace snmalloc
...
+          0);
+
+        if (r != MAP_FAILED)
+          return;
+      }
+
+      bzero(p, size);
+    }
+
+    // Reserve memory with ID 241.
```
davidchisnall: Same comment for this comment.

src/pal/pal_apple.h
```diff
@@ -22,6 +24,58 @@ namespace snmalloc
...
+    /**
+     * Anonymous page tag ID
+     *
+     * Darwin platform allows to gives an ID to anonymous pages via
+     * the VM_MAKE_TAG's macro, from 240 up to 255 are guaranteed
+     * to be free of usage, however eventually a lower could be taken
+     * (e.g. LLVM sanitizers has 99) so we can monitor their states
+     * via vmmap for instance.
+     */
+    int pal_anon_id = VM_MAKE_TAG(241);
```
davidchisnall: What other things use this facility?  Are we likely to conflict with other application usage?  Should this be a template parameter on the PAL with a default, so users can instantiate the PAL with a different value?

davidchisnall: This should be a `static const int`, not a variable field (though I think I'd prefer that it were a template parameter with a default of 241).

devnexen: That is the purpose of using ID from 240 to 255, no risk of conflicts, it is not forbidden to use lower IDs but then you would be right.

davidchisnall: As I understand it, 240-255 are reserved for non-Apple uses, but that doesn't mean that they won't have conflicts if user software is already using 241, or links another framework that uses 241.

src/pal/pal.h
```diff
@@ -27,7 +27,7 @@ namespace snmalloc
 #  if defined(_WIN32)
     PALWindows;
 #  elif defined(__APPLE__)
-    PALApple;
+    PALApple<PALAnonDefaultID>;
```
davidchisnall: No need to explicitly provide the default template value here.

src/pal/pal_apple.h
```diff
@@ -13,7 +13,8 @@ namespace snmalloc
    * XNU behaves exactly like a generic BSD platform but this class exists
    * as a place to add XNU-specific behaviour later, if required.
```
mjp41: This is no longer a placeholder as it adds specific behaviour.

PR #97, opened 2019-11-14
Memory leak

mjp41:
This PR does
* Fix a memory leak during allocator creation, if the first allocation is the same size as the message queue stub;
* Provides a more resilient debug_check_empty, that does not depend on statistics being enabled.

### DIFF
--- a/src/mem/alloc.h
+++ b/src/mem/alloc.h
@@ -911,7 +911,7 @@ namespace snmalloc
         while (prev != nullptr)
         {
           auto n = Metaslab::follow_next(prev);
-          dealloc(prev);
+          dealloc(remove_cache_friendly_offset(prev, i));
           prev = n;
         }
 
--- a/src/mem/alloc.h
+++ b/src/mem/alloc.h
@@ -917,9 +917,9 @@ namespace snmalloc
         test(small_classes[i].is_empty());
       }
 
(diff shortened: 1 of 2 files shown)


src/mem/alloc.h
```diff
@@ -875,6 +875,60 @@ namespace snmalloc
 #endif
     }
 
+    /**
+     * If result parameter is non-null, then is_empty is passsed into the
```
davidchisnall: What is `is_empty`?

src/mem/alloc.h
```diff
@@ -875,6 +875,60 @@ namespace snmalloc
...
+     * the location pointed to by result.
+     *
+     * Otherwise, asserts that it is empty.
```
davidchisnall: 'it' is this allocator?  All allocators?

src/mem/alloc.h
```diff
@@ -875,6 +875,60 @@ namespace snmalloc
...
+     **/
+    void debug_is_empty(bool* result)
+    {
+      auto test = [&result](bool value) {
```
davidchisnall: Rather than making this `bool value`, it would probably be better to make it `auto &allocator` and then do the `.is_empty()` call here and be able to report `&allocator` in the error message (or can we not do printf-like things here?).

mjp41: The Pal doesn't support printf like things, yet.

Issue #98, opened 2019-11-14
Statistics underflow


nwf:
https://github.com/microsoft/snmalloc/commit/c2799f6ec8ed43b87d520d58ff7c52e57a28a2fb#diff-a8ebb8308aa3e85926e29a0e7e02f444R913 calls `dealloc` on objects sourced from the free list.  This seems kind of dubious to me but at the very least causes the statistics counters to underflow, since it deallocates an object not accounted for.

mjp41:
Good catch thanks.

PR #99, opened 2019-11-14
Fix stats underflow

nwf:
A proposed fix to #98

### DIFF
--- a/src/mem/allocstats.h
+++ b/src/mem/allocstats.h
@@ -34,6 +34,7 @@ namespace snmalloc
 
       void dec()
       {
+        assert(current > 0);
         current--;
       }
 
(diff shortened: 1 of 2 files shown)


src/mem/allocstats.h
```diff
@@ -183,6 +183,7 @@ namespace snmalloc
       UNUSED(sc);
 
 #ifdef USE_SNMALLOC_STATS
+      assert(sizeclass[sc].count.current > 0);
```
mjp41: Can't `current` become zero?  Should we make the statistics use a signed representation, and check for below 0?  Or perhaps, we could add an assert into `dec` that it is above 0?

nwf: This check is performed before the decrement, and so still permits `current` to become zero.  But yes, I agree that inside `dec()` is probably clearer; will move.

mjp41:
Do you have a private test that caught this?  If so, it would be great to push it in, so I don't break stats in the future.

nwf:
With the assertions in place, this was tripped by func-malloc.

PR #102, opened 2019-11-20
Code quality for `free`

mjp41:
These two changes reduce the fast path for deallocation from 34 instructions on Linux with Clang to 25 instructions.  There are two key changes
* When a slab is full, set its 'used' count to 1.  This allows the detecting leaving full, and entering empty to be the same check.
* Expand the meta-data to contain a full pointer to the free list, not a relative 16 bit pointer.  This reduces the arithmetic required on the fast path.

### DIFF
--- a/src/mem/metaslab.h
+++ b/src/mem/metaslab.h
@@ -30,22 +30,14 @@ namespace snmalloc
   class Metaslab
   {
   public:
+    // Ppinter to first free entry in this slab
+    void* head;
+
     // How many entries are not in the free list of  slab.
     uint16_t used = 0;
 
     // How many entries have been allocated from this slab.
     uint16_t allocated;
-
-    // Index of first entry in this slab that forms the free
-    // list.  The list entries are stored as the first pointer
-    // in each unused object. The terminator is a pointer or
-    // offset into the block with the bottom bit set.  This means
-    //  I.e.
-    //    * an empty list has a head of 1.
-    //    * a one element list has an head contains an offset to this
-    //       this block, and then contains a pointer with the bottom
-    //       bit set.
-    Mod<SLAB_SIZE, uint16_t> head;
 
     // When a slab has free space it will be on the has space list for
     // that size class.  We use an empty block in this slab to be the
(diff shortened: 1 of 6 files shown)


mjp41:
@nwf Any comments on these changes and CHERI would be appreciated.

src/mem/metaslab.h
```diff
@@ -30,22 +30,28 @@ namespace snmalloc
   class Metaslab
   {
   public:
-    // How many entries are not in the free list of  slab.
-    uint16_t used = 0;
+    /**
+     *  Pointer to first free entry in this slab
+     * 
+     *  The list will (allocated - needed - 1) long. The -1 is
```
sylvanc: will -> will be

src/mem/metaslab.h
```diff
@@ -122,16 +130,19 @@ namespace snmalloc
...
       size_t size = sizeclass_to_size(sizeclass);
-      size_t slab_start = get_initial_offset(sizeclass, is_short);
-      size_t all_high_bits = ~static_cast<size_t>(1);
+      size_t slab_end = (address_cast(head) | ~SLAB_MASK) + 1;
+      uintptr_t allocation_start =
+        remove_cache_friendly_offset(address_cast(head), sizeclass);
 
-      size_t head_start =
-        remove_cache_friendly_offset(head & all_high_bits, sizeclass);
+      return (slab_end - allocation_start) % size == 0;
```
sylvanc: I was worried, but this is debug only code, phew.

src/mem/slab.h
```diff
@@ -163,27 +159,19 @@ namespace snmalloc
...
-      meta.head = current;
-      assert(meta.valid_head(is_short()));
+      meta.head = p;
+      assert(meta.valid_head());
 
       // Set the next pointer to the previous head.
-      Metaslab::store_next(
-        p, (head == 1) ? nullptr : pointer_offset(this, head));
+      Metaslab::store_next(p, head);
       meta.debug_slab_invariant(is_short(), this);
```
sylvanc: Does this elide `is_short()` in the NDEBUG case when compiled? Looks like it should, just double checking.

mjp41: Pretty hard to tell given all the inlining. I have rewritten to push the is_short() inside the function, and thus inside the ifndef.

PR #103, opened 2019-11-25
CHERI prep work that's just tidying

nwf:
Nothing very CHERI-specific here, I think.  If there is, let me know and I'll drop it. :)


src/ds/address.h
```diff
@@ -39,4 +42,80 @@ namespace snmalloc
...
   }
+
+  /**
+   * Test if a pointer is aligned to a given size, which must be a power of
+   * two.
+   */
+  template<size_t alignment>
+  static inline bool is_aligned_block(void* p, size_t size)
+  {
+    assert(bits::next_pow2(alignment) == alignment);
```
mjp41: Can this be a `static_assert`?

src/ds/address.h
```diff
@@ -39,4 +42,80 @@ namespace snmalloc
...
+
+  /**
+   * Align a pointer down to a statically specified granularity, which must be a
+   * power of two.
+   */
+  template<size_t granule, typename T = void>
+  inline T* pointer_align_down(void* p)
+  {
+    static_assert(granule > 0);
+    static_assert((granule & (granule - 1)) == 0);
```
mjp41: `next_pow2` like above?

src/ds/address.h
```diff
@@ -39,4 +42,80 @@ namespace snmalloc
...
+
+  /**
+   * Align a pointer up to a statically specified granularity, which must be a
+   * power of two.
+   */
+  template<size_t granule, typename T = void>
+  inline T* pointer_align_up(void* p)
+  {
+    static_assert(granule > 0);
+    static_assert((granule & (granule - 1)) == 0);
```
mjp41: `next_pow2` is clearer.

src/ds/address.h
```diff
@@ -39,4 +42,80 @@ namespace snmalloc
   {
     return reinterpret_cast<T*>(address);
   }
+
+  /**
+   * Test if a pointer is aligned to a given size, which must be a power of
+   * two.
+   */
+  template<size_t alignment>
```
mjp41: `granule` vs `alignment` as the template parameter.  I think `alignment` is more consistent with the rest of the code.

src/ds/address.h
```diff
@@ -39,4 +42,80 @@ namespace snmalloc
...
+   */
+  template<size_t granule, typename T = void>
+  inline T* pointer_align_up(void* p)
+  {
+    static_assert(granule > 0);
+    static_assert((granule & (granule - 1)) == 0);
+#if __has_builtin(__builtin_align_up)
+    return reinterpret_cast<T*>(__builtin_align_up(p, granule));
+#else
+    return reinterpret_cast<T*>(
```
mjp41: `address_cast` and `pointer_cast` might be clearer. I find `reinterpret_cast` horrible to read.

src/mem/alloc.h
```diff
@@ -40,8 +40,30 @@ namespace snmalloc
...
+     * allocation.  See SuperslabMap::set_large_size and external_address's
+     * handling of large reallocation redirections.
+     *
+     * Values 128 (inclusive) through 255 (inclusive) are as yet unused.
+     */
+
   };
 
+  /* Ensure that PageMapSuperslabKind values are actually disjoint */
+  static_assert(SUPERSLAB_BITS > 2, "Large allocations possibly too small");
```
mjp41: Why `2`?  Perhaps add a `PMEndMarker` to the enum, and then do 
```C
SUPERSLAB_BITS > next_pow2_bits_const(PMEndMarker - 1)
```

src/mem/largealloc.h
```diff
@@ -58,7 +58,7 @@ namespace snmalloc
   class MemoryProviderStateMixin : public PAL
   {
     std::atomic_flag lock = ATOMIC_FLAG_INIT;
-    address_t bump;
+    void* bump;
```
mjp41: Should this be `char*`

src/mem/largealloc.h
```diff
@@ -145,32 +145,39 @@ namespace snmalloc
...
-        if ((aligned_bump - bump) > remaining)
+        if constexpr (alignment != 0)
         {
-          new_block();
-        }
-        else
-        {
-          remaining -= aligned_bump - bump;
-          bump = aligned_bump;
+          uint8_t* aligned_bump = pointer_align_up<alignment, uint8_t>(bump);
```
mjp41: Should this be `char` rather than `uint8_t` based on @davidchisnall's comments for the earlier PR?

src/ds/address.h
```diff
@@ -39,4 +42,80 @@ namespace snmalloc
...
+    return ((static_cast<size_t>(address_cast(p)) | size) & (alignment - 1)) ==
+      0;
+  }
+
+  /**
+   * Align a pointer down to a statically specified granularity, which must be a
+   * power of two.
+   */
+  template<size_t granule, typename T = void>
+  inline T* pointer_align_down(void* p)
```
mjp41: `void*` or `T*`,  I guess the thought is that changing alignment will probably change the type?

src/mem/largealloc.h
```diff
@@ -287,16 +294,22 @@ namespace snmalloc
     template<AllowReserve allow_reserve>
     bool reserve_memory(size_t need, size_t add)
     {
-      if ((address_cast(reserved_start) + need) > address_cast(reserved_end))
+      assert(reserved_start <= reserved_end);
+
+      /*
+       * Spell this comparison in terms of pointer subtraction like this,
+       * rather than "reserved_start + need < reserved_end" becuase the
```
mjp41: becuase

src/mem/alloc.h
nwf: Something seems to have gone wrong; this hunk should have been already committed as part of #100.  I must have botched rebasing; will address when I'm back at my desk.

src/mem/largealloc.h
nwf: Probably so; fixed.

src/mem/largealloc.h
nwf: Probably so.  Fixed.

src/ds/address.h
nwf: In `largealloc`, `pagemap`, and `pal_open_enclave`, these functions get used on generic `void` or `char` pointers, but in the various slab uses, they are used to find the slab header and so give the type.  It seemed in keeping with `pointer_cast` and more concise than `static_cast<Slab*>(pointer_align_down<SLAB_SIZE>(this))`, for example.

PR #105, opened 2019-11-26
DNM WIP: CHERIfy via PAL

nwf:
This is not yet ready for merge, but I would like to solicit feedback on the general idea.  Specifically, while the previous attempt at CHERIfication of `snmalloc` relied on a pile of `ifdefs` everywhere (including `SNMALLOC_PAGEMAP_POINTERS`), this attempt vectors everything through the platform abstraction layer.  There are some caveats:

* The PAL is very early in the toposorting of dependencies, and so almost everything pagemap related needs to be passed in as template parameters.
* It uses higher-order templates, which is... somewhat exciting.
* While this compiles for CHERI, it should not be expected to actually work; at the very least, the `set` method of the CHERI chunkmap is wrong and, besides, most of the calls to the requisite hooks are not actually in place, just enough to ensure that we generate some code to test compilation.

This builds on #104, sorry for the duplicate commits at the moment.

### DIFF
--- a/src/aal/aal.h
+++ b/src/aal/aal.h
@@ -46,7 +46,7 @@ namespace snmalloc
      */
     static inline uint64_t tick()
     {
-#if __has_builtin(__builtin_readcyclecounter) && \
+#if __has_builtin(__builtin_readcyclecounter) && !defined(__CHERI__) && \
   !defined(SNMALLOC_NO_AAL_BUILTINS)
       return __builtin_readcyclecounter();
 #else
@@ -59,6 +59,8 @@ namespace snmalloc
 
 #ifdef PLATFORM_IS_X86
 #  include "aal_x86.h"
+#elif defined(__CHERI__) && defined(__mips__)
+#  include "aal_cheri_mips.h"
 #endif
 
 namespace snmalloc
--- a/src/aal/aal_cheri_mips.h
+++ b/src/aal/aal_cheri_mips.h
@@ -1,0 +1,71 @@
+#pragma once
+
+#define SNMALLOC_VA_BITS_64
+
+extern "C"
+{
+#include <machine/cheri.h>
+}
+
+namespace snmalloc
+{
+  /**
+   * CHERI-MIPS specific architecture abstraction layer.
+   */
+  class AAL_CHERI_MIPS
+  {
+    static inline void halt_out_of_order()
+    {
+      static unsigned st;
+      __sync_swap(&st, 0); // XXX: Overkill!
+    }
+
+    static inline uint64_t tickp()
+    {
+      static unsigned st;
+      __sync_swap(&st, 0); // XXX: Overkill!
+      return cheri_get_cyclecount();
+    }
+
+  public:
+    static inline void pause()
+    {
+      ;
+      /*
+       * XXX the PAUSE instruction could be correct or not, depending on the
+       * implementation details of std::atomic_flag and other callers.
+       *
+       * For PAUSE to be
+       * correct, .test_and_set must exit with the LL flag still set and
+       * .clear must store to the same word probed by test_and_set.  It
+       * seems like these will be true, but they are doubtless not required
+       * to be so.
+       */
+    }
+
+    static inline void prefetch(void* p)
+    {
+      __builtin_prefetch(p, 0, 3);
+    }
+
+    static inline uint64_t tick()
+    {
+      return cheri_get_cyclecount();
+    }
+
+    static inline uint64_t benchmark_time_start()
+    {
+      halt_out_of_order();
+      return AAL_Generic<AAL_CHERI_MIPS>::tick();
+    }
+
+    static inline uint64_t benchmark_time_end()
+    {
+      uint64_t t = AAL_Generic<AAL_CHERI_MIPS>::tickp();
+      halt_out_of_order();
+      return t;
+    }
+  };
+
+  using AAL_Arch = AAL_CHERI_MIPS;
+}
--- a/src/ds/bits.h
+++ b/src/ds/bits.h
@@ -52,8 +52,6 @@ namespace snmalloc
       return (static_cast<T>(1)) << shift;
     }
 
-    static constexpr size_t ADDRESS_BITS = is64() ? 48 : 32;
-
     inline size_t clz(size_t x)
     {
 #if defined(_MSC_VER)
--- a/src/mem/chunkmap.h
+++ b/src/mem/chunkmap.h
@@ -141,7 +141,7 @@ namespace snmalloc
    * implementation (for example, to move pagemap updates to a different
    * protection domain).
    */
-  template<typename PagemapProvider = GlobalPagemap>
+  template<typename PagemapProvider>
   struct DefaultChunkMap
   {
     /**
@@ -238,7 +238,7 @@ namespace snmalloc
   };
 
 #ifndef SNMALLOC_DEFAULT_CHUNKMAP
-#  define SNMALLOC_DEFAULT_CHUNKMAP snmalloc::DefaultChunkMap<>
+#  define SNMALLOC_DEFAULT_CHUNKMAP snmalloc::Pal::PalChunkMap<GlobalPagemap>
 #endif
 
 } // namespace snmalloc
--- a/src/mem/chunkmap.h
+++ b/src/mem/chunkmap.h
@@ -47,14 +47,19 @@ namespace snmalloc
 // Use flat map is under a single node.
 #  define SNMALLOC_MAX_FLATPAGEMAP_SIZE PAGEMAP_NODE_SIZE
 #endif
-  static constexpr bool USE_FLATPAGEMAP = pal_supports<LazyCommit>() ||
-    (SNMALLOC_MAX_FLATPAGEMAP_SIZE >=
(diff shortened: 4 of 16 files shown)


nwf:
Here's an updated version that provides an entire CHERI-compatible PAL as a worked example.  Because the existing `ChunkMap` implementations rely on static dispatch, I've marked them `final` to prevent accidental inheritance.

src/aal/aal_cheri_mips.h
```diff
@@ -0,0 +1,71 @@
...
+{
+  /**
+   * CHERI-MIPS specific architecture abstraction layer.
+   */
+  class AAL_CHERI_MIPS
+  {
+    static inline void halt_out_of_order()
+    {
+      static unsigned st;
+      __sync_swap(&st, 0); // XXX: Overkill!
```
davidchisnall: This (and the use below) look as if they're relying on a not-very-well-specified builtin emitting a particular barrier on MIPS.  I think I'd rather have inline assembly than this.  Alternatively, is one of the `std::thread_fence` variants sufficient?

src/aal/aal.h
```diff
@@ -46,7 +46,7 @@ namespace snmalloc
      */
     static inline uint64_t tick()
     {
-#if __has_builtin(__builtin_readcyclecounter) && \
+#if __has_builtin(__builtin_readcyclecounter) && !defined(__CHERI__) && \
```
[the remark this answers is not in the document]
davidchisnall: Reading the later code, is it that `__builtin_readcyclecounter` is not sufficiently serialising on MIPS?  If so, we probably want to refactor the callers of this, rather than have an  `#ifdef` here.

On a related note, please avoid using `__CHERI__` for things that are MIPS specific, we want to be able to use `__CHERI__` for CHERI/RISC-V and Morello.

src/aal/aal_cheri_mips.h
```diff
@@ -0,0 +1,71 @@
...
+{
+#include <machine/cheri.h>
+}
+
+namespace snmalloc
+{
+  /**
+   * CHERI-MIPS specific architecture abstraction layer.
+   */
+  class AAL_CHERI_MIPS
```
davidchisnall: None of this looks CHERI specific.  Should this be a MIPS or BERI AAL?

src/aal/aal_cheri_mips.h
```diff
@@ -0,0 +1,71 @@
+#pragma once
+
+#define SNMALLOC_VA_BITS_64
+
+extern "C"
+{
+#include <machine/cheri.h>
```
davidchisnall: This is a FreeBSD-specific header, but this is in platform-independent code.  Please can you provide the implementation of the CHERI cycle counter access function (which, I believe, is BERI-specific, not dependent on CHERI?)?

src/ds/bits.h
```diff
@@ -52,8 +52,6 @@ namespace snmalloc
       return (static_cast<T>(1)) << shift;
     }
 
-    static constexpr size_t ADDRESS_BITS = is64() ? 48 : 32;
```
davidchisnall: This has been moved into the PAL, but I wonder whether it's actually a mixture of PAL and AAL.  Should the PAL provide a maximum by select the minimum of its version and the AAL's version?

src/mem/chunkmap.h
```diff
@@ -163,6 +180,24 @@ namespace snmalloc
...
+     * allocation.  This provides such a hook.  On architectures where
+     * pointers are just integers, this is a no-op that should always be
+     * inlined into nothing.
+     *
+     * The template paramter controls whether the returned pointer points
+     * to the same place as the argument or to the beginning of the chunkmap
+     * region containing the argument.
+     */
+    template<bool offset = true>
+    SNMALLOC_FAST_PATH static void* getp(void* p)
```
davidchisnall: Please give this a more informative name.  I came across the uses of this before the definition in the review and had no idea what `p` was or why I was getting it from the chunk map.  I'd suggest `rederrive_pointer` or similar.

src/mem/chunkmap.h
```diff
@@ -47,14 +47,19 @@ namespace snmalloc
...
-  static constexpr bool USE_FLATPAGEMAP = pal_supports<LazyCommit>() ||
-    (SNMALLOC_MAX_FLATPAGEMAP_SIZE >=
-     sizeof(FlatPagemap<SUPERSLAB_BITS, uint8_t>));
 
-  using ChunkmapPagemap = std::conditional_t<
-    USE_FLATPAGEMAP,
-    FlatPagemap<SUPERSLAB_BITS, uint8_t>,
-    Pagemap<SUPERSLAB_BITS, uint8_t, 0>>;
+  template<typename T>
+  static constexpr bool use_flatpagemap_for()
```
davidchisnall: For what?  The name implies that the function takes an argument, but it does not.

I had to read the implementation to discover that it means 'use flatmap for `T`'.  

Please add a doc comment.

src/mem/chunkmap.h
```diff
@@ -141,9 +144,23 @@ namespace snmalloc
    * implementation (for example, to move pagemap updates to a different
    * protection domain).
    */
-  template<typename PagemapProvider = GlobalPagemap>
+  template<
+    template<typename>
+    typename PagemapProviderTemplate,
+    template<auto>
+    typename ChunkmapPagemapTemplate>
```
davidchisnall: We seem to have lost default parameters here.  Are they no longer needed?

src/mem/pagemap.h
```diff
@@ -191,7 +191,7 @@ namespace snmalloc
           return std::pair(nullptr, 0);
 
         shift -= BITS_PER_INDEX_LEVEL;
-        ix = (addr >> shift) & ENTRIES_MASK;
+        ix = (static_cast<size_t>(addr) >> shift) & ENTRIES_MASK;
```
davidchisnall: I would rather we avoid `static_cast<size_t>` throughout the code and define a function that does this conversion.

We currently have `address_t`, which is actually a pointer and not an address.  We should probably rename that to `pointer_t` or similar and have a separate `address_t`, rather than assuming `size_t` is always the correct thing (`size_t` only has to be big enough for the maximum size of a single object, it would be valid to have a C implementation with 64-bit pointers but a 32-bit `size_t`).

src/mem/pagemap.h
```diff
@@ -387,7 +387,7 @@ namespace snmalloc
      */
     size_t index_for_address(uintptr_t p)
     {
-      return reinterpret_cast<size_t>((p >> SHIFT) & (OS_PAGE_SIZE - 1));
+      return bits::align_down(static_cast<size_t>(p) >> SHIFT, OS_PAGE_SIZE);
```
davidchisnall: As above.

src/aal/aal.h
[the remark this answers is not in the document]
nwf: This may be stale; I'll check if `__builtin_readcyclecounter` does the right thing for us.

src/aal/aal_cheri_mips.h
nwf: I'll check with Alex and Jon.

src/aal/aal_cheri_mips.h
nwf: That's fair.  Though if rederivation is going to move into the AAL, that's certainly going to be CHERI-specific.  I suppose having MIPS from which CHERI-MIPS inherits is fine.

src/ds/bits.h
nwf: I suppose so?  It's a little unfortunate in that if a platform ever widens its kernel-provided mapping of userland then an old build of snmalloc is going to break.

src/mem/chunkmap.h
nwf: Sure.

src/mem/chunkmap.h
nwf: Is a C++14 templatized variable declaration OK rather than a function?

src/mem/chunkmap.h
nwf: They can't be provided because we need to forward-declare the template in the PAL (or AAL), and, AIUI, only the first declaration can provide the defaults?

nwf:
CHERI is certainly architectural, but the current implementations also all support hybrid code, so I wasn't sure whether it was a PAL or AAL thing; if you'd rather it be in the AAL, I think I'm OK with that and can push it over there.  I'm travelling until Monday, but I can get you something to look at next week.

The general idea of passing in both the `PagemapProviderTemplate` and `ChunkmapPagemapTemplate` seems OK?

src/mem/chunkmap.h
mjp41: Happy with C++14, I think the code requires C++17 in some places.

src/mem/pagemap.h
nwf: On CHERI we have a `vaddr_t` that's guaranteed to be big enough.  I guess we can have the AAL/APL define what `address_t` means?  Is there anywhere we don't want the existing `address_t` to be `uintptr_t`?

mjp41:
This is so old now, I think it should be closed.  Please feel free to update and resubmit.  I am super keen for CHERI support, especially with the [Morello work](https://www.cl.cam.ac.uk/research/security/ctsrd/cheri/cheri-morello.html).  But I think this PR is too old to be of use.

PR #106, opened 2019-11-28
atomic swap 16 flag missing fix.

devnexen:
In some platforms, it is rather amd64 than the more common x86_64,
thus missing to rightfully set this flag.


devnexen:
FreeBSD

PR #107, opened 2019-12-04
Introduce AAL feature flags

nwf:
And use them to gate `POISON`-ing pointers when `CHECK_CLIENT` is defined.  The bulk of the change is just stylistic to normalize case and the use of templated vardecls.

This extracts some small part of conversations had in #105 for incremental review.

### DIFF
--- a/src/mem/metaslab.h
+++ b/src/mem/metaslab.h
@@ -109,11 +109,12 @@ namespace snmalloc
     /// simple corruptions.
     static SNMALLOC_FAST_PATH void store_next(void* p, void* head)
     {
-#ifndef CHECK_CLIENT
       *static_cast<void**>(p) = head;
-#else
-      *static_cast<void**>(p) = head;
-      *(static_cast<uintptr_t*>(p) + 1) = address_cast(head) ^ POISON;
+#if defined(CHECK_CLIENT)
+      if constexpr (aal_supports<IntegerPointers>)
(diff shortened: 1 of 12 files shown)


src/mem/metaslab.h
```diff
@@ -109,23 +109,27 @@ namespace snmalloc
...
-      if ((next ^ chk) != POISON)
-        error("Detected memory corruption.  Use-after-free.");
+#if defined(CHECK_CLIENT)
+      if constexpr (aal_supports<IntegerPointers>)
+      {
+        uintptr_t next = *static_cast<uintptr_t*>(node);
+        uintptr_t chk = *(static_cast<uintptr_t*>(node) + 1);
+        if ((next ^ chk) != POISON)
+          error("Detected memory corruption.  Use-after-free.");
+      }
```
davidchisnall: Not necessary for this PR, but it would be interesting to think about whether we can implement equivalent functionality on CHERI by sealing these, so we can detect accidental non-malicious use.

It would also be nice to use ARM PAC here where possible.  I therefore slightly wonder if this is something that should be part of the AAL as a protect / check pointer function pair?

nwf: That (adding protect/check pointer fns to AAL) seems sensible to me, but... sealing would require that we somehow get an otype, which is going to be a Pal function.

Issue #108, opened 2020-01-05
GCC 9 strange failure


SchrodingerZhu:
GCC9 will yield error when compile target is static.
## Compile with clang (static)
```zsh
➜  build git:(master) ✗ CC=clang CXX=clang++ cmake .. -DCMAKE_BUILD_TYPE=Release -G Ninja
-- The C compiler identification is Clang 9.0.1
-- The CXX compiler identification is Clang 9.0.1
-- Check for working C compiler: /usr/bin/clang
-- Check for working C compiler: /usr/bin/clang -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/clang++
-- Check for working CXX compiler: /usr/bin/clang++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Looking for pthread.h
-- Looking for pthread.h - found
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Failed
-- Looking for pthread_create in pthreads
-- Looking for pthread_create in pthreads - not found
-- Looking for pthread_create in pthread
-- Looking for pthread_create in pthread - found
-- Found Threads: TRUE  
-- Generating clangformat target using /usr/bin/clang-format
-- Configuring done
-- Generating done
-- Build files have been written to: /home/schrodinger/CLionProject/snmalloc/build
➜  build git:(master) ✗ ninja
[4/4] Linking C static library libsnmallocshim.a
```
## Compile with GCC8 (static)
```
➜  build git:(master) ✗ CC=gcc-8 CXX=g++-8 cmake .. -DCMAKE_BUILD_TYPE=Release -G Ninja
-- The C compiler identification is GNU 8.3.0
-- The CXX compiler identification is GNU 8.3.0
-- Check for working C compiler: /usr/bin/gcc-8
-- Check for working C compiler: /usr/bin/gcc-8 -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/g++-8
-- Check for working CXX compiler: /usr/bin/g++-8 -- works
-- Detecting CXX compiler ABI info
[...]

PR #110, opened 2020-01-05
Fix GCC 9 error.

mjp41:
GCC has some aggressive warnings that can be imprecise for checking array bounds access.  This commit converts an assert to an assume to remove the warning.

I do not have GCC 9 to test this, but it follows a similar pattern to we have seen with other versions of GCC.  Should address #108.


mjp41:
@schrodingerzhu would you be able to test if this fixes the issues you observed in #108. I don't currently have access to the version of GCC 9.0 on my dev machine.

mjp41:
I have now reproduced #108 and can confirm this fixes it.

SchrodingerZhu:
I confirm the fix.

Issue #109, opened 2020-01-05
Adding Rust Support


SchrodingerZhu:
I am trying to bind `snmalloc` to rust to have a comprehensive [`malloc-bench`](https://github.com/SchrodingerZhu/malloc-bench) (**WIP**).

Basically, I have already created several bindings: [`snmalloc`](https://github.com/SchrodingerZhu/snmalloc),[`snmalloc-sys`](https://github.com/SchrodingerZhu/snmalloc-sys), [`snmallocator`](https://github.com/SchrodingerZhu/snmallocator).

Yet, there is a issue to provide efficient `realloc` with alignment. I [added one](https://github.com/SchrodingerZhu/snmalloc/blob/2ee7be2319855569e500a7783b5c2cc4b049d9c3/src/override/malloc.cc#L168) following the code in `malloc.h`, but I do not know whether it is sound and safe. So I guess I need some guidance here.

mjp41:
Hi @SchrodingerZhu 

This is awesome, firstly is there a specification of realloc aligned for Rust that I can read. There are a bunch of things that are done here to meet various C expectations that might not be required. The most efficient version will do no more than necessary. 

In particular, the interaction between alignment work and realloc work has many possibilities: What should size = 0, alignment = 16 do?
- error
- free the object
- return a pointer that is aligned to 16.
The order of checks is important to get each of these behaviours. 

Second minor point, if you reformat a file it makes code review very hard, if we are to take any changes they must be formated with our Clangformat config, our CI enforces this.

mjp41:
Is it supposed to satisfy the following:

https://doc.rust-lang.org/std/alloc/trait.GlobalAlloc.html#safety-4

How defensive do you want the code to be for undefined behaviour?  The current code checks a lot of properties and returns error codes, but as the property failures are undefined behaviour we could be more efficient here?  So it depends on the performance/usability trade-off for how this is implemented.  I would suggest we turn a lot of stuff into asserts, so the Debug version checks everything and release is fast.

SchrodingerZhu:
I am so sorry that I am being terribly busy recently. I will back to this work at this weekend. :P

mjp41:
@SchrodingerZhu no need to apologise, just let me know if there is anything I can do to help in advance.  

The project: [mimalloc_rust](https://github.com/purpleprotocol/mimalloc_rust) looks like it could be good inspiration.  It has fewer repos, then your current proposal, so probably easier to keep consistent. 

[...]

PR #111, opened 2020-01-09
Small refactoring of code for finding a sizeclass for an alignment

mjp41:
snmalloc has natural alignment for each sizeclass, this means that we can search for a sizeclass larger then the requested sizeclass whose natural alignment meets that specified by the call.

This is currently used in a single place, so was directly written into `memalign`. With #109 there is a wish to make a larger API surface that supports alignment.  This PR is intended to make that easier.

### DIFF
--- a/src/override/malloc.cc
+++ b/src/override/malloc.cc
@@ -104,14 +104,6 @@ extern "C"
   }
 #endif
 
-  SNMALLOC_EXPORT void*
-    SNMALLOC_NAME_MANGLE(aligned_alloc)(size_t alignment, size_t size)
-  {
-    assert((size % alignment) == 0);
-    (void)alignment;
-    return SNMALLOC_NAME_MANGLE(malloc)(size);
-  }
-
   inline size_t aligned_size(size_t alignment, size_t size)
   {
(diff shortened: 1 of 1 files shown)


mjp41:
CC: @schrodingerzhu  This may help your Rust efforts.

src/override/malloc.cc
```diff
@@ -104,12 +104,33 @@ extern "C"
...
-  SNMALLOC_EXPORT void*
-    SNMALLOC_NAME_MANGLE(aligned_alloc)(size_t alignment, size_t size)
+  inline size_t aligned_size(size_t alignment, size_t size)
   {
-    assert((size % alignment) == 0);
-    (void)alignment;
-    return SNMALLOC_NAME_MANGLE(malloc)(size);
+    // Client responsible for checking alignment is not zero
+    // Client responsible for checking for overflow
+    assert((size + alignment) > size);
```
Theodus: Why are we checking for overflow on this addition?

mjp41: Not sure.  It was in the previous implementation, but seems unnecessary now.  I think it may be from your original version, and we should have dropped it long ago.

PR #119, opened 2020-01-25
Make snmalloc build on Windows with Clang

mjp41:
Fixes a few places where Clang complains about Windows specific code,
and also uses macros supported by Clang on Windows.  A few places
separating platform and compiler specific code, as MSVC and WIN32 were
used interchangably previously.


mjp41:
@Theodus @plietar @davidchisnall  Could one of you approve this? Thanks

PR #121, opened 2020-01-30
Some improvements to CI

mjp41:


azure-pipelines.yml
```diff
@@ -240,14 +243,17 @@ jobs:
       cmakeArgs: '..'
 
   - script: |
+      set -eo pipefail
       make clangformat
       git diff
       if [ "$(git diff $(Build.SourceVersion))" != "" ]; then exit 1; fi
```
plietar: You can probably replace both calls to `git diff` with:

```suggestion
      git diff --exit-code $(Build.SourceVersion)
```

Nicer to rely on git's exit code than looking at its output.

azure-pipelines.yml
```diff
@@ -240,14 +243,17 @@ jobs:
...
 
     workingDirectory: build
+    failOnStderr: true
     displayName: 'Clang-Format'
 
   - script: |
+      set -eo pipefail
       clang-tidy-8  src/override/malloc.cc  -header-filter="`pwd`/*" -export-fixes=tidy.fail -- -std=c++17 -mcx16
       if [ -f tidy.fail ] ; then exit 1 ; fi
```
plietar: If you add `-warnings-as-errors='*'`, clang-tidy will report errors and exit with a non-zero code. You should then be able to remove this line, and the `-export-fixes`

Unrelated, but why do we have clang-tidy on just this file?

mjp41: We can try and expand it.

PR #123, opened 2020-02-03
Issue with low-memory notification

mjp41:
The low-memory notification was getting into an infinite loop.  This
fixes the loop termination, and provides a test for platforms which
support low-memory notification.


mjp41:
@snf this fixes the Windows performance issue you were having.

mjp41:
Not sure what the right test for 32bit Windows is.  It does have a low-memory notification, but runs out of address space, before physical memory is exhausted.

I'll probably just disable the test on 32bit windows for now.

Issue #114, opened 2020-01-22
Add alignment to the surface API


mjp41:
Based on the work in #109 alignment with `snmalloc` can be implemented extremely efficiently: just two additional instructions, which can be optimised away if the alignment is known to be `1` at compile time.

We should expand the internal `alloc` functions to take an optional alignment parameter (defaulting to 1). This will be useful for scenarios in [Project Verona](http://github.com/microsoft/verona), and will improve the Rust support in #109.

Integrating it into the internal alloc routines will allow for more efficient detecting corner cases such as zero size or zero alignment by pushing them onto the slow path.  This will allow clients to not require checking if it can't be guaranteed by construction. 

Currently the maximum alignment we support is SUPERSLAB_SIZE.  This is sufficient for most applications, but allowing more would reduce unexpected behaviour. This requires fiddling with the large allocator. 

Work items:
- [ ] Expose alignment at the API level.
- [ ] Enforce power of two alignment on large allocations.

mjp41:
I have a PR ready for
>  Enforce power of two alignment on large allocations

It is just waiting for the refactors in #120 to land on master.

PR #124, opened 2020-02-04
Make Large allocations naturally aligned

mjp41:
This makes all large allocations aligned to their size next power of two size.

This introduces a new strategy for aligning blocks on platforms without a way to fetched aligned virtual memory (e.g. VirtualAlloc2, or mmap with MAP_ALIGNED).

To aid in debugging also uses mprotect to switch off unused pages in debug on posix platforms. This enables better testing of the Commit/Decommit code from Posix platforms.

### DIFF
--- a/src/mem/largealloc.h
+++ b/src/mem/largealloc.h
@@ -273,7 +273,7 @@ namespace snmalloc
 
         // printf("Alloc %zx (size = %zx)\n", start, size);
 
-        void* result = pointer_cast<void>(start);  
+        void* result = pointer_cast<void>(start);
         if (committed)
           PAL::template notify_using<NoZero>(result, size);
 
--- a/src/mem/largealloc.h
+++ b/src/mem/largealloc.h
@@ -63,16 +63,11 @@ namespace snmalloc
 
     void new_block()
     {
-      size_t size = SUPERSLAB_SIZE;
-      void* r = reserve<false>(&size, SUPERSLAB_SIZE);
-
-      if (size < SUPERSLAB_SIZE)
-        error("out of memory");
-
+      void* r = reserve<false>(SUPERSLAB_SIZE, SUPERSLAB_SIZE);
       PAL::template notify_using<NoZero>(r, OS_PAGE_SIZE);
 
       bump = r;
-      remaining = size;
+      remaining = SUPERSLAB_SIZE;
     }
 
     /**
@@ -202,7 +197,7 @@ namespace snmalloc
     }
 
     template<bool committed>
-    void* reserve(size_t* size, size_t align) noexcept
+    void* reserve(size_t size, size_t align) noexcept
     {
       if constexpr (pal_supports<AlignedAllocation, PAL>)
       {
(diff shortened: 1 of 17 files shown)


mjp41:
#114 needs this.

src/mem/largealloc.h
```diff
@@ -202,35 +216,76 @@ namespace snmalloc
...
+        void* p = PAL::template reserve<committed>(request);
+
+        address_t p0 = address_cast(p);
+        address_t start = bits::align_up(p0, align);
+        address_t p1 = p0 + request;
+        address_t end = start + size;
+
+        // Turn off pages for extra allocations.
+        // Will turn pages back on for stack entries.
+        if (committed)
```
davidchisnall: Wouldn't it be better to `reserve<false>` above and then commit the required bits here, rather than committing everything above and then decommitting here?  The current flow requires that the OS commit a big chunk and then returns most of it almost immediately.

src/pal/pal_freebsd_kernel.h
```diff
@@ -61,9 +61,9 @@ namespace snmalloc
     }
 
     template<bool committed>
-    void* reserve(size_t* size, size_t align)
+    void* reserve(size_t size, size_t align)
     {
-      size_t request = *size;
+      size_t request = size;
```
davidchisnall: Please remove this and just use `size` everywhere.

src/pal/pal_posix.h
```diff
@@ -54,8 +54,12 @@ namespace snmalloc
     void notify_not_using(void* p, size_t size) noexcept
     {
       assert(is_aligned_block<OS_PAGE_SIZE>(p, size));
+#ifndef NDEBUG
```
davidchisnall: It would be nice to have this controlled by a separate flag that defaults to on in debug CMake configurations.

src/pal/pal_posix.h
```diff
@@ -72,11 +76,13 @@ namespace snmalloc
...
       if constexpr (zero_mem == YesZero)
         static_cast<OS*>(this)->template zero<true>(p, size);
-      else
-      {
-        UNUSED(p);
-        UNUSED(size);
-      }
+
+#ifndef NDEBUG
+      mprotect(p, size, PROT_READ | PROT_WRITE);
```
davidchisnall: As above.

src/test/func/fixed_region/fixed_region.cc
```diff
@@ -32,8 +32,9 @@ int main()
 {
   MemoryProviderStateMixin<DefaultPal> mp;
 
-  size_t size = 1ULL << 28;
-  oe_base = mp.reserve<true>(&size, 0);
+  size_t large_class = 28 - SUPERSLAB_BITS;
```
davidchisnall: Not introduced in this change, but it would be nice to have 28 not be a magic number here.

src/test/func/two_alloc_types/main.cc
```diff
@@ -45,8 +45,9 @@ int main()
 
   MemoryProviderStateMixin<DefaultPal> mp;
 
-  size_t size = 1ULL << 26;
-  oe_base = mp.reserve<true>(&size, 1);
+  size_t large_class = 26 - SUPERSLAB_BITS;
```
davidchisnall: The same with this 26.

Issue #127, opened 2020-02-16
Make Lazy decommit on Windows Asynchronous


mjp41:
Lazy decommit occurs on Windows once a low-memory notification has been received, and then any thread ends up calling into the large allocator.  The decommit, is only decommiting large chunks of memory held in global datastructures, not in thread local data structures.

If an application is idle, when the low-memory notification is received nothing would happen.  Can we make the call-back actually perform the lazy-decommit. The lock-free data structures are thread safe, so this seems like it would be a win, and only a small code change.

On Windows this seems like it would be sensible. But not sure about other platforms we might want to support.  @davidchisnall do you have any thoughts on this?

mjp41:
@buybackoff I have raised this issue as it seems like we are close to addressing your use case, but would need to make a few changes.

davidchisnall:
Currently, only Windows, Android, and XNU support low-memory notifications and only Windows delivers them via a useful API for low-level code (on XNU, there appears to be a kqueue event for low memory, but the lowest level of published API is libdispatch, on Android the lowest-level published API is Java, though presumably there's a magic `/dev` node that you can monitor [possibly only if you're root]).  I don't think it's worth worrying too much about other platforms yet (though it might be nice to expose the notification as a callback on platforms that don't have low-level API support so that iOS and Android users can invoke it from their high-level notification).  

Is your comment in relation to the possible approach that we discussed of implementing a virtual decommit in the PAL that takes ownership of the decommitted pages but doesn't actually decommit them unless there's memory pressure?  Even if that were protected by a mutex, it would likely have significantly lower overhead than anything that involves VM subsystem interaction.

mjp41:
I was thinking about something a little simpler initially, though, I want to follow up your idea. But it needs careful design for concurrency if we allow a more fine-grained decommit.  If we don't allow a more fine-grained decommit, then I think the approach below will be good enough.

So I was thinking of allowing the PAL to deliver async callbacks for low-memory notifications:
```
   class PalCallbackObject
   {
      PalCallbackObject* next;
      void (*process)(PalCallbackObject* );  
   };

   void register_low_memory_callback(PalCallbackObject* notify)
```
Then we register an object that will call `LargeAlloc::lazy_decommit`.  This means that even an idle process will decommit memory on Windows. Currently, we need churn to get off the fast path to trigger the pooling to see if the epoch has changed.

davidchisnall:
[...]

PR #128, opened 2020-02-21
Make Lazy Decomit asynchronous

mjp41:
On platforms that support low-memory notifications register callbacks
that perform lazy decommit. This allows idle processes to return memory
to the OS. Without incurring the cost of constantly committing and
decommitting memory.

### DIFF
--- a/src/pal/pal_windows.h
+++ b/src/pal/pal_windows.h
@@ -21,22 +21,24 @@ namespace snmalloc
   class PALWindows
   {
     /**
-     * The number of times that the memory pressure notification has fired.
-     */
-    static inline std::atomic<uint64_t> pressure_epoch;
-    /**
      * A flag indicating that we have tried to register for low-memory
      * notifications.
      */
     static inline std::atomic<bool> registered_for_notifications;
     static inline HANDLE lowMemoryObject;
+
+    /**
+     * List of callbacks for low-memory notification
+     **/
+    static inline PalNotifier low_memory_callbacks;
+
     /**
      * Callback, used when the system delivers a low-memory notification.  This
-     * simply increments an atomic counter each time the notification is raised.
+     * calls all the handlers registered with the PAL.
      */
     static void CALLBACK low_memory(_In_ PVOID, _In_ BOOLEAN)
     {
-      pressure_epoch++;
+      low_memory_callbacks.notify_all();
     }
 
   public:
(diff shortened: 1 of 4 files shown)


mjp41:
The addresses the discussion in #127.

@davidchisnall I am a bit unhappy with the unsafe downcast, but also don't want to pull in vtables and RTTI for downcasts.  So thoughts on how this could be better welcome.

src/pal/pal_windows.h
```diff
@@ -20,23 +20,31 @@ namespace snmalloc
...
+
     /**
      * Callback, used when the system delivers a low-memory notification.  This
-     * simply increments an atomic counter each time the notification is raised.
+     * calls all the handlers registered with the PAL.
      */
     static void CALLBACK low_memory(_In_ PVOID, _In_ BOOLEAN)
     {
-      pressure_epoch++;
+      PalNotificationObject* curr = low_memory_callbacks;
```
davidchisnall: Can this and the list be pulled out into a separate class?  I'd rather not have to copy and paste this for other platforms.

mjp41: Done

mjp41:
@buybackoff I have modified the test here, so you can run the new code in several parallel processes.  The test runs until it exhausts physical RAM, and then deallocates everything, and waits for a key press.  If you then stress the machine (say by running another copy of the test), then the RAM usages drops down to almost zero, on my machine 10GiB -> 34MiB, while it is sitting waiting for user input.  I think this matches perfectly the scenario you were seeing.

buybackoff:
@mjp41 thanks! Will give it a try.

Should shared memory (SM) paged-in page cause lazy decommit? Or new SM paged-in pages will only evict other SM pages while snmalloc's pages are not yet decommited?

mjp41:
So we are just performing lazy decommit on a low memory notification:

https://docs.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-creatememoryresourcenotification

This says it is when "available physical memory is running low".  'snmalloc' will return any "chunk" 1MiB or 16MiB range of memory we are no longer using.  Any allocations on the chunk will cause it to remain resident.  Decommiting memory at a finer granularity would involve much more synchronization on fairly common paths, so I am not currently keen to do this.

If the shared-memory is using physical pages then I believe it can trigger 'snmalloc' to give up memory it does not need.

PR #129, opened 2020-02-26
Removing some calls into the C++ runtime

mjp41:
These changes are to reduce the C++ runtime calls when the TLS state is managed externally.
This enables the code to be included in really restricted build environments such as Open Enclave where the allocator is needed before any of the C++ library is available.

The primary changes are to remove
```
GlobalPlaceHolder
```
as a global if not needed, and to convert 
```
default_memory_provider
```
from a C++ global with initialisation, into a pointer with our custom `Singleton` constructor.  This removes calls to 
```
__cxa_guard_acquire
```
and 
```
__cxa_guard_release
```

### DIFF
--- a/src/mem/globalalloc.h
+++ b/src/mem/globalalloc.h
@@ -38,7 +38,7 @@ namespace snmalloc
 
     static AllocPool* make() noexcept
     {
-      return make(default_memory_provider);
+      return make(default_memory_provider());
     }
 
     Alloc* acquire()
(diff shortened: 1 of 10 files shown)


davidchisnall:
Looks fine to me.  For the FreeBSD libc build, I just add implementations of those two functions with private visibility (they're dealing with a data structure with an ABI-defined layout, so it's safe to have multiple versions of the functions), but it's nicer not to have to bother.

PR #135, opened 2020-03-04
replace assert with SNMALLOC_ASSERT

achamayou:
This is useful to avoid link time dependencies against libc (__assert_fail etc) in the OpenEnclave + snmalloc build, which break because oecorelibc does not implement those symbols.


src/ds/defines.h
```diff
@@ -30,8 +30,25 @@
 
 #define UNUSED(x) ((void)(x))
 
+namespace snmalloc
+{
+  void error(const char* const str);
+} // namespace snmalloc
+
+#ifdef NDEBUG
+#  define SNMALLOC_ASSERT(expr) ((void 0))
```
mjp41:
```suggestion
#  define SNMALLOC_ASSERT(expr) { }
```

src/pal/pal_consts.h
```diff
@@ -112,4 +116,4 @@ namespace snmalloc
       }
     }
   };
-} // namespace snmalloc
+} // namespace snmalloc
```
mjp41:
```suggestion
} // namespace snmalloc

```

src/ds/defines.h
```diff
@@ -30,8 +30,25 @@
 
 #define UNUSED(x) ((void)(x))
 
+namespace snmalloc
+{
+  void error(const char* const str);
```
mjp41:
```suggestion
  // Forwards reference so that the platform can define how to handle errors.
  void error(const char* const str);
```

src/ds/defines.h
```diff
@@ -30,8 +30,26 @@
...
 #define UNUSED(x) ((void)(x))
 
+namespace snmalloc
+{
+  // Forwards reference so that the platform can define how to handle errors.
+  void error(const char* const str);
+} // namespace snmalloc
+
+#ifdef NDEBUG
+#  define SNMALLOC_ASSERT(expr) { }
```
mjp41:
```suggestion
#  define SNMALLOC_ASSERT(expr) \
    {}
```

PR #137, opened 2020-03-08
fix MinGW

SchrodingerZhu:
- update `CMakeLists.txt` to detect "AMD64"
- update macros
  - define `NOMINMAX` only when not defined
  - define `USE_CLZLL` for mingw
  - include `<cstdio>` for mingw
- update `clz(size_t)`
  - `__builtin_clzll` for mingw

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -72,6 +72,8 @@ if(NOT MSVC)
    target_compile_options(snmalloc_lib INTERFACE -mcx16)
   elseif(${CMAKE_SYSTEM_PROCESSOR} STREQUAL "amd64")
    target_compile_options(snmalloc_lib INTERFACE -mcx16)
+  elseif(${CMAKE_SYSTEM_PROCESSOR} STREQUAL "AMD64")
+   target_compile_options(snmalloc_lib INTERFACE -mcx16)
   elseif(${CMAKE_SYSTEM_PROCESSOR} STREQUAL "x86")
    target_compile_options(snmalloc_lib INTERFACE -mcx16)
   # XXX elseif ARM?
--- a/src/ds/bits.h
+++ b/src/ds/bits.h
@@ -1,7 +1,6 @@
 #pragma once
 
 #include <limits>
-#include <stddef.h>
 
 // #define USE_LZCNT
 
@@ -14,7 +13,7 @@
 #include <cstdint>
 #include <type_traits>
 #if defined(_WIN32) && defined(__GNUC__)
-#define USE_CLZLL
+#  define USE_CLZLL
 #endif
 #ifdef pause
 #  undef pause
--- a/src/ds/bits.h
+++ b/src/ds/bits.h
@@ -13,7 +13,9 @@
 #include <cassert>
 #include <cstdint>
 #include <type_traits>
-
+#if defined(_WIN32) && defined(__GNUC__)
+#define USE_CLZLL
+#endif
 #ifdef pause
 #  undef pause
 #endif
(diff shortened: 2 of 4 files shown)


SchrodingerZhu:
I checked that with `mingw-w64-i686-toolchain`, the project cannot be compiled. The `-mcx16` cannot work somehow. Maybe we can just drop support for `MinGW` with `i686` toolchains.

mjp41:
I wonder if there are compatibility issues with linking in the `mincore.lib`.  If you configure cmake with 
```
-DWIN8COMPAT=TRUE
```
It won't try to use `VirtualAlloc2`, or pull in `mincore.lib`.

Also, I am a little concerned the changes you are making will affect the Windows-clang target.

mjp41:
Sorry, just reread your mcx16 comment.  That is pretty crucial to one of the datastructures, so I don't think we can support this.

SchrodingerZhu:
> Sorry, just reread your mcx16 comment. That is pretty crucial to one of the datastructures, so I don't think we can support this.

For `MinGW x86_64`, `mcx16` works fine; but for `MinGW i686`, the `mcx16` feature cannot be enabled even with `-mcx16`. Therefore, I just suggest not supporting `MinGW i686`

SchrodingerZhu:
> I wonder if there are compatibility issues with linking in the `mincore.lib`. If you configure cmake with
> 
> ```
> -DWIN8COMPAT=TRUE
> ```
> 
> It won't try to use `VirtualAlloc2`, or pull in `mincore.lib`.
> 
> Also, I am a little concerned the changes you are making will affect the Windows-clang target.

thanks for reminding me of clang. I am not quite sure about it either; I remember on windows, clang will provide compatibility with MSVC rather than GNU, but it may still define `__GNU__`? MSYS2 environment also provides a clang toolchain, I have not checked the library with that.

I do not quite understand the issue of `mincore.lib`, can you provide more details?

SchrodingerZhu:
tried to compile with `-DWIN8COMPAT=TRUE` and `MinGW x86-64` on Windows 10 Build 19577.rs_prerelease.200228-1439, everything works fine. Unfortunately, I do not have a Win 8 machine for further tests.

mjp41:
> For MinGW x86_64, mcx16 works fine; but for MinGW i686, the mcx16 feature cannot be enabled even with -mcx16. Therefore, I just suggest not supporting MinGW i686

This makes complete sense.  We are only supporting 32bit on Windows with MSVC at the moment.  We don't have any 32bit Linux support. 

I think it it works without WIN8COMPAT that is fine. I had thought it was failing at runtime, and this might fix it.  But if it is running, then it is not needed.

I'll try to find some time to see if Windows Clang is setting __GNUC__.

CMakeLists.txt
```diff
@@ -72,6 +72,8 @@ if(NOT MSVC)
    target_compile_options(snmalloc_lib INTERFACE -mcx16)
   elseif(${CMAKE_SYSTEM_PROCESSOR} STREQUAL "amd64")
    target_compile_options(snmalloc_lib INTERFACE -mcx16)
+  elseif(${CMAKE_SYSTEM_PROCESSOR} STREQUAL "AMD64")
```
davidchisnall: Can this not be a case-insensitive string compare (e.g. use TOLOWER on the processor and then compare against amd64)?

SchrodingerZhu:
@davidchisnall 

This rust crate can be referred as a test, it also provides freebsd and mingw environment.

MSVC/MinGW/Linux/MacOS: [travis ci](https://travis-ci.com/SchrodingerZhu/snmalloc-rs)

FreeBSD: [Build Status](https://cirrus-ci.com/github/SchrodingerZhu/snmalloc-rs)

However, this is not a good solution. It would be better to have an extra pipeline here.

PR #138, opened 2020-03-09
Remote dealloc refactor.

mjp41:
This performs a small refactor on the remote dealloc path to split the common case out.

### DIFF
--- a/src/mem/alloc.h
+++ b/src/mem/alloc.h
@@ -492,8 +492,8 @@ namespace snmalloc
       /**
        * The total amount of memory we are waiting for before we will dispatch
        * to other allocators. Zero or negative mean we should dispatch on the
-       * next remote deallocation. This is initialised to the 0 so that we 
-       * always hit a slow path to start with, when we hit the slow path and 
+       * next remote deallocation. This is initialised to the 0 so that we
+       * always hit a slow path to start with, when we hit the slow path and
        * need to dispatch everything, we can check if we are a real allocator
        * and lazily provide a real allocator.
        */
--- a/src/mem/alloc.h
+++ b/src/mem/alloc.h
@@ -1280,9 +1280,8 @@ namespace snmalloc
       remote_dealloc_slow(target, p, sizeclass);
     }
 
-    SNMALLOC_SLOW_PATH
-    void remote_dealloc_slow(
-      RemoteAllocator* target, void* p, sizeclass_t sizeclass)
+    SNMALLOC_SLOW_PATH void
+    remote_dealloc_slow(RemoteAllocator* target, void* p, sizeclass_t sizeclass)
     {
       assert(target->id() != id());
 
--- a/src/mem/alloc.h
+++ b/src/mem/alloc.h
@@ -1281,7 +1281,8 @@ namespace snmalloc
     }
 
     SNMALLOC_SLOW_PATH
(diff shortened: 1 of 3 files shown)


mjp41:
@darach, @Licenser, this might improve your performance.  The perf trace you sent me showed 2.68% on this path, so hopefully you might see a difference.

Licenser:
Nice! We’ll bench this tomorrow:) thanks!

mjp41:
On x64 Linux this is now, 15 instructions to the branch to decide it is a remote dealloc, and then a further 13 instructions (including one branch, not taken in common case) to complete the remote deallocation.  If it actually posts it is a lot more instructions, but that is very infrequent.

Previously, it was a tangled mess of assembly, and probably twice as many instructions.

src/mem/alloc.h
```diff
@@ -1264,31 +1265,42 @@ namespace snmalloc
...
         remote.dealloc_sized(target->id(), offseted, sz);
         return;
       }
+
+      remote_dealloc_slow(target, p, sizeclass);
+    }
+
+    SNMALLOC_SLOW_PATH
+    void remote_dealloc_slow(
+      RemoteAllocator* target, void* offseted, sizeclass_t sizeclass)
```
plietar: The call to this function, a few lines up, passes p, ie. the non-offseted address,  as the second argument. 

I would duplicate the call to `apply_cache_friendly_offset`, once on the fast path inside the `if (remote.capacity > 0)` branch and once here, so you don’t have to also do a `remove_cache_friendly_offset`

mjp41: Agreed. Fixed.  Thanks for the feedback.  I had thought offset would have lead to better code, but actually it didn't.

Licenser:
I know this is merged already but I wanted to add some results from our side:

## old
```
#[Mean       =    109794.14, StdDeviation   =     92536.16]
#[Max        =       374783, Total count    =     69517595]
#[Buckets    =           30, SubBuckets     =         3968]


Throughput: 417.1 MB/s
```
pre

## new
```
#[Mean       =    109568.44, StdDeviation   =     96808.94]
#[Max        =       430079, Total count    =     68218199]
#[Buckets    =           30, SubBuckets     =         3968]


Throughput: 409.3 MB/s
```

post

mjp41:
Interesting, so the perf numbers show it is spending much less time in remote dealloc, 2.68% -> 0.92%, but this hasn't translated into a throughput win for you?  How much noise is there in the "Throughput" number?

Licenser:
Sorry I did a copy & paste oopsie, when pasting the numbers the throughput for old and new were swapped

Licenser:
the new code is ~2% faster

mjp41:
Phew, you had me questioning my ability to guess asm speed (which obviously is pretty questionable ;-) ).  2% that is pretty good.  More than I expected.

Licenser:
Ja 2% is really nice! I think snmalloc is a really really good fit for tremor :D

Licenser:
adding a latency plot (ignore the unit on the left the hdr histogram plotter is broken)
image (1)

darach:
The profile is good, but the Latency axis should read nanos. We use HDR Histogram https://hdrhistogram.github.io/HdrHistogram/plotFiles.html to plot.

PR #139, opened 2020-03-10
Extra helper functions for dllist

ashamis:
Adding extra helper function for dllist


src/ds/dllist.h
```diff
@@ -66,6 +66,32 @@ namespace snmalloc
     T* tail = Terminator();
 
   public:
+    virtual ~DLList()
```
mjp41: Why do you need the desctructor to be `virtual`?

ashamis: I can remove that but the class is not marked final so theoretically someone could inherit from it.

mjp41: I would prefer final and non-virtual.  You have increased the size of the allocator data structure with this change.  Every DLL now has a vtable, i.e. we have gone from 
```
      sizeof(DLList) == (sizeof(void*) * 2)
```
to
```
      sizeof(DLList) == (sizeof(void*) * 3)
```

ashamis: done

PR #140, opened 2020-03-11
Assert not going through the PAL

achamayou:
Happened in #138, problematic for certain platforms.

I need to work out a good way to build a test for this.


achamayou:
Some of the CI jobs are unhappy:

image

achamayou:
Trying to fix that with https://github.com/microsoft/snmalloc/pull/141

achamayou:
CI should pass after #141 is merged.

PR #142, opened 2020-03-13
AAL, basic arm implementation proposal.

devnexen:


mjp41:
@nwf any comments?

src/aal/aal_arm.h
```diff
@@ -0,0 +1,53 @@
...
+    {
+      __asm__ volatile("yield");
+    }
+
+    /**
+     * Issue a prefetch hint at the specified address.
+     */
+    static inline void prefetch(void* ptr)
+    {
+      __builtin_prefetch(ptr, 0, 1);
```
davidchisnall: Is this needed?  If you don't implement `prefetch` then you'll get this implementation from `AAL_Generic`.

src/aal/aal_arm.h
```diff
@@ -0,0 +1,53 @@
...
+    /**
+     * Return a cycle counter value.
+     * on ARM cpu counters are accessible only in privileged mode
+     */
+    static inline uint64_t tick()
+    {
+      struct timespec n = {0, 0ul};
+      clock_gettime(CLOCK_MONOTONIC, &n);
+
+      return static_cast<uint64_t>((n.tv_sec) * (1000000000ul * n.tv_nsec));
```
davidchisnall: Given that this is a fallback path, it should probably be in the generic version.

It looks as if the version in `AAL_Generic` is will unconditionally use `__builtin_readcyclecounter` if it exists, so we probably need to make that conditional on the builtin supporting a in unprivileged mode, which probably merits a new AAL feature flag.

src/ds/bits.h
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <limits>
+#include <cstddef>
```
davidchisnall: What is this needed for?

src/pal/pal_posix.h
```diff
@@ -4,6 +4,7 @@
 #include "../mem/allocconfig.h"
 
 #include <string.h>
+#include <stdlib.h>
```
davidchisnall: What is this needed for?

src/ds/bits.h
devnexen: size_t definition

src/pal/pal_posix.h
devnexen: abort

src/ds/bits.h
davidchisnall: Why don't we hit that on non-ARM?

src/pal/pal_posix.h
davidchisnall: Why don't we hit that on non-ARM?

src/aal/aal_arm.h
nwf: This is not exactly the right place to ask, but it seems roughly on topic: is there guidance for when to use a AAL/PAL feature flag vs. inheritance and overriding?

nwf:
Other than my off-topic question, this looks sensible enough to me.  Thanks for the ping. :)

src/aal/aal_arm.h
devnexen: ok but there is no inheritance unlike pal though.

src/ds/bits.h
devnexen: arm esp 32 bits is pretty limited arch in comparison. header arranged differently and all. that was coming from compiler build warning would not have caught this otherwise.

src/aal/aal_arm.h
davidchisnall: Not exactly.  It's applied centrally.  The per-arch files define an AAL_Arch type, the AAL that the rest of the code uses is defined with the `using Aal = AAL_Generic<AAL_Arch>;` line in aal.h.  We (should) never use `AAL_arm` directly, we will use only `AAL_Generic<AAL_arm>`.  The `prefetch` method in `AAL_Generic` will not call `AAL_arm::prefetch` unless we've disabled builtins for the AAL (in which case, we shouldn't be using it here either).

src/aal/aal_arm.h
davidchisnall: For the AAL, it's relatively simple: the per-arch version is more-or-less a mixin and so you can't override the generic behaviour from it, you can only provide feature flags that tell the generic version when to call into the arch-specific one.  We should probably be using a method implementation check trait to determine whether to call the arch version or use a builtin, rather than always using the builtin where available.

Note that the AAL is currently an abstraction over one thing and so is almost certainly the wrong interface.  I'm very happy to accept PRs that improve it!  It was intended as a starting point so that we could figure out what the real design should look like when we had 2-3 examples (so please do submit a PR for the MIPS one!).

For the PAL, the rule of thumb is that we should override things when we want to provide a complete implementation of something that's different on all platforms, but use feature flags when we want to have most of a method shared by provide extension hooks, or mechanism for skipping something, or to allow callers to conditionally call a particular function only if it exists.  For example, we use the `AlignedAllocation` feature flag to indicate to callers that they can call the function that requests very strongly aligned memory, `LowMemoryNotification` to allow callers to register to receive low memory notifications to lazily decommit memory if memory is low, and `LazyCommit` to determine whether we should allocate a large contiguous virtual address range for the pagemap and expect it to be backed by writealbe zeroed pages on demand. 

The `AAL_Generic` class is also used to define the interface for the AAL.  Methods on the `AAL_Arch` class are intended to be used only via the `AAL_Generic` class.  I'm not 100% sure this is the right design, but it seemed like a good idea at the time...

src/aal/aal_arm.h
devnexen: ah right ... I was using wrong compiler gcc 7 ... now the builtins are being detected.

src/aal/aal.h
```diff
@@ -11,6 +12,10 @@
 
 #if defined(__arm__) || defined(__aarch64__)
 #  define PLATFORM_IS_ARM
+   /**
+    * on ARM cpu counters are accessible only in privileged mode
+    */
+#  define SNMALLOC_NO_ALL_CPUCOUNTERS
```
davidchisnall: Can this be an AAL feature flag?  I'd rather minimise the pollution to the global namespace where possible.

src/aal/aal.h
```diff
@@ -63,7 +68,12 @@ namespace snmalloc
      */
     static inline uint64_t tick()
     {
-#if __has_builtin(__builtin_readcyclecounter) && \
+#if defined(SNMALLOC_NO_ALL_CPUCOUNTERS)
+      struct timespec n = {0, 0ul};
```
davidchisnall: This can be a constexpr if based on the AAL feature flag.

src/aal/aal_arm.h
```diff
@@ -31,21 +31,13 @@ namespace snmalloc
     /**
      * Issue a prefetch hint at the specified address.
      */
-    static inline void prefetch(void* ptr)
+    static inline void prefetch(void*)
```
davidchisnall: This method should be removed entirely.

src/aal/aal_arm.h
```diff
@@ -31,21 +31,13 @@ namespace snmalloc
...
     {
-      __builtin_prefetch(ptr, 0, 1);
     }
 
-    /**
-     * Return a cycle counter value.
-     * on ARM cpu counters are accessible only in privileged mode
-     */
     static inline uint64_t tick()
```
davidchisnall: As should this one.

src/aal/aal.h
```diff
@@ -59,6 +68,13 @@ namespace snmalloc
      */
     static inline uint64_t tick()
     {
+      if constexpr((Arch::aal_features & NoCpuCycleCounters) == NoCpuCycleCounters)
+      {
+        struct timespec n = {0, 0ul};
+        clock_gettime(CLOCK_MONOTONIC, &n);
```
mjp41: This is posix specific, can you add an 
```
#ifdef _WIN32
#  error "Implement Windows Version"
#else
   ....
#endif`
```
And when someone wants this for Windows on Arm they should get a good error message.  This should also fix the build issues for Windows Clang.

mjp41:
LGTM, @davidchisnall if you are happy, then please merge.

mjp41:
Sorry, my error suggestion didn't work out. :(

devnexen:
> Sorry, my error suggestion didn't work out. :(

Would pragma message fit better ?

src/aal/aal.h
mjp41: Could we use 
```
std::chrono::high_resolution_clock::now()
```
this is actually cross platform.  Then it should work on any platform.

src/pal/pal_posix.h
mjp41:
```suggestion
#include <stdlib.h>
#include <string.h>
```

src/ds/bits.h
mjp41:
```suggestion
#include <cstddef>
#include <limits>
```

src/aal/aal_arm.h
```diff
@@ -0,0 +1,33 @@
...
+  /**
+   * ARM-specific architecture abstraction layer.
+   */
+  class AAL_arm
+  {
+  public:
+    /**
+     * Bitmap of AalFeature flags
+     */
+    static constexpr uint64_t aal_features = IntegerPointers | NoCpuCycleCounters;
```
mjp41: Another Clangformat issue.  I don't know why it didn't report this one previously.

```suggestion
    static constexpr uint64_t aal_features =
      IntegerPointers | NoCpuCycleCounters;
```

Issue #66, opened 2019-07-09
Track slab with bump pointer in a new per allocator structure


mjp41:
With #65 we now use a queue of slabs for selecting free lists.  The queue is also how we pick up the bump pointer, this means we are not really benefiting from the bump ptr.

We should have a separate per allocator per size class structure for the slab, we are currently bump allocating from.  This would reduce the memory usage and allow us to simplify the data in the meta slab, and would remove some conditionals on the slower paths.  

We could just store the pages that have been bump allocated in the meta slab, i.e. just require 4bits.

PR #143, opened 2020-03-16
Improve slow path performance for allocation

mjp41:
* Make separate bump allocator ptr per thread and size class.
   - Now a slab in the per-class list is guaranteed to have at least one free allocation, not including 
     bump allocation.
   - This enables more separation between how we are trying to find memory

* Refactor code to 
   - Separate checks and slow paths to favour tail calls
   - Generally make all slow paths tail calls to improve codegen

The general form of allocation is now
1. Local free list for this allocator
2. Grab next free list for this allocator
3. Bump allocate from local bump ptr
4. Grab new slab to bump allocate from

There are special cases interleaved into this

  * Handle message queue
  * If the thread local allocator needs initialising.

The first occurs between 1 and 2, and the second occurs between, 3 and 4.

### DIFF
--- a/src/ds/cdllist.h
+++ b/src/ds/cdllist.h
@@ -1,8 +1,9 @@
 #pragma once
 
+#include "defines.h"
+
 #include <cstdint>
 #include <type_traits>
-#include "defines.h"
 
 namespace snmalloc
 {
@@ -44,7 +45,7 @@ namespace snmalloc
       // As this is no longer in the list, check invariant for
       // neighbouring element.
       next->debug_check();
-      
+
 #ifndef NDEBUG
       next = nullptr;
       prev = nullptr;
--- a/src/ds/cdllist.h
+++ b/src/ds/cdllist.h
@@ -18,6 +18,9 @@ namespace snmalloc
     CDLLNode* prev;
 
   public:
+    /**
+     * Single element cyclic list.  This is the empty case.
+     **/
     CDLLNode()
     {
       next = this;
@@ -29,12 +32,17 @@ namespace snmalloc
       return next == this;
     }
 
+    /**
+     * Removes this element from the cyclic list is it part of.
+     **/
     SNMALLOC_FAST_PATH void remove()
     {
       SNMALLOC_ASSERT(!is_empty());
       debug_check();
       next->prev = prev;
       prev->next = next;
+      // As this is no longer in the list, check invariant for
+      // neighbouring element.
       next->debug_check();
       
 #ifndef NDEBUG
@@ -73,6 +81,11 @@ namespace snmalloc
       debug_check();
     }
 
+    /**
+     * Checks the lists invariants
+     *   x->next->prev = x
+     * for all x in the list.
+     **/
     void debug_check()
     {
 #ifndef NDEBUG
--- a/src/ds/cdllist.h
+++ b/src/ds/cdllist.h
@@ -11,7 +11,7 @@ namespace snmalloc
    * Special class for cyclic doubly linked non-empty linked list
    *
    * This code assumes there is always one element in the list. The client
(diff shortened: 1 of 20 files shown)


src/mem/threadalloc.h
```diff
@@ -40,16 +40,27 @@ namespace snmalloc
 
   /**
    * Function passed as a template parameter to `Allocator` to allow lazy
-   * replacement.  In this case we are assuming the underlying external thread
-   * alloc is performing initialization, so this is not required, and just
-   * always returns nullptr to specify no new allocator is required.
-   */
-  SNMALLOC_FAST_PATH void* lazy_replacement(void* existing)
+   * replacement. This function returns true, if the allocated passed in,
```
davidchisnall: The allocated passed in what?

src/mem/alloc.h
```diff
@@ -1028,14 +1034,14 @@ namespace snmalloc
...
+        return small_alloc_next_free_list<zero_mem, allow_reserve>(sizeclass);
 
       return small_alloc_mq_slow<zero_mem, allow_reserve>(sizeclass);
     }
 
     /**
      * Slow path for handling message queue, before dealing with small
-     * allocation request. 
+     * allocation request.
      **/
```
davidchisnall: Missed one.

src/mem/alloc.h
```diff
@@ -1102,8 +1108,7 @@ namespace snmalloc
      * list.
      **/
```
davidchisnall: Uh, two...

src/mem/alloc.h
```diff
@@ -1140,20 +1144,20 @@ namespace snmalloc
 
     /**
      * Allocates a new slab to allocate from, set it to be the bump allocator
-     * for this size class, and then builds a new free list from the thread 
+     * for this size class, and then builds a new free list from the thread
      * local bump allocator and service the request from that new list.
      **/
```
davidchisnall: three?

mjp41:
Any idea why the pipeline "Microsoft.snmalloc" is not firing?

src/ds/cdllist.h
```diff
@@ -11,7 +11,7 @@ namespace snmalloc
    * Special class for cyclic doubly linked non-empty linked list
    *
    * This code assumes there is always one element in the list. The client
-   * must ensure there is a sentinal element.
+   */ust ensure there is a sentinal element.
    **/
```
davidchisnall: Not GitHub is highlighting them as bad!

src/ds/cdllist.h
```diff
@@ -11,7 +11,7 @@ namespace snmalloc
    * Special class for cyclic doubly linked non-empty linked list
    *
    * This code assumes there is always one element in the list. The client
-   * must ensure there is a sentinal element.
+   */ust ensure there is a sentinal element.
```
davidchisnall: I think you deleted the thing in the wrong thing...

src/ds/cdllist.h
mjp41: Don't know how that happened. I just performed a find and replace.  Reverting and doing the same worked.

mjp41:
This PR addresses #66

SchrodingerZhu:
A segfault during linking?

mjp41:
@SchrodingerZhu this is in the self-host test.  So we are LD_PRELOADing the snmalloc we just built, and it is crashing.  This is one of the harder tests to debug, but also provides us a good piece of coverage of the code.

This change is pretty big, so I wanted to go above the standard CI level, hence why it has been hanging around for a week or so.  I was going to run a lot of benchmarks through it, but haven't had the time yet.  With the changes to CI I guess we have hit an issue independently.

PR #145, opened 2020-03-18
Containerise Linux build

achamayou:
Linux CI in docker, groundwork to add QEmu/ARM build, and clang upgrade along the way.

Somewhat slower than before because Pipelines still does not have an image cache, but builds are substantially more reproducible, which is sometimes nice.


mjp41:
The CI pipeline machines have more than the stock Ubuntu. Looks like you need to add CMake at least

achamayou:
@davidchisnall if you and others would like to help own/maintain the docker hub org under which the CI images are stored, please send me your username so I can add it to https://hub.docker.com/orgs/snmallocciteam

PR #147, opened 2020-03-18
Workaround for QEMU behaviour.

mjp41:
This is not hooked up to anything yet.  Once @achamayou has completed the containerised build #145, then we can add the AARCH64 QEMU build, and this will be used.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,6 +7,7 @@ option(USE_MEASURE "Measure performance with histograms" OFF)
 option(EXPOSE_EXTERNAL_PAGEMAP "Expose the global pagemap" OFF)
 option(EXPOSE_EXTERNAL_RESERVE "Expose an interface to reserve memory using the default memory provider" OFF)
 option(SNMALLOC_RUST_SUPPORT "Build static library for rust" OFF)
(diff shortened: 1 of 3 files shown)


SchrodingerZhu:
Seems that this part also has an issue?
https://github.com/microsoft/snmalloc/blob/a6d6eecf22efcddb69f85f898eb9c413d0b1e687/src/aal/aal.h#L52
See: 
https://travis-ci.org/github/SchrodingerZhu/snmalloc-rs/jobs/664278142#L405

mjp41:
So that works in Clang-9 for Aarch.  I didn't test GCC-9.

I guess gcc is not providing a prefetch builtin.  I think the best would be to add a prefetch that does nothing in the Arm AAL layer.  @Davidchisnall do you agree?

mjp41:
@SchrodingerZhu could you check if this works for you now.  I have tested with gcc-8 and clang-9 inside QEMU.

SchrodingerZhu:
In progress: https://github.com/SchrodingerZhu/snmalloc-rs/pull/29

SchrodingerZhu:
fixed.

PR #148, opened 2020-03-19
ARM CI

achamayou:
This adds an ARM CI using QEMU. Sadly it's not possible to completely unify it with the regular Linux CI, because the docker setup is slightly odd in a way Azure Pipelines doesn't like.

I'm still planning to pull some bits out into scripts (build and test.sh), so it's reasonably common.


achamayou:
@mjp41 I think this is ok now, the build and test are common across x86/ARM, so adding new arch/compilers should be relatively easy in the future.

Before I merge though, can you confirm the change in aal_arm.h is ok?

PR #150, opened 2020-03-19
Fix for ARM 32 build, using PLD instruction for the preload.

devnexen:


mjp41:
@devnexen thanks for this.  #148 conflicts which I have applied.  But I am happy to take this change.

I also want to setup 32bit ARM CI, so we don't break this in the future.  Are there any requirements on your side for what image we should use for that?  It will have to be QEmu based Linux for now.

achamayou:
@mjp Ubuntu ARM images seem to come in two flavours, armhf which includes hw support for fp, and armel, which doesn't. Adding a CI with either (or both) should be equally easy.

achamayou:
#151, may or may not need some tweaks, let's find out.

mjp41:
Superseeded by #151, thanks so much for spotting and fixing.

PR #151, opened 2020-03-20
CI for both arms

achamayou:
Picked armhf, for no particular reason. Is that ok, or do we want both armel and armhf?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,8 @@
 cmake_minimum_required(VERSION 3.8)
 project(snmalloc C CXX)
 
+include(CheckCXXCompilerFlag)
+
 option(USE_SNMALLOC_STATS "Track allocation stats" OFF)
 option(SNMALLOC_CI_BUILD "Disable features not sensible for CI" OFF)
 option(USE_MEASURE "Measure performance with histograms" OFF)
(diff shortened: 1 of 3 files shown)


achamayou:
So for some reason, clang-9 on armhf is not happy with -march=native:

```

/usr/bin/clang++-9  -DIS_ADDRESS_SPACE_CONSTRAINED -DSNMALLOC_CI_BUILD -DSNMALLOC_QEMU_WORKAROUND -DUSE_POSIX_COMMIT_CHECKS -DUSE_SNMALLOC_STATS -I../src -std=c++17 -stdlib=libstdc++ -g   -Wsign-conversion -Wall -Wextra -Werror -Wundef -fno-exceptions -fno-rtti -g -ftls-model=initial-exec -fomit-frame-pointer -march=native -std=gnu++1z -MD -MT CMakeFiles/func-two_alloc_types-1.dir/src/test/func/two_alloc_types/alloc2.cc.o -MF CMakeFiles/func-two_alloc_types-1.dir/src/test/func/two_alloc_types/alloc2.cc.o.d -o CMakeFiles/func-two_alloc_types-1.dir/src/test/func/two_alloc_types/alloc2.cc.o -c ../src/test/func/two_alloc_types/alloc2.cc
clang: error: the clang compiler does not support '-march=native
```

CMakeLists.txt
```diff
@@ -145,12 +147,14 @@ if(NOT DEFINED SNMALLOC_ONLY_HEADER_LIBRARY)
...
   else()
     add_compile_options(-fno-exceptions -fno-rtti -g -ftls-model=initial-exec -fomit-frame-pointer)
-    if(${CMAKE_SYSTEM_PROCESSOR} STREQUAL "x86_64")
-     add_compile_options(-march=native)
-    elseif(${CMAKE_SYSTEM_PROCESSOR} STREQUAL "x86")
-     add_compile_options(-march=native)
-    elseif(${CMAKE_SYSTEM_PROCESSOR} MATCHES "arm")
-     add_compile_options(-march=native)
+
+    if((${CMAKE_SYSTEM_PROCESSOR} STREQUAL "x86_64") OR
```
achamayou: @mjp I believe we can get rid of this check and just rely on whether the compiler supports -march=native, what do you think?

mjp41: Nice if it works ;-)

achamayou: It does... so far? :)

Issue #155, opened 2020-03-23
PALOpenEnclave::reserve should the loop condition be inverted?


anakrish:
The `reserve` function is currently implemented thus: https://github.com/microsoft/snmalloc/blob/0a081cacd649e138cb83edb56f773c7ecf6b6e2d/src/pal/pal_open_enclave.h#L38-L49

The loop condition is currently:
`while (oe_base.compare_exchange_strong(old_base, next_base))`.

Should the condition instead be negated as:
`while ( ! oe_base.compare_exchange_strong(old_base, next_base))`
so that the loop immediately terminates after a successful update of `oe_base`?

mjp41:
You are correct.  That worked in the unit tests as I haven't got anything covering racing on this code. I'll submit a PR to address this and #156.

mjp41:
Closed by #157.

Issue #156, opened 2020-03-24
PALOpenEnclave: When out of memory NULL ought to be returned instead of calling oe_abort


anakrish:
The following test fails with `USE_SNMALLOC` and heap-pages bumped up to 4096.

https://github.com/openenclave/openenclave/blob/17f9ee3b4d7a84125b5fe083a92013de2c83c190/tests/stdc/enc/enc.cpp#L275-L276

mjp41:
Closed by #157

PR #157, opened 2020-03-24
OE fixes

mjp41:


mjp41:
Fixes #155 and #156

mjp41:
@anakrish I believe this should fix the issues you found.  Please could you confirm your happy with this fix for Open Enclave.

mjp41:
CI has passed twice from a manual trigger.  The failure shown of this commit appears to be a failure to acquire a VS2019 VM.

anakrish:
@mjp41 Thanks for fixing these. I confirmed that the fixes work.
Note, even with malloc returning NULL for an unfulfillable allocation, tests/stdc still fails because the allocation failure callback is not invoked. That will work once we formally support pluggable allocators in OE

PR #158, opened 2020-03-27
Increase Remote batch size

mjp41:
The remote batch size has not changed since the fast path optimisations.
The optimisations mean we are checking the queue considerably less
often, so the batch should be larger.  This has a dramatic improvement
on performance on a few of the mimalloc microbenchmarks.

It is set to 4096 as this should cover the worse case scenario of only
remote deallocation at 16 bytes for the 2^16 slab size.

### DIFF
--- a/src/ds/address.h
+++ b/src/ds/address.h
@@ -63,7 +63,7 @@ namespace snmalloc
   template<size_t alignment, typename T = void>
   SNMALLOC_FAST_PATH T* pointer_align_down(void* p)
   {
-    if constexp (alignment == 1)
+    if constexpr (alignment == 1)
       return static_cast<T*>(p);
     else
       static_assert(alignment > 0);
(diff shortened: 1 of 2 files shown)


mjp41:
@darach @licenser I think this will improve your scenario.  I was investigating a drop in perf on some producer/consumer for micro-benchmarks.  My running of your benchmark is quiet noisy at the moment, but the time spent inside the `handle_message_queue_inner` has dropped considerably with this change.

Licenser:
That sounds awesome :) we'll give it a try later! Thank you so much :)!

mjp41:
So on the mimalloc-bench configuration for xmalloc-test we get:
```
# --------------------------------------------------
# benchmark allocator elapsed rss user sys page-faults page-reclaims
xmalloc-testN mi    0.662 204872 56.41 0.64 0 375
xmalloc-testN sn    0.956 314744 50.27 2.85 0 464
xmalloc-testN sn-pr 0.488 314712 46.50 4.96 0 466
```
The time effectively halves going from 0.956, worse than mimalloc, to considerably better at 0.488.

I think there is room to do a parameter sweep on this value, but I don't have time to investigate at the moment. I think this value is reasonable.

PR #159, opened 2020-03-31
Small changes

mjp41:
* Use NoZero for fresh pages
* MADV_DONTNEED only use for greater than a whole slab
* Simplify free list threading code


mjp41:
FYI: This has passed CI just not the first time as one of the CI machines appeared to go away.

PR #160, opened 2020-03-31
Add checks for deallocation with size

mjp41:
Ensure that the size deallocated with is the correct size. If the size
supplied is wrong, then internal invariants can break.

### DIFF
--- a/src/mem/alloc.h
+++ b/src/mem/alloc.h
@@ -210,6 +210,24 @@ namespace snmalloc
       UNUSED(size);
       return free(p);
 #else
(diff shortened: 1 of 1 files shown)


src/mem/alloc.h
```diff
@@ -210,6 +210,24 @@ namespace snmalloc
       UNUSED(size);
       return free(p);
 #else
+#  if defined(CHECK_CLIENT)
```
davidchisnall: Can this duplicated code be pulled into a separate (inline) function?

Issue #50, opened 2019-05-13
Replace ABA primitives with LL/SC emulation?


nwf:
CHERI builds primarily on RISC platforms, specifically MIPS and RISC V, which have native LL/SC support.  AFAICT, MIPS R4K promises relatively strong LL/SC operation (permitting loads and stores from the core having done the LL); RISC V promises little but presumably real implementations will also be relatively strong.  Anyway, on these platforms, I think `ds/aba.h` is more of a hindrance than a help; the sole user, `ds/mpmcstack.h` is forced to phrase its operations in terms of CAS and so will always require ABA protection.  While the existing non-X86 paths in `ds/aba.h` optimistically fall back to just a pointer-sized CAS, this fails to deliver effective ABA protection

I am unsure what the correct answer is, but I think, at least broadly, that `mpmcstack` should be rewritten in terms of a LL/SC primitive that can be provided by some platform-specific code.  I haven't come up with a better answer than something like this, tho', which isn't great, so other thoughts are more than welcome.  Notably, this approach, I think, relies very, very heavily on the compiler's ability to inline to produce good code.

```C++
  template<typename T, Construction c = RequiresInit>
  class LLSC
  {
    /* Cmp defined as in existing ds/aba */
  private:
   Cmp value;

  public:
    inline bool llsc(std::function<bool, T*> f, bool* out)
    {
      T t;
      // Load link t from this, or load t and the generation counter
      *out = f(&t);
      if (*out)
      {
        // Store conditional or CAS2 with incremented generation counter
      }
      else
      {
        return true; // will cause exit from calling loop, which can look at *out
      }
    }
  }
```

mjp41:
We really require the only way to make a `Cmp` is `Cmp read()`, then I think the current interface may be fine for LL/SC.  I.e. the 
```C    
using Cmp = T*;
```
Is not good enough.  But the idea is that you can only compare exchange with a previously read value.  The previously read value would be LL, and then the compare exchange would be a SC. 

I am unsure on the semantics of LL/SC around multiple LL operations, i.e. what would happen if a thread has multiple ABA operations in progress, then I am not sure this API would protect enough.

nwf:
Most implementations permit only a single LL/SC in flight at once, which is a bit of a travesty, but is what we have.  That said, I have been hacking up a LL/SC implementation of `ds/aba.h` following your suggestion that `read()` is LL and will let you know how it goes.

mjp41:
Great thanks.  

[...]

PR #161, opened 2020-04-01
Defensive code for alloc/dealloc during TLS teardown

mjp41:
If an allocation or deallocation occurs during TLS teardown, then it is
possible for a new allocator to be created and then this is leaked. On
the mimalloc-bench mstressN benchmark this was observed leading to a
large memory leak.

This fix, detects if we are in the TLS teardown phase, and if so,
the calls to alloc or dealloc must return the allocator once they have
perform the specific operation.

[Add:] Also fixes unimplemented ABA primitive for ARM.  Currently uses locking.

[Add:] Some debug infrastructure to deal with Posix platforms.

### DIFF
--- a/src/ds/aba.h
+++ b/src/ds/aba.h
@@ -67,13 +67,13 @@ namespace snmalloc
         error("Only one inflight ABA operation at a time is allowed.");
       operation_in_flight = true;
 
-      return
-        Cmp{{independent.ptr.load(std::memory_order_relaxed),
-            independent.aba.load(std::memory_order_relaxed)},
-            this};
+      return Cmp{{independent.ptr.load(std::memory_order_relaxed),
+                  independent.aba.load(std::memory_order_relaxed)},
+                 this};
     }
 
-    struct Cmp {
+    struct Cmp
+    {
       Linked old;
       ABA* parent;
 
@@ -120,6 +120,7 @@ namespace snmalloc
   {
     std::atomic<T*> ptr = nullptr;
     std::atomic_flag lock = ATOMIC_FLAG_INIT;
+
   public:
     struct Cmp;
 
@@ -134,6 +135,7 @@ namespace snmalloc
     struct Cmp
     {
       ABA* parent;
+
     public:
       T* ptr()
       {
--- a/src/ds/aba.h
+++ b/src/ds/aba.h
@@ -2,13 +2,24 @@
 
 #include "bits.h"
 
+/**
+ * This file contains an abstraction of ABA protection. This API should be
+ * implementable with double-word compare and exchange or with load-link
+ * store conditional.
+ *
+ * We provide a lock based implementation.
+ */
 namespace snmalloc
 {
+#ifdef PLATFORM_IS_X86
+  // LL/SC typically can only perform one operation at a time
+  // check this on other platforms using a thread_local.
+  inline thread_local bool operation_in_flight = false;
+
   template<typename T, Construction c = RequiresInit>
   class ABA
   {
   public:
-#ifdef PLATFORM_IS_X86
     struct alignas(2 * sizeof(std::size_t)) Linked
     {
       T* ptr;
@@ -28,21 +39,12 @@ namespace snmalloc
       sizeof(Linked) == (2 * sizeof(std::size_t)),
       "Expecting ABA to be the size of two pointers");
 
-    using Cmp = Linked;
-#else
-    using Cmp = T*;
-#endif
-
   private:
-#ifdef PLATFORM_IS_X86
     union
     {
       alignas(2 * sizeof(std::size_t)) std::atomic<Linked> linked;
       Independent independent;
     };
-#else
-    std::atomic<T*> raw;
-#endif
 
   public:
     ABA()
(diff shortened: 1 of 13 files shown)


mjp41:
Moved to draft due to failures on Arm Debug.

mjp41:
I have addressed #50 with a locking based implementation. This should fix the ARM issues we were seeing.

src/mem/threadalloc.h
```diff
@@ -243,21 +265,24 @@ namespace snmalloc
...
-      // to say stop doing this, or just give them the initialised version.
-      return local_alloc;
+      local_alloc = current_alloc_pool()->acquire();
     }
-    local_alloc = current_alloc_pool()->acquire();
-    SNMALLOC_ASSERT(local_alloc != get_GlobalPlaceHolder());
-    ThreadAlloc::register_cleanup();
-    return local_alloc;
+    // Check if we have already run the destructor for the TLS.  If so, then
+    // we need the caller to handle this specially as we are in teardown.
```
plietar, line 281: I’m not sure what “the caller” means in this context. Is it the caller of register_cleanup, aka the current function? Because I don’t see what the caller of init_thread_allocator needs to do

src/mem/threadalloc.h
```diff
@@ -243,21 +265,24 @@ namespace snmalloc
    * The simple check that this is the global placeholder is inlined, the rest
    * of it is only hit in a very unusual case and so should go off the fast
    * path.
+   * The second component of the return indicates if this TLS is being torndown.
    */
-  SNMALLOC_SLOW_PATH inline void* init_thread_allocator()
+  SNMALLOC_FAST_PATH void* init_thread_allocator(std::function<void*(void*)>& f)
```
plietar, line 270: What’s does the code gen look for this? Does the call to `f` get inlined?

src/ds/aba.h
```diff
@@ -53,66 +55,104 @@ namespace snmalloc
...
+      return Cmp{{independent.ptr.load(std::memory_order_relaxed),
+                  independent.aba.load(std::memory_order_relaxed)},
+                 this};
     }
 
-    bool compare_exchange(Cmp& expect, T* value)
+    struct Cmp
     {
-#ifdef PLATFORM_IS_X86
+      Linked old;
```
plietar, line 77: I think this should be an `Independent`, otherwise reading it will result in sequentially consistent atomic loads (obviously that generates the same assembly anyway on x86).

src/mem/globalalloc.h
```diff
@@ -7,7 +7,7 @@
 namespace snmalloc
 {
   inline bool needs_initialisation(void*);
-  void* init_thread_allocator();
+  void* init_thread_allocator(std::function<void*(void*)>&);
```
plietar, line 10: `init_thread_allocator` is hardcoded to use `current_alloc_pool`, so I don't think we should default `AllocPool` and `make_alloc_pool` to use these functions.

We probably want `AllocPool` and `make_alloc_pool` to be parametrized instead, and only specialize for the top-level `Alloc` and for `current_alloc_pool`.

src/mem/threadalloc.h
```diff
@@ -167,10 +182,15 @@ namespace snmalloc
 #  ifdef USE_MALLOC
       return get_reference();
 #  else
-      auto alloc = get_reference();
+      auto*& alloc = get_reference();
       if (unlikely(needs_initialisation(alloc)))
```
plietar, line 186: Adding an `&& !has_run` avoids a useless aquire/release in `init_thread_allocator`

src/mem/threadalloc.h
```diff
@@ -167,10 +182,15 @@ namespace snmalloc
...
       {
-        alloc = reinterpret_cast<Alloc*>(init_thread_allocator());
+        std::function<void*(void*)> f = [](void*) { return nullptr; };
+        // Call `init_thread_allocator` to perform down call in case
+        // register_clean_up does more.
+        // During teardown for the destructor based ThreadAlloc this will set
+        // alloc to GlobalPlaceHolder;
+        init_thread_allocator(f);
```
plietar, line 193: "The returned Alloc* is guaranteed to be initialised" is not true during teardown right? I reckon we can drop that guarantee, I doubt this is a useful guarantee anyway.

src/pal/pal_posix.h
```diff
@@ -35,12 +38,25 @@ namespace snmalloc
...
 
+    static void print_stack_trace()
+    {
+      constexpr int SIZE = 1024;
+      void* buffer[SIZE];
+      auto nptrs = backtrace(buffer, SIZE);
+      fflush(stdout);
+      backtrace_symbols_fd(buffer, nptrs, STDOUT_FILENO);
+      puts("");
+      fflush(stdout);
```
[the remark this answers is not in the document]
mjp41: It gives a new line after the backtrace, which was useful for handling some weird cases to see where one stack trace ended and the next started.  The `fflush` is needed as the `abort` directly after can cause the stack trace not to be printed.

src/mem/threadalloc.h
mjp41: Stale comment from before I added the `std::function`.

src/ds/aba.h
plietar: Indeed, I confused the two names, sorry.

src/pal/pal_posix.h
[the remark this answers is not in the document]
plietar: Oh right, forgot puts() adds a newline

src/mem/globalalloc.h
mjp41: The refactor was much easier than I expected.

mjp41:
> LGTM, although I'm slightly concerned about the use of `std::function`. Creating them can result in an allocation, which could be bad if snmalloc is used as the default allocator.
> 
> We currently have few enough captures that it probably doesn't actually allocate (from what I can tell, GCC and clang can fit two words without allocation), but it feels pretty fragile.
> 
> Adding a template to InitThreadAllocator would help, but you can't have templated non type template arguments (ie. `template<template<typename Fn> void* (*InitThreadAllocator)(const Fn&)>`).
> 
> Instead we can wrap the two functions as static methods:
> 
> ```c++
> struct AllocInitializer {
>   static bool needs_initialization() { [...] }
>   template<typename Fn>
>   static auto with_initialized_allocator(Fn& fn) {
>     [...]
>     auto value = fn(allocator);
>     [...]
>     return value;
>   }
> };
> ```
> 
> Then have Alloc parametrized over the concrete initializer instead and use it like:
> 
> ```c++
> template<typename Initializer, [...]>
> struct Allocator {
>   void* small_alloc_first_alloc(size_t size) {
>     return Initializer::with_initialized_allocator([size](Allocator* allocator) {
>       return allocator->alloc(size);
>     });
>   }
> }
> ```
> 
> I think this is guaranteed to not have any allocation.

I really like this idea, I have not managed to get it to work with the current layering and header only nature AllocInitialiser needs to know about Alloc, and Alloc needs to know about AllocInitialiser. So that is a little problem.

I would like to not use std::function, and I think I can make this work.  I propose I complete this PR, as it is already getting quite big, and submit this idea as a separate PR later.

PR #164, opened 2020-04-09
Ensure TLS callbacks don't allocate

mjp41:


plietar:
I don't think we need to forego lambdas completely, only std::function.

Based on http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0792r3.html, I managed to get this working:
```c++
template<typename Fn>
struct function_ref;
template<typename R, typename... Args>
struct function_ref<R(Args...)>
{
     template<typename Fn>
     function_ref(Fn&& fn) {
        data_ = static_cast<void*>(&fn);
        fn_ = execute<Fn>;
    }
    R operator()(Args... args) const {
        return execute_(data_, args...);
    }

  private:
    void* data_;
    R (*fn_)(void*, Args...);

    template<typename Fn>
    static R execute(void* p, Args... args) {
        return (*static_cast<Fn*>(p))(args...);
    };
};
```

It's pretty much a drop-in replacement for std::function, and has pretty good codegen and no allocation: https://godbolt.org/z/62tvQp

mjp41:
@plietar, I am not sure.  We only have two callbacks, and the code is fairly compact now.  Can you submit a "rival" PR with your changes, so I can see how it would look?  Or add a commit to this one?

plietar:
> @plietar, I am not sure. We only have two callbacks, and the code is fairly compact now. Can you submit a "rival" PR with your changes, so I can see how it would look? Or add a commit to this one?

See #166.

mjp41:
Closing in favour of #166

Issue #168, opened 2020-04-11
Broken build on x86_64 GCC environment 


SchrodingerZhu:
broken in `gcc-9`, `mingw-w64 gcc 9` and `gcc-10`
see https://travis-ci.com/github/SchrodingerZhu/snmalloc-rs/jobs/318380181.
also reproduced locally on my archlinux.

SchrodingerZhu:
> Starting with GCC 9, -Warray-bounds detects cases when memset is called
starting on a member of a struct but the size to be cleared ends up
writing over further members.

mjp41:
This appears to have been caused by #166. I guess the call `function_ref` is confusing the static analysis. But it is quite strange as the `SNMALLOC_ASSUME` had previously fixed this.  I am going to disable this warning as it causes more trouble than it is worth.

PR #170, opened 2020-04-12
Addressing #168

mjp41:


mjp41:
Addresses #168, @SchrodingerZhu does this fix it for you?

SchrodingerZhu:
I confirm the fix.

src/ds/defines.h
```diff
@@ -62,10 +62,15 @@ namespace snmalloc
...
 #    define SNMALLOC_ASSUME(x) \
-      do \
-      { \
-      } while (0)
+      if (!(x)) \
+        __builtin_unreachable();
+#  else
+do
+{
+} while (0)
```
plietar, line 74: This is missing the #define and backlashes.

Does this path even get used? `__GNUC__ˋ includes both GCC and clang.

mjp41: Should have realised, was very confused by clangformat on that one.

mjp41: Not used by anything yet.

mjp41:
So we are looking up a value in a table that is guaranteed to be less than `NUM_SMALL_SIZECLASSES` I guess it cannot determine that this is the case.  It would be a fairly complex static analysis to get this right.  I seem to remember @davidchisnall said this analysis is conservative, it can loose precision and will report errors based on that.  The `SNMALLOC_ASSUME` stuff was helping but I guess that is not the case with more recent GCC.

Thought, it it strange that your PR caused the issue.  I wonder if the code pointer dispatch in `function_ref` is causing a lose of precision in the analysis, which is strange as the table is `constexpr`.

plietar:
I did a bit more digging, and managed to get the following minimal reproduction:
```c++
#include <cstddef>

bool maybe1();
bool maybe2();

class C
{
  void* data[42];
  int extra;

public:
  void* foo(size_t index)
  {
    if (index < 42)
    {
      return nullptr;
    }
    if (maybe1())
    {
      return bar(index);
    }
    return nullptr;
  }

  void* bar(size_t index)
  {
    if (maybe2()) {
      return data[index];
    }
    return nullptr;
  }
};

int call_foo(C* obj, int index)
{
  obj->foo(index);
  return 0;
}
```

From what I was able to find online, the GCC warning is defined as "assuming the array access is reached, it is always OOB", rather than "this array access is reachable and is out-of-bounds".

In the code above, I think GCC inlines foo and bar into call_foo *before* doing the OOB analysis, and figures out that if the `data[index]` line is reached (ie. maybe1 and maybe2 are true) then it must be OOB. It makes sense in the inlined version, but in the original version `bar` could have been called directly, with a valid index. That's a pretty weird behaviour and probably a bug in GCC.

> Though, it it strange that your PR caused the issue. I wonder if the code pointer dispatch in function_ref is causing a lose of precision in the analysis, which is strange as the table is constexpr.

I think it's the other way around, `function_ref` gets inlined better than `std::function`, so GCC finds the OOB condition (from the large alloc code path recursing into the small alloc one) whereas it didn't before.

mjp41:
So do we have an issue on that path?

mjp41:
So I had a bit of a play based on your observation.  If we call back with `small_first_alloc` calling back in to `small_alloc_inner`, and `medium_alloc` into `medium_alloc` and `large_alloc` into `large_alloc` then the warning goes away.

In doing this, I also released the `zero_mem` flag is not correctly set for the call-back. So I'll add a test for this, and adjust the code, and reenable the warning.

PR #171, opened 2020-04-13
add android support

SchrodingerZhu:


SchrodingerZhu:
You can test the compiled binaries with the methods mentioned in https://stackoverflow.com/questions/9868309/how-to-compile-c-into-an-executable-binary-file-and-run-it-in-android-from-andro.

works fine on my side.

SchrodingerZhu:
@mjp41 is it better to implement a specified `stacktrace` using `libunwind` here rather than simply removing it?

davidchisnall:
Does Android have libexecinfo?  It would be nice to make the backtrace support conditional on the backtrace symbols existing, with a CMake check.

mjp41:
@davidchisnall are you proposing a check like we do in Verona for filesystem support:

https://github.com/microsoft/verona/blob/f193b1ce246333d6826e4538bebb29e117996ce6/src/compiler/CMakeLists.txt#L3-L11

SchrodingerZhu:
I do not know whether this would work or not, will give it a try: https://cmake.org/cmake/help/v3.0/module/CheckLibraryExists.html

CMakeLists.txt
```diff
@@ -2,21 +2,23 @@ cmake_minimum_required(VERSION 3.8)
...
-option(SNMALLOC_ANDROID "Build Target Android" OFF)
+option(SNMALLOC_ANDROID "Set cross build target to Android" OFF)
 set(CACHE_FRIENDLY_OFFSET OFF CACHE STRING "Base offset to place linked-list nodes.")
 
 if (SNMALLOC_ANDROID)
-set(CMAKE_SYSTEM_PROCESSOR "arm")
+  set(CMAKE_SYSTEM_PROCESSOR "arm")
 endif()
 
+check_include_file(execinfo.h HAS_EXECINFO_H)
```
SchrodingerZhu, line 20: This macro can help us check the header file directly.

mjp41:
Could you add some documentation about how to use this feature?

CMakeLists.txt
```diff
@@ -74,6 +74,11 @@ target_include_directories(snmalloc_lib INTERFACE src/)
 
 if(NOT MSVC)
   find_package(Threads REQUIRED COMPONENTS snmalloc_lib)
+  if (SNMALLOC_ARM_ANDROID)
+    if (NOT "${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
+      target_link_libraries(snmalloc_lib INTERFACE log c++_static c++abi m)
```
SchrodingerZhu, line 79: with this I can build the library and run it correctly for armv7, but the size is larger.

CMakeLists.txt
```diff
@@ -10,10 +10,10 @@ option(EXPOSE_EXTERNAL_PAGEMAP "Expose the global pagemap" OFF)
 option(EXPOSE_EXTERNAL_RESERVE "Expose an interface to reserve memory using the default memory provider" OFF)
 option(SNMALLOC_RUST_SUPPORT "Build static library for rust" OFF)
 option(SNMALLOC_QEMU_WORKAROUND "Disable using madvise(DONT_NEED) to zero memory on Linux" Off)
-option(SNMALLOC_ANDROID "Set cross build target to Android" OFF)
+option(SNMALLOC_ARM_ANDROID "Set cross build target to ARM Android" OFF)
```
SchrodingerZhu, line 13: The reason why I change this is that android does not always mean `arm`

CMakeLists.txt
SchrodingerZhu: dynamic linking failed to resolve `__emutls_get_address`

CMakeLists.txt
davidchisnall: Why is this needed separately from setting `CMAKE_SYSTEM_PROCESSOR`?  Normally, when you cross-compile with CMake, you provide a CMake toolchain file that sets things like this.  A quick search suggests that the Android NDK comes with one of these (apparently it's in `<NDK>/build/cmake/android.toolchain.cmake`).  Does that not set the correct things?

CMakeLists.txt
davidchisnall: Do we need this as a configure-time check?  C++17 standardised `__has_include()` and we depend on a C++17 compiler.

src/override/malloc.cc
```diff
@@ -43,10 +43,18 @@ extern "C"
...
+#ifndef __ANDROID__
   SNMALLOC_EXPORT size_t SNMALLOC_NAME_MANGLE(malloc_usable_size)(void* ptr)
   {
     return Alloc::alloc_size(ptr);
   }
+#else
+  SNMALLOC_EXPORT
+  size_t SNMALLOC_NAME_MANGLE(malloc_usable_size)(const void* ptr)
+  {
+    return Alloc::alloc_size(const_cast<void*>(ptr));
```
davidchisnall, line 55: `alloc_size` shouldn't need to modify anything through the pointer.  I'd prefer to pass a `const void*` here and unify both code paths.

src/pal/pal_posix.h
```diff
@@ -2,8 +2,9 @@
 
 #include "../ds/address.h"
 #include "../mem/allocconfig.h"
-
-#include <execinfo.h>
+#ifdef SNMALLOC_HAS_EXECINFO_H
```
davidchisnall, line 5:
```suggestion
#if __has_include(<execinfo.h>)
#define SNMALLOC_HAS_BACKTRACE 1
```

src/pal/pal_posix.h
```diff
@@ -38,6 +39,7 @@ namespace snmalloc
      */
     static constexpr uint64_t pal_features = LazyCommit;
 
+#ifdef SNMALLOC_HAS_EXECINFO_H
```
davidchisnall, line 42:
```suggestion
#ifdef SNMALLOC_HAS_BACKTRACE
```

src/pal/pal_posix.h
```diff
@@ -48,14 +50,17 @@ namespace snmalloc
...
     }
+#endif
 
     /**
      * Report a fatal error an exit.
      */
     static void error(const char* const str) noexcept
     {
       puts(str);
+#ifdef SNMALLOC_HAS_EXECINFO_H
```
davidchisnall, line 61: Please remove these guards and move the guards in the callee inside the function.  The compiler will then elide this call.

davidchisnall:
Thanks, I've added a few comments inline, but they're all minor cleanups.

CMakeLists.txt
SchrodingerZhu: You are right. `CMAKE_TOOLCHAIN_FILE` is definitely a good choice, will look into it later.

I set this just because passing `CMAKE_SYSTEM_PROCESSOR` via command line is not working. 

It should be set by `android.toolchain.cmake`.

SchrodingerZhu:
using toolchain file however, may make it hard for other language like rust to automatically compile this library. see https://github.com/alexcrichton/cmake-rs/issues/80.

I think I will adopt a dirty hack to let the user pass a `ANDROID_NDK` environment variable in my rust crate. Do you have any thought on how to improve this?

CMakeLists.txt
SchrodingerZhu: using toolchain file for `androidebi` will solve the linking issue too. the linking result of `emutls` after changing to the new way:
```
0002d7e0 t __emutls_get_address
0002d8b8 t __emutls_register_common
0003eff4 r __emutls_t._ZL16g_current_locale
0003e15c r __emutls_t._ZN10__cxxabiv112_GLOBAL__N_15dtorsE
00040dc0 V __emutls_t._ZZN8snmalloc17ThreadAllocCommon13get_referenceEvE5alloc
0002d7c8 t __emutls_unregister_key
00043020 V __emutls_v._ZGVZN8snmalloc27ThreadAllocThreadDestructor16register_cleanupEvE6tidier
00043080 d __emutls_v._ZL16g_current_locale
00043050 d __emutls_v._ZN10__cxxabiv112_GLOBAL__N_111dtors_aliveE
00043060 d __emutls_v._ZN10__cxxabiv112_GLOBAL__N_15dtorsE
00043030 V __emutls_v._ZN8snmalloc17ThreadAllocCommon18destructor_has_runE
00043040 V __emutls_v._ZN8snmalloc19operation_in_flightE
00043000 V __emutls_v._ZZN8snmalloc17ThreadAllocCommon13get_referenceEvE5alloc
00043010 V __emutls_v._ZZN8snmalloc27ThreadAllocThreadDestructor16register_cleanupEvE6tidier
0002d6ec t emutls_alloc
0002d6a8 t emutls_destroy
0002d784 t emutls_init
00046dd8 b emutls_key
00046de0 b emutls_key_created
00046ddc b emutls_mutex
00046de8 b emutls_size
```

src/override/malloc.cc
SchrodingerZhu: Do you mean changing the inner paths within snmalloc?

davidchisnall: Yes, exactly.  I don't think `Alloc::alloc_size` should need to take a non-`const` pointer, so let's make that one `const`.

davidchisnall:
Does Rust depend on the NDK at all for Android cross-compilation?  I don't have much experience there, but it looks as if the NDK has a fairly standard CMake cross-compile process and is gradually moving in the direction of better alignment with the normal CMake mechanisms where they differ.  

I'd assume that if you're cross-compiling any CMake project you'd use a toolchain file (every other approach I've tried in the past has resulted in far more pain in the long term), but the cmake-rs issue that you link to seems to imply that they don't want to do that.

CMakeLists.txt
```diff
@@ -210,7 +209,7 @@ if(NOT DEFINED SNMALLOC_ONLY_HEADER_LIBRARY)
     add_shim(snmallocshim-1mib SHARED ${SHARED_FILES})
     target_compile_definitions(snmallocshim-1mib PRIVATE IS_ADDRESS_SPACE_CONSTRAINED)
   endif()
-
+
```
davidchisnall, line 212: The only changes in this file are now whitespace.  Can you remove them to reduce churn?

plietar:
> `i686` target is not supported because of the design of `snmalloc`.

Is there a reason for this? We have 32-bit windows on the CI, and I can get 32-bit linux building with a very small change (just a spurious `#error` in aba.h, I'll make a PR later).

SchrodingerZhu:
> > `i686` target is not supported because of the design of `snmalloc`.
> 
> Is there a reason for this? We have 32-bit windows on the CI, and I can get 32-bit linux building with a very small change (just a spurious `#error` in aba.h, I'll make a PR later).

@plietar It was the previous discussion that make me write something like this. Perhaps my words were not proper here. Sorry about that.
https://github.com/microsoft/snmalloc/pull/137#issuecomment-596467464

CMakeLists.txt
SchrodingerZhu: I have just rebased away the changes on `CMakeLists.txt`

src/ds/address.h
```diff
@@ -70,7 +70,7 @@ namespace snmalloc
    * power of two.
    */
   template<size_t alignment, typename T = void>
-  SNMALLOC_FAST_PATH T* pointer_align_down(void* p)
+  SNMALLOC_FAST_PATH T* pointer_align_down(const void* p)
```
SchrodingerZhu, line 73: Notice that this result in an inconsistent qualifier with other alignment cast operations.

src/override/malloc.cc
```diff
@@ -43,7 +43,8 @@ extern "C"
     return ThreadAlloc::get_noncachable()->alloc<ZeroMem::YesZero>(sz);
   }
 
-  SNMALLOC_EXPORT size_t SNMALLOC_NAME_MANGLE(malloc_usable_size)(void* ptr)
+  SNMALLOC_EXPORT
+    size_t SNMALLOC_NAME_MANGLE(malloc_usable_size)(const void* ptr)
```
davidchisnall, line 47: I think you still need the non-const version for other platforms that declare this in a header that we include here.

SchrodingerZhu: Currently, it seems to work for other platforms.

I guess the case seems to be this:
- Android somehow includes the `malloc.h` via the headers.
- Other platforms does not do this.
- Compiler discards CV-qualifiers so linking stage is not affected for most platforms.
- However, the definitions conflicts on android because the compiler can find the real definitions.

So, do you still suggest me to provide two different versions? if so, I will work on it ASAP.

SchrodingerZhu: @davidchisnall How about something like:

```cmake
include(CheckCSourceCompiles)
CHECK_C_SOURCE_COMPILES("
#include <malloc.h>
size_t malloc_usable_size(const void* ptr) { return 0; }
int main() { return 0; }
" CONST_MALLOC_USABLE_SIZE)
```

SchrodingerZhu:
Rebased again to master:
- remove `i686` marks in documentation
- use cmake to detect the const qualifier

mjp41:
@SchrodingerZhu thanks for this.

PR #173, opened 2020-04-15
Don't require 16-byte CAS on x86.

plietar:
We only need to CAS 2 pointers, which is always possible on x86.

This uses the `plietar/snmalloc-build_linux_x86:latest` docker image, obviously that should be changed to something under snmallocciteam.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -108,7 +108,7 @@ jobs:
         CMakeArgs: ''
         Image: plietar/snmalloc-build_linux_x86:latest
 
-  container: $(Image)
+  container: $[ variables['Image'] ]
   steps:
   - script: |
       ci/scripts/build.sh
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -9,80 +9,106 @@ jobs:
   displayName: Linux
   pool:
     vmImage: 'ubuntu-18.04'
-  container: snmallocciteam/build_linux_x64:latest
 
   strategy:
     matrix:
-      Clang-7 Debug:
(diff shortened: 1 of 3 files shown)


azure-pipelines.yml
```diff
@@ -9,80 +9,106 @@ jobs:
...
         CMakeArgs: '-DCACHE_FRIENDLY_OFFSET=64'
+        Image: snmallocciteam/build_linux_x64:latest
+
+      32-bit Clang-9 Debug:
+        CC: clang-9
+        CXX: clang++-9
+        BuildType: Debug
+        SelfHost: false
+        CMakeArgs: ''
+        Image: plietar/snmalloc-build_linux_x86:latest
```
achamayou, line 101: I'll add you to snmallocciteam, so you can update/replace images there.

src/ds/aba.h
```diff
@@ -99,9 +99,10 @@ namespace snmalloc
           (__int64)value,
           (__int64*)&old);
 #  else
-#    if defined(__GNUC__) && !defined(__GCC_HAVE_SYNC_COMPARE_AND_SWAP_16)
+#    if defined(__GNUC__) && defined(SNMALLOC_VA_BITS_64) && !defined(__GCC_HAVE_SYNC_COMPARE_AND_SWAP_16)
```
achamayou, line 102: I'll try not to forget to that out in Open Enclave.

azure-pipelines.yml
achamayou: Done now.

plietar: Thanks

PR #174, opened 2020-04-25
Disable compare-exchange/mxc16  unsupported on android as warning are…

devnexen:
… treated as errors.


SchrodingerZhu:
@devnexen 
what is your `android-ndk` version? It is interesting that it seems that I cannot reproduce the problem with
```
cmake .. -DCMAKE_BUILD_TYPE=Release -DCMAKE_TOOLCHAIN_FILE=/opt/android-ndk/build/cmake/android.toolchain.cmake -DANDROID_ABI=x86_64
```

SchrodingerZhu:
FYI, I didn't see any problem in my compile commands.

[compile-commands.zip](https://github.com/microsoft/snmalloc/files/4533481/compile-commands.zip)

devnexen:
The very last (21.x) ... in fact there is no so much pb but when warnings as treated as in errors in y case that break the build for unsupported flag.

SchrodingerZhu:
Could u please take a look at my dumped commands and compare with yours?
As it is written in the `CMakeLists.txt`, this flag should be clear for those `arm` targets.
As for `x86_64`, those commands could run without problem on my side. A little bit strange.

My version is also `r21.b-1` which is carrying a clang `9.0.8`.

devnexen:
Ah right I see in my side what went wrong sorry for the noise.

mjp41:
This issue makes me strangely happy: the first PR/Issue that didn't involve an MSFT person. :-)

PR #184, opened 2020-05-12
test/contention: release size store first

nwf:
It is important, in test_tasks_f, that the store of the size to the
allocated block be made visible to other processors before the store of
the pointer itself.  Otherwise, other cores are justified in reading
junk.

This manifests on PowerPC as tripping the "Deallocating with incorrect
size supplied" assertion in alloc.h:/check_size because the value read
from the allocated block may not be a size but rather an internal queue
pointer, which is implausibly large, as sizes go.

### DIFF
--- a/src/test/perf/contention/contention.cc
+++ b/src/test/perf/contention/contention.cc
@@ -86,7 +86,7 @@ void test_tasks_f(size_t id)
 
     *res = size;
     size_t* out =
-      contention[n % swapsize].exchange(res, std::memory_order_relaxed);
+      contention[n % swapsize].exchange(res, std::memory_order_acq_rel);
 
     if (out != nullptr)
     {


src/test/perf/contention/contention.cc
```diff
@@ -86,7 +86,7 @@ void test_tasks_f(size_t id)
 
     *res = size;
     size_t* out =
-      contention[n % swapsize].exchange(res, std::memory_order_relaxed);
+      contention[n % swapsize].exchange(res, std::memory_order_release);
```
mjp41, line 89: I think you are right, but shouldn't it be `std::memory_order_acq_rel` as this performs as a read as well, that needs to induce the happens-before edge?

nwf: Yea, I think you're right.  I think I thought the data dependencies in this thread would be sufficient, but they're not.

PR #185, opened 2020-05-12
Move OS_PAGE_SIZE to PAL, add Linux PowerPC w/ 64KiB pages

nwf:
At least the latter half is necessary for Debian/PowerPC64 support.

The first commit here should have no functional changes and just lay the groundwork.  Admittedly, tho', it's selected out of a much larger mass of changes, so perhaps I've missed something.  Let's see what CI says.

Should we add a check to `pal_posix` at startup that `OS_PAGE_SIZE` is divisible by `sysconf(_SC_PAGESIZE)`?

### DIFF
--- a/src/aal/aal_powerpc.h
+++ b/src/aal/aal_powerpc.h
@@ -1,0 +1,37 @@
+#pragma once
+
+#if defined(__powerpc64__)
+#  define SNMALLOC_VA_BITS_64
+#else
+#  define SNMALLOC_VA_BITS_32
+#endif
+
+namespace snmalloc
+{
+  /**
+   * ARM-specific architecture abstraction layer.
+   */
+  class AAL_PowerPC
+  {
+  public:
+    /**
+     * Bitmap of AalFeature flags
+     */
+    static constexpr uint64_t aal_features = IntegerPointers;
+
+    static constexpr enum AalName aal_name = PowerPC;
+
+    static constexpr size_t smallest_page_size = 0x1000;
+
+    /**
+     * On pipelined processors, notify the core that we are in a spin loop and
+     * that speculative execution past this point may not be a performance gain.
+     */
+    static inline void pause()
+    {
+      __asm__ volatile("or 27,27,27"); // "yield"
+    }
+  };
+
+  using AAL_Arch = AAL_PowerPC;
+} // namespace snmalloc
--- a/src/pal/pal_apple.h
+++ b/src/pal/pal_apple.h
@@ -32,9 +32,9 @@ namespace snmalloc
     template<bool page_aligned = false>
     void zero(void* p, size_t size)
     {
-      if (page_aligned || is_aligned_block<OS_PAGE_SIZE>(p, size))
+      if (page_aligned || is_aligned_block<page_size>(p, size))
       {
-        SNMALLOC_ASSERT(is_aligned_block<OS_PAGE_SIZE>(p, size));
+        SNMALLOC_ASSERT(is_aligned_block<page_size>(p, size));
         void* r = mmap(
           p,
           size,
--- a/src/pal/pal_bsd.h
+++ b/src/pal/pal_bsd.h
@@ -33,7 +33,7 @@ namespace snmalloc
      */
     void notify_not_using(void* p, size_t size) noexcept
     {
-      SNMALLOC_ASSERT(is_aligned_block<OS_PAGE_SIZE>(p, size));
+      SNMALLOC_ASSERT(is_aligned_block<OS::page_size>(p, size));
       madvise(p, size, MADV_FREE);
     }
   };
--- a/src/pal/pal_linux.h
+++ b/src/pal/pal_linux.h
@@ -25,7 +25,8 @@ namespace snmalloc
      */
     static constexpr uint64_t pal_features = PALPOSIX::pal_features;
 
-    static constexpr size_t page_size = 0x1000;
+    static constexpr size_t page_size =
+      Aal::aal_name == PowerPC ? 0x10000 : 0x1000;
 
     /**
      * OS specific function for zeroing memory.
--- a/src/pal/pal_linux.h
+++ b/src/pal/pal_linux.h
@@ -2,7 +2,6 @@
 
 #if defined(__linux__)
 #  include "../ds/bits.h"
-#  include "../mem/allocconfig.h"
 #  include "pal_posix.h"
 
 #  include <string.h>
@@ -26,6 +25,8 @@ namespace snmalloc
      */
     static constexpr uint64_t pal_features = PALPOSIX::pal_features;
 
+    static constexpr size_t page_size = 0x1000;
+
     /**
      * OS specific function for zeroing memory.
      *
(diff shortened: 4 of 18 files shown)


src/pal/pal_apple.h
```diff
@@ -23,6 +23,8 @@ namespace snmalloc
      */
     static constexpr uint64_t pal_features = PALBSD::pal_features;
 
+    static constexpr size_t OS_PAGE_SIZE = 0x1000;
```
davidchisnall, line 26: `PalApple` inherits from `PALBSD`.  Why do you need this here, but not in other *BSD PALs.

src/pal/pal_linux.h
```diff
@@ -26,6 +25,16 @@ namespace snmalloc
      */
     static constexpr uint64_t pal_features = PALPOSIX::pal_features;
 
+#  ifdef SNMALLOC_OS_PAGE_SIZE
+    /*
+     * Allow build-time customization of OS_PAGE_SIZE since Linux is apparently
+     * a little less opinionated across its various ports than many other OSes.
+     */
+    static constexpr size_t OS_PAGE_SIZE = SNMALLOC_OS_PAGE_SIZE;
```
davidchisnall, line 33: Do we want to allow a global override and assert that it's at least equal to the OS's limit?  We can always support a page size that's greater than the OS wants, just not one that's smaller.

I also wonder if we want to have a page size in the AAL and the PAL and take the max of the two.

src/pal/pal_open_enclave.h
```diff
@@ -19,6 +19,9 @@ namespace snmalloc
      * PAL supports.
      */
     static constexpr uint64_t pal_features = 0;
+
+    static constexpr size_t OS_PAGE_SIZE = 0x1000;
```
davidchisnall, line 23: OE is somewhat interesting, because its back end doesn't actually care about pages.  I wonder what happens if we use a smaller than 4KiB page here...

src/pal/pal_apple.h
nwf: Mm, I probably don't and added this one first to appease CI.  Let's see what happens without.

src/pal/pal_linux.h
nwf: Do any architectures out in the wild not *support* 4K pages, at least in principle?  AFAIK ppc64 does, it just happens that Debian/ppc64le builds their kernel with 64K pages.

src/pal/pal_open_enclave.h
nwf: Well, experimentally dropping it to `0x800` and re-running the `two_alloc_types` suggests that things don't immediately fall over (that is, the tests pass, but I don't think that one test has phenomenal coverage of OE).

mjp41:
So I think some of this constraint was to ensure that the Medium allocations could be turned off independently, and that the slabs for small allocations could be turned off.

I removed most of this code, so I think probably as long as SUPERSLAB_SIZE is a multiple of OS_PAGE_SIZE, then everything should work.  I think we are only turning things on and off at the superslab granularity.

src/aal/aal_powerpc.h
```diff
@@ -0,0 +1,33 @@
...
+     */
+    static constexpr uint64_t aal_features = IntegerPointers;
+
+    /**
+     * On pipelined processors, notify the core that we are in a spin loop and
+     * that speculative execution past this point may not be a performance gain.
+     */
+    static inline void pause()
+    {
+      __asm__ volatile("or 27,27,27"); // "yield"
```
davidchisnall, line 28: Thank you for reminding me why I hate PowerPC assembly so much.

src/pal/pal_bsd.h
```diff
@@ -24,6 +24,8 @@ namespace snmalloc
      */
     static constexpr uint64_t pal_features = PALPOSIX<OS>::pal_features;
 
+    static constexpr size_t OS_PAGE_SIZE = 0x1000;
```
davidchisnall, line 27: It's probably sensible to put this in pal_posix (if it's useful in a superclass at all).

src/pal/pal_linux.h
davidchisnall: I think SPARC supports only 8KiB pages (though they might have fixed that in later SPARC)?  That said, I don't particularly want to support SPARC.  

Why is the external override for Linux?  I think that I'd prefer to structure this as:

 - Add an `enum` for the architecture.
 - Make the AAL expose a constexpr field with the correct enum value.
 - Make this a `std::conditional` keyed off that field.
 - Use `SNMALLOC_OS_PAGE_SIZE` in the global constant that we currently initialise with the version from the PAL.
 - Add a static assert that this global is a multiple of `PAL::OS_PAGE_SIZE`

Does that make sense?

src/aal/aal_powerpc.h
nwf: I'm not sure why there isn't a pseudoinstruction hint in LLVM, but so it goes.  This is no worse than any other "idempotent operator's encoding becomes hint instead" that everybody and their brother loves, is it?

src/pal/pal_linux.h
nwf: Ah, indeed, UltraSPARC sayeth:

> The MMUs may support up to eight page sizes: 8 KBytes, 64 KBytes, 512 KBytes, 4 MBytes, 32 MBytes, 256 MBytes, 2 GBytes, and 16 GBytes.

When you say "Add an `enum` for the architecture" do you mean something like
```
enum aal_page_size {
  AAL_PAGE_SIZE_4K,
  // ...
  AAL_PAGE_SIZE_64K
};
```
or something else?

davidchisnall: I meant replace / augment the `PLATFORM_IS_X86` and similar with something like:
```c
enum Architecture
{
    X86,
    ARM,
    PowerPC
};
```

And then be able to write something like this, rather than relying on `#ifdef`s:

```c++
static constexpr size_t OS_PAGE_SIZE = AAL::Arch == Architecture::PowerPC ? 0x10000 : 0x1000;
```

Though, actually, it would make me even happier to have these in one of the helper headers:

```c++
constexpr unsigned long long int operator "" _KiB(unsigned long long int v)
{
        return v * 1024;
}

constexpr unsigned long long int operator "" _MiB(unsigned long long int v)
{
        return v * 1024 * 1024;
}
```
And then to write it as:

```c++
static constexpr size_t OS_PAGE_SIZE = AAL::Arch == Architecture::PowerPC ? 64_KiB : 4_KiB;
```

But that's not necessary for this PR unless you feel like it.

src/aal/aal_powerpc.h
davidchisnall: It's more the complete lack of syntactic hint to let you know that 27 is a register number or a literal that I dislike.  PowerPC assembly seems designed to discourage anyone from ever reading any, let alone writing some.

nwf:
Unfortunately, untangling the dependencies here is harder than I thought it would be.  We can't just have concluded the correct value of `OS_PAGE_SIZE` by the end of processing `src/pal/pal.h` because `pal_posix.h`, at least, wants to know the answer.  I've pushed what I have so far to https://github.com/nwf/snmalloc/tree/pal-page-take2, but it doesn't compile (so I won't waste CI time on it by pushing here).

davidchisnall:
I see.  I wonder if we want to just not bother with the ability to override the PAL then (you can always subclass a PAL and provide your own versions if you want to).  Then, the POSIX PAL can just pull the value from its template parameter.

nwf:
The CI failure seems to have been a timeout of `MacOS Release`.  Can someone prod the pipeline to try that one again?

nwf:
Rebased to fix conflicts.

PR #187, opened 2020-05-15
Use cmake to find backtrace()

nwf:
On FreeBSD, this notably requires the use of -lexecinfo, as backtrace()
is not available in -lc.  Rather than testing in C, test in cmake.


mjp41:
@schrodingerzhu is this okay with your android prototyping.

SchrodingerZhu:
tested. it works fine for android.

PR #188, opened 2020-05-15
Use pointers, not addresses, in more places

nwf:
This should produce equivalent code everywhere and more obviously preserves provenance where desired.  Ultimately, eliminate `pointer_cast` entirely since it won't make sense to cast an address back to a pointer on CHERI platforms.

### DIFF
--- a/src/ds/address.h
+++ b/src/ds/address.h
@@ -43,16 +43,6 @@ namespace snmalloc
   }
 
   /**
-   * Cast from an address back to a pointer of the specified type.  All uses of
-   * this will eventually need auditing for CHERI compatibility.
-   */
-  template<typename T>
-  inline T* pointer_cast(address_t address)
-  {
-    return reinterpret_cast<T*>(address);
-  }
-
-  /**
    * Test if a pointer is aligned to a given size, which must be a power of
    * two.
    */
@@ -81,7 +71,8 @@ namespace snmalloc
 #if __has_builtin(__builtin_align_down)
       return static_cast<T*>(__builtin_align_down(p, alignment));
 #else
-      return pointer_cast<T>(bits::align_down(address_cast(p), alignment));
+      return reinterpret_cast<T*>(
+        bits::align_down(reinterpret_cast<uintptr_t>(p), alignment));
 #endif
     }
   }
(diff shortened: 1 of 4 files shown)


src/ds/address.h
```diff
@@ -81,7 +71,10 @@ namespace snmalloc
 #if __has_builtin(__builtin_align_down)
       return static_cast<T*>(__builtin_align_down(p, alignment));
 #else
-      return pointer_cast<T>(bits::align_down(address_cast(p), alignment));
+      static_assert(
+        aal_supports<IntegerPointers>,
```
davidchisnall, line 75: I am not convinced by this. If into intptr_t exposes an address, not an offset, we can handle this case for CHERI too.

src/ds/address.h
```diff
@@ -119,7 +134,10 @@ namespace snmalloc
 #if __has_builtin(__builtin_align_up)
     return static_cast<T*>(__builtin_align_up(p, alignment));
 #else
-    return pointer_cast<T>(bits::align_up(address_cast(p), alignment));
+    static_assert(
+      aal_supports<IntegerPointers>,
```
davidchisnall, line 138: Since we have this assert in multiple places, it would be good to factor it out and then we can relax the constraint easily later.

src/ds/address.h
nwf: The intent was that `address_t` would become `vaddr_t`, and not `intptr_t` on CHERI, so that it truly was an address.

davidchisnall: Does `reinterpret_cast<T*>(bits::align_down(reinterpret_cast<uintptr_t>(p), alignment));` work?  If so, the comment is misleading: we can realign via `bits` without integer pointers, we can't realign via bits if we explicitly cast to an address first.

nwf: That's fair; I'll drop the assert and change the casts.

src/mem/alloc.h
```diff
@@ -423,41 +423,35 @@ namespace snmalloc
...
       }
 
-      auto ss = address_cast(super);
+      auto ss = super;
 
       while (size > 64)
       {
         // This is a large alloc redirect.
-        ss = ss - (1ULL << (size - 64));
+        ss = pointer_offset_signed(ss, -(1 << (size - 64)));
```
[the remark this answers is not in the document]
nwf: Annoyingly, when I use `1LL`, my host compiler barfs:

> /cheri/source/mainline/snmalloc/src/mem/alloc.h:431:40: error: implicit conversion changes signedness: 'long long' to 'ptrdiff_t' (aka 'long') [-Werror,-Wsign-conversion]

which just seems wrong to me?  Anyway, I've worked around it by doing `static_cast<ptrdiff_t>(1)` which should make everyone happy on all platforms, yes?

PR #189, opened 2020-05-16
add static library

SchrodingerZhu:
this can be useful if others want to embed snmalloc into their programs. 

For example: https://github.com/antirez/redis/pull/7262


mjp41:
Could you make the CI build the static library on as many platforms as makes sense? Otherwise looks great.  Does this make sense as an option for Windows even?

SchrodingerZhu:
mimalloc compiles both dynamic and static library on all platforms by default. Maybe I can just set this option to `On` by default?

davidchisnall:
I am curious as to the intended use case for this?  If you are deeply embedding snmalloc, it is a header-only project that you can just include without needing a standard library.  The shim library is built to be `LD_PRELOAD`ed, but you can't do that with a static library.  The only use case I can see for this is if you are statically linking and replacing the system `malloc`, but that seems a somewhat dangerous thing to encourage.

SchrodingerZhu:
@davidchisnall unlike jemalloc and mimalloc, snmalloc do not have a header file define the drop-in replacement like: `sn_malloc`, `sn_free`, maybe providing that will be a better solution?

SchrodingerZhu:
I think we can use `SNMALLOC_NAME_MANGLE` to provide a similar approach like `jemalloc`, but still, this requires building static library.

davidchisnall:
In other projects where I am providing a new malloc API, I define `SNMALLOC_NAME_MANGLE` and then `#include` the override files in a .cc file in my project.  That said, I am also using a custom PAL for these projects, so I would not be able to use a static library.  I don't object to providing one, I'm just curious as to how it would be used.

SchrodingerZhu:
I see, may 
```c++
 c++ -Isrc -c src/override/malloc.cc -mcx16 -std=c++17 -D"SNMALLOC_NAME_MANGLE(x)=sn_##x" -o override.o
```
is good enough to use.

mjp41:
So perhaps have a Cmake option:
   `SNMALLOC_STATIC_LIBRARY_PREFIX`
The default is "sn_", and always build the static library on all platforms.

SchrodingerZhu:
BTW, is it better to output `libsnmallocshim.a` on unix platforms, and `libsnmallocshim-static.a` on Win32 platforms?

see https://github.com/microsoft/mimalloc/blob/master/CMakeLists.txt#L246

mjp41:
I am happy with everything being called `-static`.

CMakeLists.txt
```diff
@@ -230,6 +231,11 @@ if(NOT DEFINED SNMALLOC_ONLY_HEADER_LIBRARY)
 
   if(NOT WIN32)
     set(SHARED_FILES src/override/new.cc src/override/malloc.cc)
+    if (SNMALLOC_STATIC_LIBRARY)
+      add_shim(snmallocshim-static STATIC ${SHARED_FILES})
```
mjp41, line 235: Should the static lib be compiling `new.cc`  this is overriding the `new` and `delete` operations globablly without namespace?

CMakeLists.txt
```diff
@@ -230,6 +231,11 @@ if(NOT DEFINED SNMALLOC_ONLY_HEADER_LIBRARY)
 
   if(NOT WIN32)
     set(SHARED_FILES src/override/new.cc src/override/malloc.cc)
+    if (SNMALLOC_STATIC_LIBRARY)
```
mjp41, line 234: Can you lift the static lib outside the `NOT WIN32`

PR #191, opened 2020-05-19
Allow the AAL to define address_t

nwf:
On CHERI, we will want `address_t` to be `vaddr_t`, an integer type big enough to store an address but that cannot carry provenance (i.e., is too small to hold a capability).

### DIFF
--- a/src/aal/aal.h
+++ b/src/aal/aal.h
@@ -46,6 +46,31 @@ namespace snmalloc
   template<class Arch>
   struct AAL_Generic : Arch
   {
+    /*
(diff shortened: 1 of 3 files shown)


src/aal/aal.h
```diff
@@ -46,6 +46,16 @@ namespace snmalloc
   template<class Arch>
   struct AAL_Generic : Arch
   {
+    struct default_address_t
```
davidchisnall, line 49: Please add some comments here.  It took me a little while to figure out what this code is doing.

PR #195, opened 2020-05-21
Make up for the lack of intrinsic headers in Open Enclave

achamayou:
Fix the main remaining problem that prevents building oesnmalloc with -nostdinc. There's still a issue with the order in which headers are copied in the OE build but that's nothing to do with snmalloc.

Paging @anakrish because I can't work how to add him to Reviewers for some reason.

It is probably worth noting that this can be removed if/when 
a subset of https://github.com/openenclave/openenclave/issues/1766 is implemented in OE. That's been open for over a year though, so I would still like to have this change considered.

### DIFF
--- a/src/aal/aal_x86_sgx.h
+++ b/src/aal/aal_x86_sgx.h
@@ -3,8 +3,6 @@
 #ifdef _MSC_VER
 #  include <immintrin.h>
 #  include <intrin.h>
-#else
-#  include <emmintrin.h>
 #endif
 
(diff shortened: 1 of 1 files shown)


achamayou:
I'm not quite sure how this happened or why:

```
##[error]Bash exited with code '128'.
##[error]Bash wrote one or more lines to the standard error stream.
##[error]fatal: bad object 6b4c75a659ce6f15ef32a2ec7661af57aceb3664
```

anakrish:
Thanks for doing this @achamayou

PR #198, opened 2020-05-23
WIP: Allow the AAL to set bounds on returned pointers and let the ChunkMap amplify them back

nwf:
<s>This builds atop #191 and #185; sorry, it's hard to have commutative patches. :(</s>

I would appreciate feedback about the AAL function `apply_bounds` and the use of the `ReturnPtr` type alias to ensure that the internal allocation functions have `apply_bounds` at their leaf returns.

<s>This is, of course, only half the story, and the easy half at that.  The `dealloc` functions must re-derive the internal pointers once the PAL is doing anything other than passing through `apply_bounds`.</s>

### DIFF
--- a/src/mem/alloc.h
+++ b/src/mem/alloc.h
@@ -412,30 +412,38 @@ namespace snmalloc
       error("Unsupported");
       UNUSED(p);
 #else
-      uint8_t size = ChunkMap::get(address_cast(p));
+      ReturnPtr p = unsafe_return_ptr(p_);
+      uint8_t size = ChunkMap::get(address_cast(p.ptr));
 
-      Superslab* super = Superslab::get(p);
       if (size == CMSuperslab)
       {
-        Slab* slab = Metaslab::get_slab(p);
+        Slab* slab = Metaslab::get_slab(chunkmap().amplify(p));
+        Superslab* super = Superslab::get(slab);
         Metaslab& meta = super->get_meta(slab);
 
         sizeclass_t sc = meta.sizeclass;
-        void* slab_end = pointer_offset(slab, SLAB_SIZE);
+        ReturnPtr slab_end =
+          Aal::reprovenance(p, pointer_offset(slab, SLAB_SIZE));
 
         return external_pointer<location>(p, sc, slab_end);
       }
       if (size == CMMediumslab)
       {
-        Mediumslab* slab = Mediumslab::get(p);
+        Mediumslab* slab = Mediumslab::get(chunkmap().amplify(p));
 
         sizeclass_t sc = slab->get_sizeclass();
-        void* slab_end = pointer_offset(slab, SUPERSLAB_SIZE);
+        ReturnPtr slab_end =
+          Aal::reprovenance(p, pointer_offset(slab, SUPERSLAB_SIZE));
 
         return external_pointer<location>(p, sc, slab_end);
       }
 
-      auto ss = super;
+      /*
+       * XXX: On CHERI, amplify() will give us a pointer with its
+       * base pointing right at where we end up after this log walk.
+       * We should use that.
+       */
+      Superslab* ss = Superslab::get(chunkmap().amplify(p));
 
       while (size > 64)
       {
@@ -842,18 +850,19 @@ namespace snmalloc
 
     template<Boundary location>
     static void*
-    external_pointer(void* p, sizeclass_t sizeclass, void* end_point)
+    external_pointer(ReturnPtr p, sizeclass_t sizeclass, ReturnPtr end_point)
     {
       size_t rsize = sizeclass_to_size(sizeclass);
 
+      void* epp = end_point.ptr;
       void* end_point_correction = location == End ?
-        pointer_offset_signed(end_point, -1) :
+        pointer_offset_signed(epp, -1) :
         (location == OnePastEnd ?
-           end_point :
-           pointer_offset_signed(end_point, -static_cast<ptrdiff_t>(rsize)));
+           epp :
+           pointer_offset_signed(epp, -static_cast<ptrdiff_t>(rsize)));
 
       size_t offset_from_end =
-        pointer_diff(p, pointer_offset_signed(end_point, -1));
+        pointer_diff(p.ptr, pointer_offset_signed(epp, -1));
 
       size_t end_to_end = round_by_sizeclass(rsize, offset_from_end);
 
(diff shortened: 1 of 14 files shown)


nwf:
Oh bah humbug.  Why is this incompatible with `declspec(allocator)`?  I guess I should allow the PAL to specify the return type.

src/mem/alloc.h
```diff
@@ -1069,7 +1079,7 @@ namespace snmalloc
         {
           large_allocator.memory_provider.zero(p, sizeclass_to_size(sizeclass));
         }
-        return p;
+        return Pal::apply_bounds(p, size);
```
mjp41, line 1082: Can't the freelists contain bounded pointers?  Or do you want a more precise bound than the allocation size, i.e. the request size? Personally, I would prefer the free lists to contain bounded pointers.  Then we can apply the bounds at deallocation based on the pagemap lookup, and that is the only place it is required.

mjp41:
> Oh bah humbug. Why is this incompatible with declspec(allocator)? I guess I should allow the PAL to specify the return type.

We could probably push the ALLOCATOR annotation to just the top-level alloc functions.  That would remove this error.  I am not sure having them on intermediates helps.

src/mem/alloc.h
nwf: I'd be content bounding things to the underlying allocation size, I think.  And I don't immediately see any reason the free lists couldn't store bounded pointers; I'll give that a shot.

davidchisnall:
I’m not sure why this belongs in the PAL and not the AAL. I would expect CHERI Windows, Linux, and FreeBSD to all have the same behaviour here, but MTE and CHERI FreeBSD to behave differently.

I wonder how either choice composes with things like recording ASan allocation state, if we wanted to support that. In either case, we’d probably want mixins for the memory safety strategy. I wonder if we want an abstraction layer for those, with helpers in the AAL for the mechanism.

nwf:
OK, I'll move BoundedPtr over to the AAL.  The CHERI AAL mixin is probably the natural place for this, yes; I haven't done that yet because I haven't yet re-done the "pointer re-derivation via pagemap" work, but it's next on the list.

nwf:
OK, after a long while of type-directed refactoring, I have a new prototype for consideration.  This is a little hack-and-slash, especially towards the core.  I could believe that a superset, subset, or an incomparable set of `void *` and `ReturnPtr` exchanges should have taken place.

This one introduces Aal::ReturnPtr, Aal::apply_bounds, and ChunkMap::amplify() as stubs for type-assisted StrictProvenance enforcement.

I agree that the AAL is probably the right place for `apply_bounds` but it (rather belatedly) occurs to me that CheriBSD defines its `VMMAP` software permission bit, which we need to clear, but doesn't interpret the other swperms, which we may wish to leave set for consumer's use.  So, we may wish to allow the Pal to modulate or influence `apply_bounds` somehow, if we don't just want to zero out the sw perms bits on CHERI arches.

nwf:
A little bit of churn; upon further consideration, it's not great that `ReturnPtr` itself is in the templated `AAL_Generic`, since that means that the AAL `Arch` classes can't use it (easily).  Push it out to a new `ds/returnptr.h`.

nwf:
Phew!  OK, so, with this much larger collection of commits (which are *hopefully* detangled enough to ease reviewing) and an initial CHERI-bounds-setting AAL, `func-malloc-1` passed on CHERI-MIPS.

There are probably still missing amplifications (`func-memory-1` seems to be tickling one at the very least, from the looks of it) but this is probably a good enough prototype that asking people to look at the full set of functions stapled on to the AAL is a good idea.

nwf:
ETA: These patches are not intended to improve temporal safety; rather, as something of an experiment, it tries to leave the pointers on free queues bounded and suitable for immediate reuse.  

That doesn't really matter for the sweeping revocation approach we've been taking -- the free queues are going to hold things only post-quarantine -- but it maybe is the wrong approach for MTE, where we need to be repainting objects' colors.

There are known spatial safety gotchyas, too.  Notably, all ReturnPtr-s are bounded to the underlying allocation, including all padding.  So if `realloc()` shrinks in place, the bounds are not changed.  For temporal safety, we generally took the stance that we could have at most one of in-place shrinking or in-place growth.  Because shrinking was much less common, I believe the Cornucopia version of `snmalloc` always moved on shrinking but permitted growth within the sizeclass (i.e., where bounds were not actually changing).

nwf:
That's maybe reasonable, but AIUI the size of chunk metadata differs between the types.  Would you put enough room for a Superslab's worth of metadata in the ChunkMap?

Relatedly, the ChunkMap on CHERI, with 16 bytes per 16M chunk of AS, is big enough (more than `INT_MAX` bytes!) that it's blowing the minds of our linker and our kernel both.  I've some patches applied to move `bits::ADDRESS_BITS` to `Pal::ADDRESS_BITS` and have taught the FreeBSD PAL that on MIPS AALs, it has only 39 bits worth of address space to manage.  This happens to be true, but is vaguely uncomfortable, as it makes the actual mappable space something of an ABI parameter.  Maybe it should be?

mjp41:
@nwf I have had issues in experiments where there are more than 4GiB of globals, the relocations required - mcmodel=large. Is this related to what you are describing?

nwf:
Related, but the checks I'm slamming into are specific to CHERI's llvm and the CheriBSD kernel.  Notwithstanding the `ADDRESS_BITS` patches and what follows, I aim to remove these limitations on our side.

While I understand the appeal of simplicity as simply having a big wad of zeros in `.bss`, could we instead `mmap` (or equivalent) in said big wad?  Would there be additional overheads?

mjp41:
Having it in bss, means the loader updates the constant to point directly at the block of memory. This means we don't have to follow an indirection, which I think we would if it was in `.ctors`.  

Also, if any other thing in `.ctors` performs allocation, then we need to be careful about accessing the pagemap, or test that it has been initialised.  This is mostly okay for allocation as that only accesses the pagemap on the slow path, but deallocation accesses it on a fast path.  Now deallocation should never be first, so that can probably be handled correctly.  But there are a few other things we support such as `external_pointer`.  We currently allow that to be called on any location, if it is not an object allocated by snmalloc, we give the start/end of the address space, as this allows us to check bounds on snmalloc allocated objects, and not give spurious failures on other objects. That would now require a test on the fast path if the `.ctors` has run.

We could prototype to see if the indirection costs a noticable amount.

PR #201, opened 2020-05-25
OpenEnclave PAL: Store enclave heap base/end in inline variables.

anakrish:
PALOpenEnclave object is lazily constructed. I couldn't
figure out a straight-forward way to pass the heap bounds to
the constructor of PALOpenEnclave object.
As an alternative, store the bounds in class-static members of a template
class.

- two_alloc_types/main.cc
  OE use of snmalloc is renamed to snmalloc_enclave.
  Declare oe_base and oe_end inline variables in snmalloc_enclave
  namespace and initialize them.

- fixed_region/fixed_region.cc
  Initialize oe_base and oe_end variables.
  Cache the starting value of oe_base in orig_oe_base and
  use that for asserting that the allocated memory lies
  within the PAL heap. This is because, in pal_open_enclave.h,
  reserve will change the value of oe_base.

CC: @achamayou 
Signed-off-by: Anand Krishnamoorthi <anakrish@microsoft.com>

### DIFF
--- a/src/pal/pal_open_enclave.h
+++ b/src/pal/pal_open_enclave.h
@@ -2,8 +2,6 @@
 
 #include "pal_plain.h"
 #ifdef OPEN_ENCLAVE
-extern "C" const void* __oe_get_heap_base();
-extern "C" const void* __oe_get_heap_end();
 extern "C" void* oe_memset_s(void* p, size_t p_size, int c, size_t size);
 extern "C" [[noreturn]] void oe_abort();
 
@@ -11,9 +9,19 @@ namespace snmalloc
 {
   class PALOpenEnclave
   {
-    std::atomic<void*> oe_base = nullptr;
+    static inline std::atomic<void*> oe_base;
+    static inline void* oe_end = nullptr;
 
   public:
+    /**
+     * This will be called by oe_allocator_init to set up enclave heap bounds.
+     */
+    static void setup_initial_range(void* base, void* end)
+    {
+      oe_base = base;
+      oe_end = end;
+    }
+
     /**
      * Bitmap of PalFeatures flags indicating the optional features that this
      * PAL supports.
@@ -28,17 +36,9 @@ namespace snmalloc
     template<bool committed>
     void* reserve(size_t size) noexcept
     {
-      if (oe_base == 0)
-      {
-        void* dummy = NULL;
-        // If this CAS fails then another thread has initialised this.
-        oe_base.compare_exchange_strong(
-          dummy, const_cast<void*>(__oe_get_heap_base()));
-      }
-
       void* old_base = oe_base;
       void* next_base;
-      auto end = __oe_get_heap_end();
+      auto end = oe_end;
       do
       {
         auto new_base = old_base;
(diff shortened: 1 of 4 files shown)


src/pal/pal_open_enclave.h
```diff
@@ -9,6 +9,20 @@ extern "C" [[noreturn]] void oe_abort();
 
 namespace snmalloc
 {
+  // Use a template to hold the heap base and end variables.
+  // Templates prevent multiple-definition error for these variables,
+  // without having to use a .cpp file otherwise.
+  template<typename T>
+  struct HeapBounds
+  {
+    static std::atomic<void*> heap_base;
```
mjp41, line 18: We can use C++ inline statics here to remove the need for the template.
```C++
inline static std::atomic<void*> heap_base;
```

src/pal/pal_open_enclave.h
```diff
@@ -9,6 +9,20 @@ extern "C" [[noreturn]] void oe_abort();
...
+    static void* heap_end;
+  };
+  template<typename T>
+  std::atomic<void*> HeapBounds<T>::heap_base;
+  template<typename T>
+  void* HeapBounds<T>::heap_end;
+
   class PALOpenEnclave
   {
     std::atomic<void*> oe_base = nullptr;
```
mjp41, line 28: This fields should be dropped.

anakrish: Fixed.

src/pal/pal_open_enclave.h
anakrish: I had to use `inline` variables instead of `inline static` so that I could initialize these in src/test/func/two_alloc_types/main.cc

anakrish:
@mjp41 Thanks for the suggestions. Can you review the latest iteration?

src/pal/pal_open_enclave.h
mjp41: I think `inline static` would be required if you kept the struct. But the `inline` global is perfect here.  Thanks for addressing.

src/pal/pal_open_enclave.h
```diff
@@ -2,17 +2,18 @@
...
 #include "pal_plain.h"
 #ifdef OPEN_ENCLAVE
-extern "C" const void* __oe_get_heap_base();
-extern "C" const void* __oe_get_heap_end();
 extern "C" void* oe_memset_s(void* p, size_t p_size, int c, size_t size);
 extern "C" [[noreturn]] void oe_abort();
 
 namespace snmalloc
 {
+  // The following bounds variables will be setup via the
```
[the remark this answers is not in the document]
anakrish: Nice suggestion. I have incorporated it.

src/test/func/two_alloc_types/main.cc
```diff
@@ -1,22 +1,11 @@
 #include "../../../snmalloc.h"
 
+#include <atomic>
```
mjp41, line 3: Is this change required?

anakrish: Good catch. Removed.

anakrish:
> Thanks for updating it. Just a minor clangformat issue, then this is good to go.
> 
> You can do
> 
> ```
> ninja clangformat
> ```
> 
> to run clangformat with the relevant settings.

Done. Looks like the CI is happy too.

PR #203, opened 2020-05-26
WIP: MIPS and RISC-V AALs

nwf:
These may or may not be ready for merging; only really tested with CHERI-MIPS and compilation testing on RISC-V.  Still, opinions welcome.


mjp41:
How do we move this away from WIP?  Should we add CI for these platforms using QEMU?

nwf:
Having CI would be great, yea.  I wonder if qemu's userspace support is good enough to run the snmalloc tests on x86-64 platforms?

mjp41:
It does for ARM, but I have no idea about other architectures. If you can set up the Docker images like @achamayou did for ARM, then happy to add it to CI.  @achamayou might be able to give you pointers if needed.

nwf:
Here's a first stab at the CI bits for Linux/MIPS.  I do not expect them to work as the Docker images are not built, but "commit early and often", right?  Do these look right?

achamayou:
@nwf sorry I missed the notification on this, but the setup looks good to me.

src/aal/aal.h
```diff
@@ -23,6 +23,15 @@
...
 #endif
 
+#if defined(__mips__)
+#  define PLATFORM_IS_MIPS
+#endif
+
+#if defined(__riscv)
+#  if __riscv_xlen == 64
+#    define PLATFORM_IS_RV64
+#  endif
```
jrtc27, line 33: I don't see why this needs to care about XLEN. The generic AAL bits don't care about 32-bit vs 64-bit X86/Arm/PowerPC/MIPS (though some of those probably lack 32-bit support). Just have a PLATFORM_IS_RISCV and an `aal_riscv.h`, with SNMALLOC_VA_BITS_32 vs SNMALLOC_VA_BITS_64 conditioned on `__riscv_xlen` inside the header itself?

ci/linux_mips64el
```diff
@@ -0,0 +1,7 @@
+FROM multiarch/debian-bootstrap:mips64el-bullseye
```
jrtc27, line 1: You probably want buster not bullseye? buster is current stable, bullseye is current testing.

nwf:
@achamayou Can you add me to the snmalloc docker organization so I can make the images for this?

src/aal/aal_mips.h
```diff
@@ -0,0 +1,36 @@
+#pragma once
+
+#define SNMALLOC_VA_BITS_64
```
davidchisnall, line 3: We probably need to guard this on a `__mips64` or similar define.  Looking at the rest of the PAL, there's no reason that it couldn't work with the n32 or o32 ABIs though, so maybe we should define it conditionally on those.  The `__mips_n64`, `__mips_n32`, and `__mips_o32` macros can be used to differentiate on these.

src/aal/aal_mips.h
```diff
@@ -0,0 +1,36 @@
...
+
+    static constexpr size_t smallest_page_size = 0x1000;
+
+    static constexpr int aal_name = MIPS;
+
+    static void inline pause()
+    {
+      ;
+      /*
+       * XXX the PAUSE instruction could be correct or not, depending on the
```
davidchisnall, line 24: I don't see a `PAUSE` instruction in the MIPS64 ISA reference.  What is this referring to?

src/aal/aal_riscv64.h
```diff
@@ -0,0 +1,34 @@
...
+
+#define SNMALLOC_VA_BITS_64
+
+namespace snmalloc
+{
+  /**
+   * RISC-V 64 architecture layer, phrased as generically as possible.  Specific
+   * implementations may need to adjust some of these.
+   */
+  class AAL_RV64
```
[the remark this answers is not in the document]
jrtc27: Does it even need a different `aal_name` if they're going to be otherwise identical, or can it just be `RISCV` for both?

src/aal/aal_mips.h
jrtc27: Those are FreeBSD-specific (Clang will define those all the time, but GCC doesn't for Linux); the correct preprocessor checks for MIPS are checking `_MIPS_SIM` against `_ABIO32`/`_ABIN32`/`_ABI64`.

achamayou:
@nwf happy to, please send me your docker hub id.

src/aal/aal_mips.h
nwf: Terribly belatedly, MIPS64 Volume II-A, Revision 6.05 (https://s3-eu-west-1.amazonaws.com/downloads-mips/documents/MIPS_Architecture_MIPS64_InstructionSet_%20AFP_P_MD00087_06.05.pdf) defines a PAUSE instruction on page 374.  It snoops on the LL bit, which is why it may or may not be the right thing to do.

SchrodingerZhu:
For CI, I think http://musl.cc/ provides all we need to compile the library. and then we can simply use QEMU userspace program to run the tests.

nwf:
That would probably be better than running the full cross-compilation under qemu-userspace emulation, yes.

SchrodingerZhu:
It seems that pause hint for riscv is already merged?

SchrodingerZhu:
```
PAUSE is encoded as a FENCE instruction withpred=W,succ=0,fm=0,rd=x0, andrs1=x0.
```

nwf:
> ```
> PAUSE is encoded as a FENCE instruction withpred=W,succ=0,fm=0,rd=x0, andrs1=x0.
> ```

Yes, it's just (AFAIK) not yet in LLVM, or, well, at least not in the CHERI branch.  I've just been waiting for that to land, since this hasn't been urgent.

davidchisnall:
> It seems that pause hint for riscv is already merged?

I didn't pay attention to the end or the process, but last time I looked this was being pushed through the fast-track process and the defined semantics were somewhat useless for software.  The semantics that we want are a hint to the CPU that it's a spin loop and so shouldn't fill up the pipeline speculating instructions on the assumption that the spin-lock variable hasn't changed, which are guaranteed to be either redundant work or cancelled instructions.  The RISC-V semantics are 'stop retiring instructions for *n* cycles, where *n* is implementation defined'.

nwf:
The RISC-V aspect here is, at present, superseded by #402 (but could move), and it looks like the MIPS aspect may be superseded by #399.  I think this PR's time has come.

PR #206, opened 2020-05-27
Make binaries more compatible by default

mjp41:
Turn `-march=native` off by default. This makes binaries more portable,
but may harm performance. However, fast paths look unaltered

Fixes 204.


CMakeLists.txt
```diff
@@ -169,12 +170,14 @@ if(NOT DEFINED SNMALLOC_ONLY_HEADER_LIBRARY)
...
     endif()
 
-    if((${CMAKE_SYSTEM_PROCESSOR} STREQUAL "x86_64") OR
-       (${CMAKE_SYSTEM_PROCESSOR} STREQUAL "x86") OR
-       (${CMAKE_SYSTEM_PROCESSOR} MATCHES "arm"))
-      check_cxx_compiler_flag(-march=native SUPPORT_MARCH_NATIVE)
-      if (SUPPORT_MARCH_NATIVE)
-        add_compile_options(-march=native)
+    if(SNMALLOC_OPTIMISE_FOR_CURRENT_MACHINE)
+      if((${CMAKE_SYSTEM_PROCESSOR} STREQUAL "x86_64") OR
```
Theodus, line 174: I think that if the user explicitly requests architecture-specific optimization, they should get it regardless of the system processor.

PR #207, opened 2020-05-28
Add some default initialisers.

mjp41:


mjp41:
This is required for VS 2019 build.  This is required to fix https://github.com/microsoft/verona/issues/125
along with other changes I will submit to Verona once this lands.

PR #209, opened 2020-06-06
Fix sizeclass rounding error

Theodus:


mjp41:
@aganea, FYI this might have negatively impacted your LLD results.

PR #210, opened 2020-06-07
WIP: Update default clang tools to stable branch for CI

Theodus:


Theodus:
The docker images for Linux platforms will need to be updated for CI to pass.

PR #212, opened 2020-06-17
New PAL for open enclave

mjp41:
This changes the implementation of the open enclacve pal to support
aligned allocations.  This reduces the amount of memory required for
the initial reservation as the large allocator doesn't have to
overallocate to get alignment.


mjp41:
FYI @anakrish

mjp41:
I added the new configuration for OE to this branch, and a bunch of tests that the general slab/superslab combination works.

PR #215, opened 2020-06-23
fix mingw

SchrodingerZhu:
please wait for test.

### DIFF
--- a/src/pal/pal_windows.h
+++ b/src/pal/pal_windows.h
@@ -6,6 +6,7 @@
 #ifdef _WIN32
 #  ifndef _MSC_VER
 #    include <cstdio>
+#    include <utility>
 #  endif
 #  define WIN32_LEAN_AND_MEAN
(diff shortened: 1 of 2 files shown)


SchrodingerZhu:
"unknown conversion type character 'z' in format"

SchrodingerZhu:
@mjp41 sorry, but I suggest revert this commit; since it only partially fixes the problem. I run into some other problems under debug mode.

SchrodingerZhu:
it shows a drawback of not having a pipeline for mingw at the pipeline. Is it possible to add one here?

mjp41:
@SchrodingerZhu if you PR a pipeline, then we will take it.  You can create the PR as draft while you are experimenting, and that should allow you to test CI without emailing everyone for each commit. 

I also suggest you disable all the other pipelines while you are testing to get faster turnaround time. I.e. initially delete all the other pipelines from the azurepipelines.yaml file.

PR #220, opened 2020-06-30
Reduce code duplication.

davidchisnall:
The Apple and Haiku PALs both had *almost* identical code to the POSIX
PAL, differing only in some small argument variables.  This is fragile
and easy to accidentally get out of sync.  For example, the changes to
`reserve_at_least` involved copying identical code into multiple PALs
and it's easy to accidentally miss one.

This change introduces two optional fields on POSIX-derived PALs:

 - `default_mmap_flags` allows a PAL to provide additional `MAP_*`
   flags to all `mmap` calls.
 - `anonymous_memory_fd` allows the PAL to override the default file
   descriptor used for memory mappings.

If a PAL does not provide these, default values are used.

### DIFF
--- a/src/pal/pal_posix.h
+++ b/src/pal/pal_posix.h
@@ -30,6 +30,64 @@ namespace snmalloc
   template<class OS>
   class PALPOSIX
   {
+    /**
+     * Helper class to access the `default_mmap_flags` field of `OS` if one
+     * exists or a default value if not.  This provides the default version,
+     * which is used if `OS::default_mmap_flags` does not exist.
+     */
+    template<typename T, typename = int>
+    struct DefaultMMAPFlags
+    {
+      /**
+       * If `OS::default_mmap_flags` does not exist, use 0.  This value is
+       * or'd with the other mmap flags and so a value of 0 is a no-op.
+       */
+      static const int flags = 0;
+    };
+
+    /**
+     * Helper class to access the `default_mmap_flags` field of `OS` if one
+     * exists or a default value if not.  This provides the version that
+     * accesses the field, allowing other PALs to provide extra arguments to
+     * the `mmap` calls used here.
+     */
+    template<typename T>
+    struct DefaultMMAPFlags<T, decltype((void)T::default_mmap_flags, 0)>
+    {
+      static const int fd = T::default_mmap_flags;
+    };
+
+    /**
+     * Helper class to allow `OS` to provide the file descriptor used for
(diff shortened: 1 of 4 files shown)


davidchisnall:
@devnexuen / @dcarlier-afilias, this touches your contributions, please test / review!

devnexen:
Works perfect under Apple but I have build issue under Haiku due to the posix flags member field (`no member named 'flags' in ... snmalloc::PALPOSIX<snmalloc::PALHaiku>`... line 220)

src/pal/pal_posix.h
```diff
@@ -30,6 +30,64 @@ namespace snmalloc
...
+    /**
+     * Helper class to access the `default_mmap_flags` field of `OS` if one
+     * exists or a default value if not.  This provides the version that
+     * accesses the field, allowing other PALs to provide extra arguments to
+     * the `mmap` calls used here.
+     */
+    template<typename T>
+    struct DefaultMMAPFlags<T, decltype((void)T::default_mmap_flags, 0)>
+    {
+      static const int fd = T::default_mmap_flags;
```
davidchisnall, line 57:
```suggestion
      static const int flags = T::default_mmap_flags;
```

Fix copy-and-paste error.

davidchisnall:
@devnexen, should be fixed now, let me know if it works!

devnexen:
Still issue but this time from the AnonFd struct (line 88) which only serves darwin.

davidchisnall:
What's the error?  Both cases for that (if the PAL does or doesn't provide the field) look  correct now.

devnexen:
`anonymous_memory_fd is not a member of 'snmalloc::PALHaiku'`

src/pal/pal_posix.h
```diff
@@ -30,6 +30,64 @@ namespace snmalloc
...
+       */
+      static const int fd = -1;
+    };
+
+    /**
+     * Helper class to allow `OS` to provide the file descriptor used for
+     * anonymous memory. This exposes the `anonymous_memory_fd` field in `OS`.
+     */
+    template<typename T>
+    struct AnonFD<T, decltype((void)T::default_mmap_flags, 0)>
```
davidchisnall, line 80:
```suggestion
    struct AnonFD<T, decltype((void)T::anonymous_memory_fd, 0)>
```

davidchisnall:
Thanks.  Two copy-and-paste errors.  This should work with Haiku now.

PR #221, opened 2020-06-30
Fix external pagemap usage.

davidchisnall:
At some point in the refactoring, these were broken.

### DIFF
--- a/src/test/func/external_pagemap/external_pagemap.cc
+++ b/src/test/func/external_pagemap/external_pagemap.cc
@@ -1,0 +1,21 @@
+#define SNMALLOC_EXPOSE_PAGEMAP 1
+#  include <override/malloc.cc>
+
+int main()
+{
+	auto &p = ExternalGlobalPagemap::pagemap();
+	auto &global = GlobalPagemap::pagemap();
+	assert(&p == &global);
+	// Get a valid heap address
+	uintptr_t addr = reinterpret_cast<uintptr_t>(malloc(42));
+	// Make this very strongly aligned
+	addr &= ~0xfffffULL;
+	void *page = p.page_for_address(addr);
+	assert(page == p.page_for_address(addr + 128));
+	assert(page == p.page_for_address(addr + 128));
+	size_t idx = p.index_for_address(addr);
+	size_t idx2 = p.index_for_address(addr + SUPERSLAB_SIZE);
+	// If the pagemap ends up storing things that are not uint8_t, this test
+	// will need modifying.
+	assert(idx2 = ((idx + 1) % OS_PAGE_SIZE));
+}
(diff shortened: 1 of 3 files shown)


mjp41:
Could you add unit tests, so these don't break in the future?

davidchisnall:
I've been pondering how to do this.  These only show up when snmalloc is used as the malloc implementation in libc and also in another library.

mjp41:
So we have tests that build two different allocators, could you modify that to cover this case:
https://github.com/microsoft/snmalloc/tree/master/src/test/func/two_alloc_types

Does it have to be in libc?

davidchisnall:
Possibly not.  I'll have a look at whether we can make one of those find the other's pagemap.

davidchisnall:
I've added a unit test.  This is blocking work in Verona, so it would be good to get it merged soon.

src/test/func/external_pagemap/external_pagemap.cc
```diff
@@ -0,0 +1,21 @@
...
+	auto &p = ExternalGlobalPagemap::pagemap();
+	auto &global = GlobalPagemap::pagemap();
+	assert(&p == &global);
+	// Get a valid heap address
+	uintptr_t addr = reinterpret_cast<uintptr_t>(malloc(42));
+	// Make this very strongly aligned
+	addr &= ~0xfffffULL;
+	void *page = p.page_for_address(addr);
+	assert(page == p.page_for_address(addr + 128));
+	assert(page == p.page_for_address(addr + 128));
```
mjp41, line 15: Duplicate assert?

src/test/func/external_pagemap/external_pagemap.cc
```diff
@@ -0,0 +1,21 @@
+#define SNMALLOC_EXPOSE_PAGEMAP 1
+#  include <override/malloc.cc>
+
+int main()
+{
+	auto &p = ExternalGlobalPagemap::pagemap();
+	auto &global = GlobalPagemap::pagemap();
+	assert(&p == &global);
```
mjp41, line 8: Can you use `SMALLOC_CHECK` in the test please.  This should be checked in release builds.

mjp41 (approved):
Looks good minor fixes

src/test/func/external_pagemap/external_pagemap.cc
davidchisnall: `SNMALLOC_CHECK` was added after this PR was originally raised.  I've rebased it on master.

src/test/func/external_pagemap/external_pagemap.cc
```diff
@@ -0,0 +1,30 @@
+#if defined(SNMALLOC_PASS_THROUGH) || defined(_WIN32)
```
mjp41, line 1: If you set a `SNMALLOC_NAME_MANGLE` then you can make the test work on Windows.

davidchisnall: That's not the problem.  The Windows declarations for the standard malloc functions have a bunch of declspecs in them that we're not using and this causes compile failures.

mjp41: If you had set a name mangle then they wouldn't conflict.

Issue #224, opened 2020-07-05
Add back off strategy for large reservations 


mjp41:
In the POSIX and Windows PALs, if aligned allocation isn't provided we reserve a large amount of memory, so that we can perform alignment work in the `AddressSpaceManager`.  However, if the large reservation fails, then the system will determine out-of-memory.

The PALs should attempt to reduce the size of reservation in the failure case, so that the program can continue working if possible.

SchrodingerZhu:
I see. I mentioned this at #223 because it is somehow related to the allocation overhead, but now I can understand the true reason is behind this reservation behavior.

mjp41:
@SchrodingerZhu thanks for your investigation into the other issue, and report this finding.  It is important to get this right.

PR #225, opened 2020-07-05
add limited test

SchrodingerZhu:
corresponding tests for #224.

### DIFF
--- a/src/test/func/memory/memory.cc
+++ b/src/test/func/memory/memory.cc
@@ -4,8 +4,63 @@
 #include <test/setup.h>
 #include <test/xoroshiro.h>
 #include <unordered_set>
+#ifdef __linux__
+#  include <sys/resource.h>
+#  include <sys/sysinfo.h>
+#  include <unistd.h>
+#  include <wait.h>
+#  define TEST_LIMITED
+#  define KiB (1024ull)
+#  define MiB (KiB * KiB)
+#  define GiB (KiB * MiB)
+#endif
 
 using namespace snmalloc;
+
+#ifdef TEST_LIMITED
+void test_limited(rlim64_t as_limit)
+{
+  auto pid = fork();
+  if (!pid)
+  {
+    auto limit = rlimit64{.rlim_cur = as_limit, .rlim_max = RLIM64_INFINITY};
+    if (setrlimit64(RLIMIT_AS, &limit))
+    {
+      std::abort();
+    }
+    if (getrlimit64(RLIMIT_AS, &limit))
+    {
+      std::abort();
+    }
+    std::cout << "limiting memory to " << limit.rlim_cur / KiB << " KiB"
+              << std::endl;
+    struct sysinfo info
+    {};
+    if (sysinfo(&info))
+    {
+      std::abort();
+    }
+    std::cout << "host freeram: " << info.freeram / KiB << " KiB" << std::endl;
+    auto upper_bound =
+      std::min(static_cast<unsigned long long>(limit.rlim_cur >> 1u), 2 * GiB);
+    upper_bound = std::min(
+      upper_bound, static_cast<unsigned long long>(info.freeram >> 1u));
+    std::cout << "trying to alloc " << upper_bound / KiB << " KiB" << std::endl;
+    auto alloc = ThreadAlloc::get();
+    auto chunk = alloc->alloc(upper_bound);
+    alloc->dealloc(chunk);
+    std::cout << "success" << std::endl;
+    std::exit(0);
+  }
+  else
+  {
+    int status;
+    waitpid(pid, &status, 0);
+    if (status)
+      std::abort();
+  }
+}
+#endif
 
 void test_alloc_dealloc_64k()
 {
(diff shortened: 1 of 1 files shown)


SchrodingerZhu:
@mjp41 I do not know whether `snmalloc` promises to be able to allocate a block of half of the available VM, but this test addresses the problem discussed in #224.

mjp41:
> I do not know whether snmalloc promises to be able to allocate a block of half of the available VM

This is not something I think we can promise, we make certain alignment assumptions to make aligned allocation very fast.  This mean we can't promise half.  Probably quarter of the size is true, but I am not sure I want to set anything like that as a guarantee.

src/test/func/memory/memory.cc
```diff
@@ -313,7 +369,11 @@ void test_calloc_large_bug()
 int main(int argc, char** argv)
 {
   setup();
-
+#ifdef TEST_LIMITED
+  test_limited(8 * GiB);
```
mjp41, line 373: This test is asking for 8GiB of memory, but is not checking for 32bit?  Seems that this could be bad. Could you wrap the 8GiB and 2 GiB test to check for 64bit platform?  
```C++
if constexpr (bits::is64())
```

SchrodingerZhu: It is actually setting the `RLIMIT_AS` to `8GiB`, the size to allocate the finally determined by `sysinfo.freeram` and upper-bounded by `2 GiB`. I was just using `8GiB` to provide a loose limit.

mjp41: So the first test shouldn't fail on 32bit, as it should do anything, and that test is passing. But the second one is failing. 
```
20/36 Test #23: func-memory-16 ...................***Exception: Child aborted  0.43 sec
limiting memory to 8388608 KiB
host freeram: 1559880 KiB
trying to alloc 194985 KiB
allocator initialised
success
...
limiting memory to 2097152 KiB
host freeram: 1558872 KiB
trying to alloc 194859 KiB
Out of memory
```
So it seems either the 2GiB test is failing, or tests are failing after 8GiB.  Could you order the tests in increasing size?  That would give us another piece of information.  Or even make it run them all and accumulate failures?  I believe this will be a QEMU issue as x86 32bit Linux is working.

SchrodingerZhu:
now the result is
```
limiting memory to 524288 KiB
host freeram: 1534232 KiB
trying to alloc 65536 KiB
Out of memory

failed
limiting memory to 2097152 KiB
host freeram: 1524536 KiB
trying to alloc 190567 KiB
Out of memory

failed
limiting memory to 8388608 KiB
host freeram: 1483792 KiB
trying to alloc 185474 KiB
allocator initialised
success
```
It is not obvious for me why ARM is failing.

SchrodingerZhu:
I assume there is something different with `setrlimit` of `qemu-arm`:
```
limiting memory to 18014398509481983 KiB
host freeram: 1953088 KiB
trying to alloc 244136 KiB
allocator initialised
```
The above result is generated from my PC. Thou it succeeded, the `rlmit_cur` is not obviously correct.

mjp41:
So I ran the following program QEMU image we are using:
```C++
#include <sys/resource.h>
#include <sys/sysinfo.h>
#include <sys/mman.h>

#include <iostream>

int main()
{
    size_t as_limit = 1 << 28;
    auto limit = rlimit64{.rlim_cur = as_limit, .rlim_max = RLIM64_INFINITY};
    if (setrlimit64(RLIMIT_AS, &limit))
    {
        std::abort();
    }
    if (getrlimit64(RLIMIT_AS, &limit))
    {
        std::abort();
    }
    std::cout << "limiting memory to " << limit.rlim_cur << std::endl;
    for (size_t size_bits = 16; size_bits < 31; size_bits++)
    {    
        size_t size = 1 << size_bits;
        std::cout << "trying to mmap " << size << std::endl;
        void* p = mmap(
            nullptr,
            size,
            PROT_READ | PROT_WRITE,
            MAP_PRIVATE | MAP_ANONYMOUS,
            -1,
            0);
        if (p != MAP_FAILED)
        {
            std::cout << "success" << std::endl;
            munmap(p, size);
        }
        else 
            std::cout << "failure" << std::endl;
    }
}
```
which leads to the following output:
```
limiting memory to 268435456
trying to mmap 65536
failure
trying to mmap 131072
failure
trying to mmap 262144
failure
trying to mmap 524288
failure
trying to mmap 1048576
failure
trying to mmap 2097152
failure
trying to mmap 4194304
failure
trying to mmap 8388608
failure
trying to mmap 16777216
failure
trying to mmap 33554432
failure
trying to mmap 67108864
failure
trying to mmap 134217728
failure
trying to mmap 268435456
failure
trying to mmap 536870912
failure
trying to mmap 1073741824
failure
```

Not sure what we can do if `setrlimit64` prevents basically any size of subsequent `mmap` call.

SchrodingerZhu:
hopefully, the problem in #224 is already solved. I think we can just let it go and close this test PR?

mjp41:
So #228 puts the flag for the QEMU builds in (not sure when it got removed). We can then have your test, but only when that flag is not set.

mjp41:
@schrodingerzhu, if you use the flag SNMALLOC_QEMU_WORKAROUND

mjp41:
Sorry clicked wrong button.  Happy to take this, if you make QEMU ARM not perform the test. using `ifdef`s.

PR #226, opened 2020-07-05
Illumos systems/Solaris support.

devnexen:
MAP_NORESERVE flag usage for memory over commit permitted only
 in the mmap context.

### DIFF
--- a/src/pal/pal_solaris.h
+++ b/src/pal/pal_solaris.h
@@ -1,0 +1,29 @@
+#pragma once
+
+#if defined(__sun)
+#  include "pal_posix.h"
+
+namespace snmalloc
+{
+  /**
+   * Platform abstraction layer for Solaris.  This provides features for this
+   * system.
+   */
+  class PALSolaris : public PALPOSIX<PALSolaris>
+  {
+  public:
+    /**
+     * Bitmap of PalFeatures flags indicating the optional features that this
+     * PAL supports.
+     *
+     */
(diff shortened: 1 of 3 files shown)


SchrodingerZhu:
will this library be compiled by Solaris Studio or GCC?

devnexen:
tried with clang and gcc. would sun studio handles c++17 tough ?

src/pal/pal_solaris.h
```diff
@@ -0,0 +1,33 @@
+#pragma once
+
+#if defined(__sun)
+// clang-format off
```
davidchisnall, line 4: Why is this needed?

devnexen: ah ... clang-format had the tendency to reorder the inclusions.

davidchisnall: It shouldn't reorder between blocks - isn't the blank line between them enough?

devnexen: ah ... still reordering it.

mjp41: I would sooner you add 
```
#include <strings.h>
```
into the Pal posix header.  That is calling `bzero`, which is coming from `strings.h`.  I guess other platforms have exposed it without requiring the inclusion.

devnexen: I considered this but was not sure all posix systems has this. We can always try.

PR #227, opened 2020-07-06
Implement PAL backoff

mjp41:
The POSIX PAL and in some configurations the Windows PAL overallocate
address space. If address space has become exhausted then this can lead
to issues, where the PAL fails, even though there is enough aligned
address space to satisfy the underlying request.

This commit tries increasingly smaller overallocation sizes in an
attempt to succeed in more cases.

### DIFF
--- a/src/pal/pal_posix.h
+++ b/src/pal/pal_posix.h
@@ -208,23 +208,30 @@ namespace snmalloc
      */
     std::pair<void*, size_t> reserve_at_least(size_t size) noexcept
     {
+      SNMALLOC_ASSERT(size == bits::next_pow2(size));
+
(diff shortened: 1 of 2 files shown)


mjp41:
@schrodingerzhu I think this addresses the issue you found.

SchrodingerZhu:
#225 is a test for this. 
- Now the initialization has no problem
- Bash or other smaller programs run smoothly.
- A large allocation (1/4 and 1/2 of AS) has a high probability to fail. (So I loose the `upper_bound` to 1/8). As you have mentioned, maybe this failure is acceptable.

SchrodingerZhu:
so on ARM Linux 32bit, the allocator still fails to initialize when AS is limited.

PR #228, opened 2020-07-08
Add missing QEMU build flag.

mjp41:


achamayou:
@mjp41 that change seems reasonable but I am confused about why the tests passed without it...

mjp41:
I guess we aren't callocing a large enough block in the testsuite.

PR #229, opened 2020-07-08
Change default chunksize to 1MiB

mjp41:
This change makes the original 16MiB option not the common option.

It also changes the names of the defines to
  SNMALLOC_USE_LARGE_CHUNKS
  SNMALLOC_USE_SMALL_CHUNKS

The second should be set for Open Enclave configuration, and results in
256KiB chunk sizes.  The first being set builds the original 16MiB chunk
sizes.  If neither is set, then we default to 1MiB chunk sizes.


mjp41:
@schrodingerzhu this is going to need changes downstream in `snmalloc-rs`

mjp41:
Once, I land this changes and a few other PRs I plan to release, 0.5.

mjp41:
@achamayou @anakrish this change will change the defines needed in `allocator.cpp`, but makes things a bit clearer.

https://github.com/openenclave/openenclave/blob/469f75adffa8aa302315db469af7e233f2efd41e/3rdparty/snmalloc/allocator.cpp#L24-L25

will become
```c++
#define SNMALLOC_USE_SMALL_CHUNKS
```

achamayou:
@mjp41 thank you for the heads up, I'll PR the update on the OE side once 0.5 is out.

SchrodingerZhu:
>  this is going to need changes downstream in snmalloc-rs

will make it happen when this is merged.

PR #232, opened 2020-07-19
PAL OE: remove redundant spinlock

nwf:
Since reserve_at_least is called only by AddressSpaceManager::reserve, which
internally uses a FlagLock around this call, this lock is superfluous.

### DIFF
--- a/src/pal/pal_open_enclave.h
+++ b/src/pal/pal_open_enclave.h
@@ -1,7 +1,6 @@
 #pragma once
 
 #include "ds/address.h"
-#include "ds/flaglock.h"
 #include "pal_plain.h"
 
 #include <array>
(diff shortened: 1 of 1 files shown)


davidchisnall:
I think that is an artefact of the current implementation, rather than a guarantee of the PAL API.  We don't document that this PAL entry point is not concurrency-safe.  We also expose the PAL to other non-snmalloc consumers, so we'd need to expose a mechanism for synchronising them with the `AddressSpaceManager`.

We could potentially move the `FlagLock` into the PAL and provide a PAL interface that returned a `std::lock_guard` equivalent that we could then use in `AddressSpaceManager`.

mjp41:
I want to assume the Pal is correct in a concurrently called setting.  All the other Pal's are as they are passing to the underlying system calls.

Making this change could lead to failures in the future given refactors that might seem sensible.

Issue #231, opened 2020-07-19
Revisiting CHERI rederivation w/ 1MiB Superslabs.


nwf:
After #229 landed, the default granularity of the chunkmap is 1 MiB, meaning that a 48 bit address space requires a 256 MiB flat pagemap.  That's large, but not unreasonably huge.  Unfortunately, the approach suggested in #198 stores a full CHERI capability for every granule, which is to say that this is suddenly 4 GiB, which is... probably an unreasonable thing to ask.  The CHERI `lld` and CheriBSD kernel both look askance at objects declaring their `.bss` to be so substantial.

I wonder if it's reasonable, then, to suggest some kind of hybridized design, wherein the pagemap stores something like...

```
struct {
  void * __capability amplifier;
  enum ChunkMapSuperslabKind superkinds[16]; /* uint8_t[16], really */
}
```

This would require only 512 MiB in the flat pagemap, but requires that we be asking the kernel for 16 MiB (sized and aligned) regions of AS even if we are internally parcelling them out in 1 MiB increments, requiring that the `AddressSpaceManager` allocator be brought in to the CHERI story.  Further, the interface to the chunkmap would need to be adjusted to understand that it operates on two different granularities, one for amplification and one for `ChunkMapSuperslabKind` values.

Assuming this multiple-granularity idea passes the smell test, three follow-on design decisions present themselves:

* Rather than packing the two kinds of state into one table, should we instead use two pagemaps, one for amplifiers and one for superkinds?  Note that `dealloc(p)` will need both pieces of state (though `dealloc(p,n)` and `dealloc<n>(p)` may just need the amplifiers, unless we're `CHECK_CLIENT`-ing).

* The exact choice of "16" is not required, but seemed to get the metadata size down while not grossly inflating the required alignment of amplification granules and preserving cacheline-adjacency of amplifiers and superkinds.

* #198 was using the freedom of the `offset` field within the CHERI capability to encode the `ChunkMapSuperslabKind`, because our amplifiers are always highly-aligned.  In this design, these bits are now available for reuse, so: is there anything else we'd like to store there?

mjp41:
That seems reasonable to me.

In the highly address space constrained world we are already not using the flat map, so reserving larger space for this scenarios seems reasonable.

The Pal could perhaps contain the amplification code, then it grabs blocks that are larger.  I.e. you could make the 4GiB that is returned by `reserve_at_least` be the unit of amplification?

davidchisnall:
[...]

PR #237, opened 2020-07-31
WIP: Foreign frees

nwf:
A sketch of one possible route towards allowing `snmalloc`'s `free` to trampoline out to another allocator and possibly even across security domains.

This introduces another, separate `Pagemap` structure, the `OOBMap`, which stores a `uintptr_t` per granule, which is taken to be 16M right now.  It's not clear whether this should scale with chunk size or be completely independent.

This is a WIP; there are some scattered TODOs in the code at the moment, as this is just intended to be a sketch for design review.  The "interface", `snmalloc_designate_foreign`, is almost surely not the desired API, and so I am more than open to suggestions!

The `OOBMap` should be useful as a different substrate for #198 as per #231, but there is the concern that CHERI will require that we access both maps in many cases, whereas this PR results in code that accesses the `OOBMap` only on the slow path.  Some restructuring may be in order when we seek to merge both, but that can happen over there.

### DIFF
--- a/src/mem/alloc.h
+++ b/src/mem/alloc.h
@@ -10,6 +10,7 @@
 #include "../test/histogram.h"
 #include "allocstats.h"
 #include "chunkmap.h"
+#include "foreignalloc.h"
 #include "largealloc.h"
 #include "mediumslab.h"
 #include "oobmap.h"
@@ -243,7 +244,9 @@ namespace snmalloc
 
     /*
      * Free memory of a statically known size. Must be called with an
-     * external pointer.
+     * external pointer from this security domain.
+     *
+     * TODO: make work with foreign allocators.
      */
     template<size_t size>
     void dealloc(void* p)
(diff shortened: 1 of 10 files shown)


nwf:
Now that this is passing CI, it probably makes sense for @mjp41 and @davidchisnall to look it over when time allows.

mjp41:
So if we had
```C++
template <void backup_free(void*) = error_free>
free(void* p)
{
  auto pm = pagemap().get(p);
  ...
  if (pm == NOT_SNMALLOC)
  {
     backup_free(p);
  }
}
```

This could be useful for places where the override was unable to intercept a particular malloc, for example on 
Windows:
```C++
free<HeapFree>(p)
```

Then the smarts for determining which allocator to go to are not part of snmalloc?  But can be layered on top.  You would probably want a modified Pal, which when it requested address space registered, so that other allocators would know it is responsible for that space.

nwf:
Closing for staleness.  We can dust this back off when/if we decide it's really important, but for the moment we seem content with "you must free an object back to the allocator/sandbox that sourced it"

PR #239, opened 2020-08-18
Linux/musl test: build fix proposal.

devnexen:
Build complain about wrong header inclusion, sys/wait.h ough to be
 instead.


devnexen:
it did work as this section is only about Linux.

PR #243, opened 2020-08-27
netbsd: reserve_aligned making the segment shared to fix

devnexen:
an unit test crashing.

### DIFF
--- a/src/pal/pal_netbsd.h
+++ b/src/pal/pal_netbsd.h
@@ -24,6 +24,34 @@ namespace snmalloc
      * add new features that they should add any required feature flags.
      */
     static constexpr uint64_t pal_features = PALBSD_Aligned::pal_features;
+
+    /**
+     * Reserve memory at a specific alignment.
+     * Oddily, it appears it needs to be shared segment
+     * as it should not need to but func-memory-16 crashes otherwise.
+     */
+    template<bool committed>
+    void* reserve_aligned(size_t size) noexcept
+    {
+      // Alignment must be a power of 2.
+      SNMALLOC_ASSERT(size == bits::next_pow2(size));
+      SNMALLOC_ASSERT(size >= minimum_alloc_size);
+
+      size_t log2align = bits::next_pow2_bits(size);
+
+      void* p = mmap(
+        nullptr,
+        size,
+        PROT_READ | PROT_WRITE,
+        MAP_SHARED | MAP_ANONYMOUS | MAP_ALIGNED(log2align),
+        -1,
+        0);
+
+      if (p == MAP_FAILED)
+        PALBSD_Aligned<PALNetBSD>::error("Out of memory");
+
+      return p;
+    }
   };
 } // namespace snmalloc
 #endif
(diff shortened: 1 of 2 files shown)


src/pal/pal_netbsd.h
```diff
@@ -24,6 +24,34 @@ namespace snmalloc
...
+      SNMALLOC_ASSERT(size == bits::next_pow2(size));
+      SNMALLOC_ASSERT(size >= minimum_alloc_size);
+
+      size_t log2align = bits::next_pow2_bits(size);
+
+      void* p = mmap(
+        nullptr,
+        size,
+        PROT_READ | PROT_WRITE,
+        MAP_SHARED | MAP_ANONYMOUS | MAP_ALIGNED(log2align),
```
davidchisnall, line 46: Can we not do this by slightly extending the flag that is passed to mmap that comes via the template parameter in the superclass?

davidchisnall (commented):
This makes me *very* nervous, is there a NetBSD bug that we can track to revert once versions that don’t have a fix are EOL?

src/pal/pal_netbsd.h
devnexen: Sure why not as it s most likely a temporary fix I did not bother think of it.

mjp41:
Where does the test fail?  It would be useful to see a stack trace just to understand what might be going on.

devnexen:
`
...
superslab.h
105           for (size_t i = 0; i < SLAB_COUNT - used - 1; i++)
106           {
107             curr = (curr + meta[curr].next + 1) & (SLAB_COUNT - 1);
108           }
109           if (curr != 0)
110             abort();
...
curr = {value = 255 '\377'}
`

mjp41:
Thanks, @devnexen, I meant more which test fails.  I.e. the line in memory.cc.

mjp41:
I have no clue why MAP_SHARED would affect that line of a test. 

As @davidchisnall asks, is there a Net BSD bug we can attribute this too?  Or could this be a sign of something incorrect in snmalloc?

devnexen:
Weird indeed, but it seems the issue disappears if I avoid the aligned class inheritance but just grabs the posix one.

davidchisnall:
That makes it sound as if there's a bug in NetBSD's handling of `MAP_ALIGNED`.  Can you try adding an assert that the result of the `mmap` call is actually as aligned as we asked for?  I don't know if NetBSD does transparent superpage promotion, but you might also see a problem if it promoted but didn't then demote the pages when we split them by zeroing large chunks.  It's probably worth filing a NetBSD bug to track this - I suspect that we're hitting some code paths in the VM subsystem that are not particularly well stress-tested by other workloads.

devnexen:
NetBSD uses large pages too seems most likely the second case as otherwise the addresses are aligned. Will see if there is any existing bug reported.

mjp41:
@devnexen, any ideas why this is only on NetBSD?  I would like to merge this, but I don;t want to mask the discovery on an underlying bug in snmalloc.

devnexen:
Well I ve just tested the last master version on newly fresh netbsd installation and it passes and indeed I went back into a older git hash and was able to reproduce the bug we can most likely close this one instead.

PR #244, opened 2020-09-02
use old-style ::value

SchrodingerZhu:
`macOS`'s default clang toolchain may have some special behaviors with `type_traits`. `std::is_same_v` failed to be inferred as a `constexpr` in `snmalloc-rs` CI.

### DIFF
--- a/src/ds/bits.h
+++ b/src/ds/bits.h
@@ -74,15 +74,15 @@ namespace snmalloc
       return BITS - index - 1;
 #  endif
 #else
-      if constexpr (std::is_same_v<unsigned long, std::size_t>)
+      if constexpr (std::is_same<unsigned long, std::size_t>::value)
       {
         return static_cast<size_t>(__builtin_clzl(x));
       }
-      else if constexpr (std::is_same_v<unsigned long long, std::size_t>)
+      else if constexpr (std::is_same<unsigned long long, std::size_t>::value)
       {
         return static_cast<size_t>(__builtin_clzll(x));
       }
-      else if constexpr (std::is_same_v<unsigned int, std::size_t>)
+      else if constexpr (std::is_same<unsigned int, std::size_t>::value)
       {
         return static_cast<size_t>(__builtin_clz(x));
       }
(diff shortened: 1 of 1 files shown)


davidchisnall:
Can we perhaps instead add an `snmalloc::is_same_v` that is just `using is_same_v = std::is_same_v` by default and is a `constexpr` function with the macOS toolchain?

I'm a bit surprised that we don't have problems with this in our macOS CI, given that it's using the default macOS toolchain.

SchrodingerZhu:
it is even annoying that [a new build passed](https://github.com/SchrodingerZhu/snmalloc-rs/pull/82/checks?check_run_id=1060713206) without this patch. I will check the version info.

mjp41:
@schrodingezhu, which platform/compiler precisely is not working.

SchrodingerZhu:
only macOS with clang 9.1 on the travis CI.

SchrodingerZhu:
> using is_same_v = std::is_same_v
Did you mean?
```
namespace snmalloc {
    template <typename A, typename B>
    using is_same_v = typename std::is_same<A, B>::value;
}
```

davidchisnall:
Given how simple it is, that's probably sufficient.  I was thinking of conditional compilation to use `std::is_same_v` if it worked, but that's probably overly complex.  

Our macOS systems have AppleClang 11.0.3.11030032, which is the current default.  Is there a reason that Rust is testing with such an old version?

SchrodingerZhu:
> s there a reason that Rust is testing with such an old version?
it is just the version from `travis ci`. perhaps I could use `brew` to fetch a newer version.

davidchisnall:
Using brew to install a new compiler is a bit error prone (sometimes Apple needs AppleClang not upstream clang), but it looks as if Travis is shipping images with an old version of the XCode tools.  Is it possible to prod them to update it?  GitHub / Azure CI images have a newer version, which is why we don't see that failure.

SchrodingerZhu:
already fixed. Upgraded `xcode` to version 12+.

Issue #233, opened 2020-07-19
Should PALs be entirely static &| can the inheritance diagram be simplified?


nwf:
Presently, some methods in PALs are not `static`: `notify_using`, `notify_not_using`, `zero`, `reserve_aligned`, `reserve_at_least`, `expensive_low_memory_check`, and possibly others I've missed.

That might be OK, but AIUI the PAL actually gets instantiated repeatedly and (IMHO) confusingly (I may be about to reveal that my understanding of C++ is somewhere between "stone age" and "stumbling"; if I'm overlooking something obvious, I apologize; I did at least attempt to verify my assertions with some strategically placed `std::cerr` debugging, but that may have kicked me into an entirely different basin of attraction within the C++ standard):

* Twice on the stack, inside `MemoryProviderStateMixin<PAL>::make()`: https://github.com/microsoft/snmalloc/blob/4e1f5829a754ab9c170ec090979d3a670d2d5d1a/src/mem/largealloc.h#L85 

  * Once, since `MemoryProviderStateMixin<PAL>` extends `PAL`: https://github.com/microsoft/snmalloc/blob/4e1f5829a754ab9c170ec090979d3a670d2d5d1a/src/mem/largealloc.h#L60

  * And *again*, since `MemoryProviderStateMixin<PAL>` has a member of type `AddressSpaceManger<PAL>` and that type extends `PAL`: https://github.com/microsoft/snmalloc/blob/4e1f5829a754ab9c170ec090979d3a670d2d5d1a/src/mem/largealloc.h#L71 and https://github.com/microsoft/snmalloc/blob/4e1f5829a754ab9c170ec090979d3a670d2d5d1a/src/mem/address_space.h#L17

* "The same" twice in that method again, this time in the nascent heap: https://github.com/microsoft/snmalloc/blob/4e1f5829a754ab9c170ec090979d3a670d2d5d1a/src/mem/largealloc.h#L88-L89 more specifically at
https://github.com/microsoft/snmalloc/blob/4e1f5829a754ab9c170ec090979d3a670d2d5d1a/src/mem/largealloc.h#L191

The on-heap `address_space` gets paved over with bits from the on-stack `address_space` using `::memcpy`: https://github.com/microsoft/snmalloc/blob/4e1f5829a754ab9c170ec090979d3a670d2d5d1a/src/mem/largealloc.h#L102-L105
[...]

PR #245, opened 2020-09-07
Fully static PALs and simplified inheritance

nwf:
See #233 .

### DIFF
--- a/src/ds/aba.h
+++ b/src/ds/aba.h
@@ -85,6 +85,13 @@ namespace snmalloc
       Linked old;
       ABA* parent;
 
+      /*
+       * MSVC apparently does not like the implicit constructor it creates when
+       * asked to interpret its input as C++20; it rejects the construction up
+       * in read(), above.  Help it out by making the constructor explicit.
+       */
+      Cmp(Linked old, ABA* parent) : old(old), parent(parent) {}
+
       T* ptr()
       {
         return old.ptr;
--- a/src/ds/concept.h
+++ b/src/ds/concept.h
@@ -1,0 +1,40 @@
+#pragma once
+
+/**
+ * C++20 concepts are referenced as if they were types in declarations within
+ * template parameters (e.g. "template<FooConcept Foo> ...").  That is, they
+ * take the place of the "typename"/"class" keyword on template parameters.
+ * If the compiler understands concepts, this macro expands as its argument;
+ * otherwise, it expands to the keyword "typename", so snmalloc templates that
+ * use concept-qualified parameters should use this to remain compatible across
+ * C++ versions: "template<SNMALLOC_CONCEPT(FooConcept) Foo>"
+ */
+#ifdef __cpp_concepts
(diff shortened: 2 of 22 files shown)


src/ds/aba.h
```diff
@@ -85,6 +85,12 @@ namespace snmalloc
       Linked old;
       ABA* parent;
 
+      Cmp(Linked old, ABA* parent)
```
davidchisnall, line 88: Please add a comment explaining why we need this (MSVC bug?)

src/ds/aba.h
```diff
@@ -85,6 +85,12 @@ namespace snmalloc
...
+      {
+        this->old = old;
+        this->parent = parent;
```
davidchisnall, line 91: Is there a reason this isn't using initialiser syntax?

src/ds/concept.h
```diff
@@ -0,0 +1,21 @@
+#pragma once
+
+#ifdef __cpp_concepts
+
+#  define SNMALLOC_CONCEPT(c) c
```
davidchisnall, line 5: Doc comment.

src/mem/alloc.h
```diff
@@ -1336,7 +1335,7 @@ namespace snmalloc
 
         slab->init(public_state(), sizeclass, rsize);
         chunkmap().set_slab(slab);
-        p = slab->alloc<zero_mem>(size, large_allocator.memory_provider);
+        p = slab->alloc<zero_mem, MemoryProvider>(size);
```
mjp41, line 1338: Would it make more sense to take
```C++
MemoryProvider::Pal
```
as the template parameter?

src/mem/slab.h
```diff
@@ -37,11 +37,8 @@ namespace snmalloc
      * `fast_free_list` for further allocations.
      */
     template<ZeroMem zero_mem, typename MemoryProvider>
```
mjp41, line 39: Just take the Pal as a static(template) parameter?

src/mem/mediumslab.h
```diff
@@ -77,7 +77,7 @@ namespace snmalloc
     }
 
     template<ZeroMem zero_mem, typename MemoryProvider>
```
mjp41, line 79: Take Pal as template parameter?

src/pal/pal_concept.h
```diff
@@ -0,0 +1,75 @@
...
+
+#ifdef __cpp_concepts
+#  include "../ds/concept.h"
+#  include "pal_consts.h"
+
+#  include <utility>
+
+namespace snmalloc
+{
+  template<typename PAL>
```
davidchisnall, line 11: Lots of doc comments here please!

mjp41 (changes requested):
Could you update the documentation about writing a Pal:
https://github.com/microsoft/snmalloc#porting-snmalloc-to-a-new-platform

This needs littering with `static` now.

mjp41 (approved):
LGTM

PR #247, opened 2020-09-11
Haiku debug build, rdynamic flag unsupported.

devnexen:


CMakeLists.txt
```diff
@@ -175,10 +175,10 @@ if(NOT DEFINED SNMALLOC_ONLY_HEADER_LIBRARY)
     # Static TLS model unsupported on Haiku
```
nwf, line 175: Mind updating this comment while you're at it?

davidchisnall:
Is this just a compiler driver bug? Does -Wl,-rdynamic work? If so, that should work everywhere and we can fix at least the clang driver to forward -rdynamic to the linker quite easily to prevent this being a problem for other projects in the future.

devnexen:
It s really unsupported (even what you propose).

davidchisnall:
That’s surprising. -rdynamic just puts everything in the dynamic symbol table. I don’t see how you can have an ELF target that supports shared libraries and doesn’t support it.

devnexen:
All symbols are exported by default on Haiku, it is `-shared` implicitly ; it behaves differently than other oses in that matter.

davidchisnall:
I see, so `-rdynamic` is the default?  A bit surprising it isn't accepted as an argument then.  Can you add that as a comment in this PR?

devnexen:
I think I updated the comment in CMakeLists.txt earlier

CMakeLists.txt
```diff
@@ -172,13 +172,13 @@ if(NOT DEFINED SNMALLOC_ONLY_HEADER_LIBRARY)
     set(CMAKE_EXE_LINKER_FLAGS_RELEASE "${CMAKE_EXE_LINKER_FLAGS_RELEASE} /DEBUG")
   else()
     add_compile_options(-fno-exceptions -fno-rtti -g -fomit-frame-pointer)
-    # Static TLS model unsupported on Haiku
+    # Static TLS model and -rdynamic flag (always -shared) are unsupported on Haiku
```
davidchisnall, line 175:
```suggestion
    # Static TLS model is unsupported on Haiku.  
    # All symbols are always dynamic on Haiku and so -rdynamic is redundant (and unsupported).
```

davidchisnall (approved):

PR #248, opened 2020-09-22
Natural alignment for USE_MALLOC

mjp41:
Building snmalloc with USE_MALLOC defined passes through all the snmalloc calls to the system allocator.  This is primarily used by Project Verona to use sanitizers that replace the allocator like ASAN.  snmalloc provides very aligned blocks of memory, i.e. a 2^n block is guaranteed to be 2^n aligned, however, this was not the case when pass through the allocation.

This PR makes the pass through impose the alignment that snmalloc would guarantee for the allocations.


mjp41:
@devnexen any chance you could let me know how of the platforms you got working, that I don't have CI for, this breaks?

src/mem/alloc.h
```diff
@@ -6,6 +6,36 @@
...
 #endif
 
+#ifdef SNMALLOC_PASS_THROUGH
+#  include <cstdlib>
+#  if defined(_WIN32) || defined(__APPLE__)
+#    error "Pass through not supported on this platform"
+//   The Windows aligned allocation API is not capable of supporting the
+//   snmalloc API Apple was not providing aligned memory in some tests.
+#  else
+#    define ALIGNED_ALLOC std::aligned_alloc
```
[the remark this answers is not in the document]
davidchisnall: Also, note that `std::aligned_alloc` is just a C++ namespacing of the C11 `aligned_alloc`, which isn't very well supported.  On everything that's not Windows, `posix_memalign` is probably more likely to exist and work (it's from POSIX2001, so people have had 10 more years to get around to implementing it); however, POSIX does not provide a way to free memory allocated with `posix_memalign` (glibc allows it to be freed with `free`) so the whole thing is a bit of a mess.

davidchisnall (commented):
I wonder if it might be easier to teach snmalloc to update the asan bitmap than to get this working portably.

If this is only for ASan, then note that the ASan runtime intercepts and replaces `memalign` with `asan_memalign`, so calling `asan_memalign` is probably a better option.

mjp41:
Teaching snmalloc to do the right thing with ASAN is a significant undertaking.  
We would need to 
* annotate most method so they aren't instrumented by ASAN,
* mark allocated objects and deallocated objects in the shadow heap
* adjust layout to contain red zones
* potentially add a quarantine to delay reuse, but snmalloc might already have a reasonable behaviour for that a lot of the time.

It is probably worth doing, but it is not something I have time for at the moment.

I can use `posix_memalign`, and call free and hope for the best?

src/mem/alloc.h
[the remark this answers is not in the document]
mjp41: Will switching to `using` have the desired effect? If I do a direct replacement, it is an unnamespaced `using` so will affect all future files like a `#define`.  I can put it inside a namespace, but the includes will be outside the namespace, so I will be doubling up the `#ifdef` testing.

I can put the SNMALLOC prefix on, but it is kind of weird as they aren't snmalloc.  Perhaps
```C++
NOT_SNMALLOC_MSIZE
```
makes the most sense?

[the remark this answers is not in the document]
davidchisnall: If they're in the `snmalloc` namespace, then there's no pollution with `using`.  I'd generally prefer to have any macros defined by snmalloc start with `SNMALLOC` because no other library should be using that prefix by accident.  Having them start with `NOT_` is a bit weird: they're not snmalloc functions, but they are snmalloc macro definitions.

davidchisnall:
> Teaching snmalloc to do the right thing with ASAN is a significant undertaking.
> We would need to
> 
> * annotate most method so they aren't instrumented by ASAN,

Does asan have range-based opt-out?  If not, we could probably contribute one.

> * mark allocated objects and deallocated objects in the shadow heap

Yup, though this can be done at the outer alloc call, so probably isn't too invasive.

> * adjust layout to contain red zones

I think that's mostly a case of adding padding, so could be done at the outer function layer as well.

> * potentially add a quarantine to delay reuse, but snmalloc might already have a reasonable behaviour for that a lot of the time.

I think if we're doing reuse in queue mode rather than stack, that's probably fine too.

> It is probably worth doing, but it is not something I have time for at the moment.

Fair enough.

> I can use `posix_memalign`, and call free and hope for the best?

If the use case of this is using snmalloc with asan, rather than using malloc via snmalloc interfaces in the general case, using `asan_memalign` and `asan_free` might be the best option, assuming that they're exported from the asan runtime library.

davidchisnall:
Unfortunately, it looks as if libclang_rt.asan-{arch}.so does not export any symbols.  I should probably start a cfe-dev thread about the best way to do this when one of us has time to work on it.

devnexen:
> @devnexen any chance you could let me know how of the platforms you got working, that I don't have CI for, this breaks?

Tried FreeBSD amd64, suffers the same problem as Mac, aligned alloc does not pass.

src/mem/alloc.h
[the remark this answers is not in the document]
mjp41: So 
```C++
# elif defined(__FreeBSD__)
#      include <malloc_np.h>
#      using MSIZE = malloc_usable_size;
#    else
```
if this is inside a namespace, then that namespace will apply to `malloc_np.h`, which will probably cause issues?  And otherwise they are global.  So it means duplicating the `#ifdef`, both inside and outside the namespace.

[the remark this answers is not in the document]
davidchisnall: Alternatively you could write it as:

```
#    elif defined(__FreeBSD__)
#      include <malloc_np.h>
namespace snmalloc { namespace external_malloc {
        using malloc_size = malloc_usable_size;
}}
#    else
```
You could then `using namespace snmalloc::external_malloc` or similar in the `Alloc` class to import these.

mjp41:
@devexen I have switched to `posix_memalign` on @davidchisnall's advice, so hopefully that fixes FreeBSD.

I can disable Haiku for these tests.

devnexen:
Even with the last change, the alignment check still fails.

davidchisnall (approved):

CMakeLists.txt
```diff
@@ -280,7 +280,15 @@ if(NOT DEFINED SNMALLOC_ONLY_HEADER_LIBRARY)
   foreach(TEST_CATEGORY ${TEST_CATEGORIES})
     subdirlist(TESTS ${TESTDIR}/${TEST_CATEGORY})
     foreach(TEST ${TESTS})
-      foreach(SUPER_SLAB_SIZE 1;16;oe)
+      if (WIN32 OR APPLE)
```
devnexen, line 283: I still think FreeBSD should be included here it fails just as Darwin despite posix_memalign.

davidchisnall: The failure on FreeBSD isn't a FreeBSD bug, it's a bug in the realloc implementation.  We allocate 4KiB with 4KiB alignment but then `malloc_useable_size` returns 16KiB, so we don't reallocate and copy.  We then end up with a 4KiB-aligned 16KiB allocation.  We should be reallocating if the result of `malloc_useable_size` is more than three times the alignment of the allocation.  I don't know if the macOS failures have the same cause.

devnexen: Otherwise OpenBSD does not have malloc*size api nor any similar-ish call to it.

mjp41: Fixed Apple, so hopefully FreeBSD too.  Should I disable OpenBSD for these tests as the tests won't compile?

devnexen: It might be better to disable for OpenBSD indeed.

mjp41: I believe I have disabled OpenBSD. I also disabled NetBSD as I could not find a malloc*size API in the documentation.

devnexen: There is sort of malloc_usable_size present in the libc symbols but not exposed most likely not functional yet,so you re right.

devnexen (approved):

PR #252, opened 2020-10-02
DragonFly support (userland).

devnexen:
- Close to OpenBSD as there is no malloc*size api nor arbritrary
 alignment support.
- Like FreeBSD, MAP_NORESERVE never had been implemented even tough
 still present in the header but not mentioned in the man page,
FreeBSD has reserved the value for another later usage seems
 DragonFly has just out of sync header.

### DIFF
--- a/src/pal/pal_dragonfly.h
+++ b/src/pal/pal_dragonfly.h
@@ -1,0 +1,30 @@
+#pragma once
+
+#if defined(__DragonFly__) && !defined(_KERNEL)
+#  include "pal_bsd.h"
+
+namespace snmalloc
+{
+  /**
+   * DragonflyBSD-specific platform abstraction layer.
+   *
+   * This adds DragonFlyBSD-specific aligned allocation to the BSD
+   * implementation.
+   */
+  class PALDragonfly : public PALBSD<PALDragonfly>
+  {
+  public:
+    /**
+     * Bitmap of PalFeatures flags indicating the optional features that this
+     * PAL supports.
+     *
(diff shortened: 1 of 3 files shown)


src/pal/pal_dragonfly.h
```diff
@@ -0,0 +1,30 @@
...
+
+#if defined(__DragonFly__) && !defined(_KERNEL)
+#  include "pal_bsd.h"
+
+namespace snmalloc
+{
+  /**
+   * DragonflyBSD-specific platform abstraction layer.
+   *
+   * This adds DragonBSD-specific aligned allocation to the BSD
```
mjp41, line 11:
```suggestion
   * This adds DragonFlyBSD-specific aligned allocation to the BSD
```

src/pal/pal_dragonfly.h
```diff
@@ -0,0 +1,30 @@
...
+   */
+  class PALDragonfly : public PALBSD<PALDragonfly>
+  {
+  public:
+    /**
+     * Bitmap of PalFeatures flags indicating the optional features that this
+     * PAL supports.
+     *
+     * The DragonflyBSD PAL does not currently add any features beyond
+     * of the BSD.
```
mjp41, line 22:
```suggestion
     * those of the BSD Pal.
```

mjp41 (changes requested):
Thanks for doing this.  Couple of minor nits, but otherwise LGTM.

davidchisnall (approved):

mjp41 (approved):

Issue #253, opened 2020-10-12
Typo in macro definition in CMakeLists.txt


rmn30:
At https://github.com/microsoft/snmalloc/blob/49b9856ed0f4681e41e16576aff968c817f19368/CMakeLists.txt#L262 the macro definition is missing and 'L' in SNMALLOC. This means the 16MiB shared library shim has been using the default 1MiB chunk size since 4e1f5829a754ab9c170ec090979d3a670d2d5d1a .

mjp41:
Closed by #254

PR #254, opened 2020-10-12
Fix typo in macro definition for 16MiB shared library shim (#253).

rmn30:
This has been present since 4e1f5829a754ab9c170ec090979d3a670d2d5d1a and resulted in the 16MiB shared library shim using the default 1MiB chunk size.


mjp41 (approved):
Good catch.  My benchmarking to justify the change of default was done before that `#define` was introduced.

PR #255, opened 2020-10-13
pass through Haiku build fix.

devnexen:
Haiku has a malloc_usable_size api but via the posix layer.

### DIFF
--- a/src/mem/external_alloc.h
+++ b/src/mem/external_alloc.h
@@ -1,6 +1,9 @@
 #pragma once
 
 #ifdef SNMALLOC_PASS_THROUGH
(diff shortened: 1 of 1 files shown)


mjp41:
CI is not happy with the 
```
#define _GNU_SOURCE
```
It is redefining something in the headers. Perhaps, put this in Haiku only.

Also, clangformat is not happy,  could you run the clangformat build target, thanks.

devnexen:
ah right I missed the CI report. fixing.

mjp41 (approved):
LGTM, thanks for doing this.

PR #259, opened 2020-11-09
Superslab bug fix

mjp41:
Fixes the meta-data initialisation for the superslab in the 16MiB case.

This superseeds #258 as the problem is more general.


mjp41:
@devnexen could you confirm this fixes your issue from #258 ? Thanks

devnexen (approved):

davidchisnall (approved):

PR #260, opened 2020-11-10
Separate out building and porting docs.

davidchisnall:
Fix omissions in the porting doc (`page_size` and C++20 concepts).


mjp41 (approved):

davidchisnall:
@nwf, can you check that the things I've written about `page_size` and the concepts are not nonsense please?

docs/BUILDING.md
```diff
@@ -0,0 +1,119 @@
...
+snmalloc has platform abstraction layers for XNU (macOS, iOS, and so on),
+FreeBSD, NetBSD, OpenBSD, and Linux and is expected to work out of the box on
+these systems.
+Please open issues if it does not.
+Note that NetBSD, by default, ships with a toolchain that emits calls to
+`libatomic` but does not ship `libatomic`.
+To use snmalloc on NetBSD, you must either acquire a `libatomic` implementation
+(for example, from the GCC or LLVM project) or compile with clang.
+
+snmalloc has very few dependencies, CMake, Ninja, Clang 6.0 or later and a C++17
```
nwf, line 38:
```suggestion
snmalloc has very few dependencies: CMake, Ninja, Clang 6.0 or later and a C++17
```

docs/PORTING.md
```diff
@@ -0,0 +1,125 @@
...
+
+```c++
+template<bool committed>
+static void* reserve_aligned(size_t size) noexcept;
+static std::pair<void*, size_t> reserve_at_least(size_t size) noexcept;
+```
+Only one of these needs to be implemented, depending on whether the underlying
+system can provide strongly aligned memory regions.
+If the system guarantees only page alignment, implement the second. The Pal is 
+free to overallocate based on the platforms desire and snmalloc
```
nwf, line 55:
```suggestion
free to overallocate based on the platform's desire and snmalloc
```

docs/BUILDING.md
```diff
@@ -0,0 +1,119 @@
...
+libatomic for 128-bit atomic operations.
+
+To build a debug configuration:
+```
+mkdir build
+cd build
+cmake -G Ninja .. -DCMAKE_BUILD_TYPE=Debug
+ninja
+```
+To build a release configuration:
```
nwf, line 50:
```suggestion
To build a release configuration (that is, one with optimizations turned on):
```

docs/BUILDING.md
```diff
@@ -0,0 +1,119 @@
...
+`LD_PRELOAD`ed to use the allocator in place of the system allocator, for
+example, you can run the build script using the snmalloc as the allocator for
+your toolchain:
+
+```
+LD_PRELOAD=/usr/local/lib/libsnmallocshim.so ninja
+```
+
+## Cross Compile for Android
+Android support is out-of-the-box.
```
nwf, line 76: Maybe?

```suggestion
Android is supported out-of-the-box.
```

docs/PORTING.md
```diff
@@ -0,0 +1,125 @@
...
+will find suitably aligned blocks inside the region.  `reserve_at_least` should 
+not commit memory as snmalloc will commit the range of memory it requires of what 
+is returned.
+
+If the system provides strong alignment, implement the first to return memory
+at the desired alignment. If providing the first, then the `Pal` should also 
+specify the minimum size block it can provide: 
+```
+static constexpr size_t minimum_alloc_size = ...;
+```
```
nwf, line 65: This is not at present enforced by the concept, but probably should be.

docs/PORTING.md
```diff
@@ -0,0 +1,125 @@
...
+
+The PAL is also responsible for advertising the page size:
+
+```c++
+static constexpr size_t page_size = 0x1000;
+```
+
+This is the size at which the PAL is able to mark memory as in-use or not-in-use.
```
nwf, line 73:
```suggestion
This is the granularity at which the PAL is able to mark memory as in-use or not-in-use.
```

nwf (commented):
Some nits, but otherwise LGTM.

docs/BUILDING.md
davidchisnall: Release also implies no asserts and no debug info, so I think this is a misleading change.

nwf: Are asserts implied by `RelWithDebugInfo`?  Clarifying both branches might be worthwhile.

davidchisnall: No, asserts are based on the NDEBUG macro not being defined.  All of this is inherited from CMake and I'm hesitant to try to redefine core CMake configs in our build.

nwf: Sorry, to be clear: I wasn't suggesting changing anything about CMake or snmalloc's use of it, I was just reacting to the next stanza in the docs, which describe `RelWithDebugInfo` as "with optimizations on but with debug info".  That doesn't make it clear whether it's Release-like in having shed assertions or Debug-like in having kept them.

PR #263, opened 2020-11-28
Fix ARM 32 bits build, prfm is a 64 bits instruction.

devnexen:


mjp41:
Do you know why the ARM 32-bit CI was not failing?  What should we change on that pipeline?

devnexen:
I do not really know azure pipelines, I do not know if the 32 bits build are genuine 32 bits or labels ?
Anyway I was just testing the build on a raspberry.

davidchisnall (approved):

davidchisnall:
@devnexen, what's the compiler that you're using?  I think we'll probably use `__builtin_prefetch()` with clang and only use this version on gcc Arm.  We might want to expand the CI to try gcc on the Arm builds.

devnexen:
ah right I did not notice but I was using gcc indeed :)

mjp41:
@devexen, thanks for fixing this.

PR #264, opened 2020-12-13
Sparc support proposal.

devnexen:
Also cmpxchg16gb support detection is based on the host but on
 a cross compile context it fails to apply properly.

### DIFF
--- a/src/aal/aal.h
+++ b/src/aal/aal.h
@@ -21,7 +21,6 @@
 
 #if defined(__powerpc__) || defined(__powerpc64__)
 #  define PLATFORM_IS_POWERPC
-#  define SNMALLOC_NO_AAL_BUILTINS
 #endif
 
 #if defined(__sparc__)
--- a/src/aal/aal.h
+++ b/src/aal/aal.h
@@ -21,6 +21,11 @@
 
 #if defined(__powerpc__) || defined(__powerpc64__)
 #  define PLATFORM_IS_POWERPC
+#  define SNMALLOC_NO_AAL_BUILTINS
+#endif
+
+#if defined(__sparc__)
+#  define PLATFORM_IS_SPARC
 #endif
 
 namespace snmalloc
@@ -54,6 +59,7 @@ namespace snmalloc
     PowerPC,
     X86,
     X86_SGX,
+    Sparc,
   };
 
   /**
@@ -147,6 +153,8 @@ namespace snmalloc
 #  include "aal_arm.h"
 #elif defined(PLATFORM_IS_POWERPC)
 #  include "aal_powerpc.h"
+#elif defined(PLATFORM_IS_SPARC)
+#  include "aal_sparc.h"
 #endif
 
 namespace snmalloc
--- a/src/aal/aal_sparc.h
+++ b/src/aal/aal_sparc.h
@@ -17,7 +17,7 @@ namespace snmalloc
     /**
      * Bitmap of AalFeature flags
      */
-    static constexpr uint64_t aal_features = StrictProvenance;
+    static constexpr uint64_t aal_features = IntegerPointers;
 
     static constexpr enum AalName aal_name = Sparc;
 
@@ -30,8 +30,6 @@ namespace snmalloc
     static constexpr size_t smallest_page_size = 0x1000;
 #endif
 
-    using address_t = uintptr_t;
-
     /**
      * On Sparc ideally pause instructions ought to be
      * optimised per Sparc processor but here a version
@@ -40,7 +38,7 @@ namespace snmalloc
      */
     static inline void pause()
     {
-      __asm__ volatile("rd %ccr, %g0 \n\trd %ccr, %g0 \n\trd %ccr, %g0");
+      __asm__ volatile("rd %%ccr, %%g0" ::: "memory");
     }
 
     static inline void prefetch(void* ptr)
@@ -48,15 +46,14 @@ namespace snmalloc
 #ifdef SNMALLOC_VA_BITS_64
       __asm__ volatile("prefetch [%0], 0" ::"r"(ptr));
 #else
-      (void)ptr;
+      UNUSED(ptr);
 #endif
     }
 
     static inline uint64_t tick()
     {
       uint64_t tick;
-      __asm__ volatile(".byte 0x83, 0x41, 0x00, 0x00");
-      __asm__ volatile("mov %%g1, %0" : "=r"(tick));
+      __asm__ volatile("rd %%asr4, %0" : "=r"(tick));
       return tick;
     }
   };
--- a/src/aal/aal_sparc.h
+++ b/src/aal/aal_sparc.h
@@ -1,0 +1,65 @@
+#pragma once
+
+#if defined(__arch64__) // More reliable than __sparc64__
+#  define SNMALLOC_VA_BITS_64
+#else
+#  define SNMALLOC_VA_BITS_32
+#endif
+
+namespace snmalloc
+{
+  /**
+   * Sparc architecture abstraction layer.
+   */
+  class AAL_Sparc
+  {
+  public:
+    /**
+     * Bitmap of AalFeature flags
+     */
+    static constexpr uint64_t aal_features = StrictProvenance;
+
+    static constexpr enum AalName aal_name = Sparc;
+
+#ifdef SNMALLOC_VA_BITS_64
+    /**
+     * Even Ultra-Sparc I supports 8192 and onwards
+     */
+    static constexpr size_t smallest_page_size = 0x2000;
+#else
+    static constexpr size_t smallest_page_size = 0x1000;
+#endif
+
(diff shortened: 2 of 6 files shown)


src/aal/aal_sparc.h
```diff
@@ -0,0 +1,65 @@
...
+#else
+      (void)ptr;
+#endif
+    }
+
+    static inline uint64_t tick()
+    {
+      uint64_t tick;
+      __asm__ volatile(".byte 0x83, 0x41, 0x00, 0x00");
+      __asm__ volatile("mov %%g1, %0" : "=r"(tick));
```
davidchisnall, line 59: Why is this written disassembled and why is it two asm blocks?  `asm volatile` may produce side effects and may not be discarded, but the compiler *is* free to move things between the blocks and so there's no guarantee that the two are adjacent and that %g1 is not clobbered between the two.  

The LLVM disassembler can disassemble this, so I'd assume that both gas and the LLVM assembler can handle assembling it.

```suggestion
      __asm__ volatile("rd %asr4, %0" : "=r"(tick)");
```

src/pal/pal_open_enclave.h
```diff
@@ -39,7 +39,8 @@ namespace snmalloc
      */
     static constexpr uint64_t pal_features = 0;
 
-    static constexpr size_t page_size = 0x1000;
+    static constexpr size_t page_size =
+      Aal::aal_name == Sparc ? Aal::smallest_page_size : 0x1000;
```
davidchisnall, line 43: I don't believe OpenEnclave supports SPARC, so this change looks a bit odd.

src/pal/pal_linux.h
```diff
@@ -26,7 +26,7 @@ namespace snmalloc
     static constexpr uint64_t pal_features = PALPOSIX::pal_features;
 
     static constexpr size_t page_size =
-      Aal::aal_name == PowerPC ? 0x10000 : 0x1000;
+      (Aal::aal_name == PowerPC ? 0x10000 : (Aal::aal_name == Sparc ? Aal::smallest_page_size : 0x1000));
```
davidchisnall, line 29: Does this actually need to be guarded on `Sparc`?  I think we should use the smallest page size or the larger value on PowerPC.

```suggestion
      (Aal::aal_name == PowerPC ? 0x10000 : Aal::smallest_page_size);
```

src/pal/pal_solaris.h
```diff
@@ -19,6 +19,8 @@ namespace snmalloc
      */
     static constexpr uint64_t pal_features = PALPOSIX::pal_features;
 
+    static constexpr size_t page_size =
+      Aal::aal_name == Sparc ? Aal::smallest_page_size : 0x1000;
```
davidchisnall, line 23: As above:

```suggestion
    static constexpr size_t page_size = Aal::smallest_page_size;
```

davidchisnall (commented):
Thanks, great to see SPARC support!

Comments inline.

src/aal/aal.h
```diff
@@ -21,6 +21,11 @@
 
 #if defined(__powerpc__) || defined(__powerpc64__)
 #  define PLATFORM_IS_POWERPC
+#  define SNMALLOC_NO_AAL_BUILTINS
```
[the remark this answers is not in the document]
devnexen: Ah right I forgot to remove it I wanted to try the assembly correctness ;-)

src/aal/aal_sparc.h
```diff
@@ -0,0 +1,65 @@
...
+  /**
+   * Sparc architecture abstraction layer.
+   */
+  class AAL_Sparc
+  {
+  public:
+    /**
+     * Bitmap of AalFeature flags
+     */
+    static constexpr uint64_t aal_features = StrictProvenance;
```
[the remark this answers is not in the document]
devnexen: Right I was thinking of strong alignment requirement for this platform.

src/pal/pal_open_enclave.h
devnexen: Right I was testing under qemu and led to this code path

src/aal/aal_sparc.h
devnexen: Right it looks simpler. In fact I stumbled across it in an old sparc book assembly but let s go with your version.

src/pal/pal_open_enclave.h
mjp41:
```suggestion
    static constexpr size_t page_size = Aal::smallest_page_size;
```
Does this work here?

src/pal/pal_solaris.h
mjp41: @devnexen  does David's suggestion not work?

mjp41 (changes requested):
Thanks for the PR.

I wonder if we could change 
https://github.com/microsoft/snmalloc/blob/975a2bd6db79adfc66b65ba11cffc83d5a10ba6c/src/pal/pal_posix.h#L101
to be:
```C++
```
Then 
https://github.com/microsoft/snmalloc/blob/975a2bd6db79adfc66b65ba11cffc83d5a10ba6c/src/pal/pal_linux.h#L28-L29
could become
```C++
    static constexpr size_t page_size =
      Aal::aal_name == PowerPC ? 0x10000 : PALPOSIX::page_size;
```
I don't believe you would need to specify it in the Solaris Pal, and then you could make the OE pal:
https://github.com/microsoft/snmalloc/blob/975a2bd6db79adfc66b65ba11cffc83d5a10ba6c/src/pal/pal_open_enclave.h#L42
to 
```C++
```
Then, I think there is less branching.

devnexen:
I appreciate the simplifications fair enough.

mjp41 (approved):

src/aal/aal_sparc.h
```diff
@@ -0,0 +1,62 @@
...
+      __asm__ volatile("prefetch [%0], 0" ::"r"(ptr));
+#else
+      UNUSED(ptr);
+#endif
+    }
+
+    static inline uint64_t tick()
+    {
+      uint64_t tick;
+      __asm__ volatile("rd %%asr4, %0" : "=r"(tick));
```
jrtc27, line 56: SPARC V8 did not define any standard ASRs, only Y/PSR/WIM/TBR, and had this to say about the existing implementations (L.3. Characteristics of Existing Implementations):

> None of the five implementations implements any user or privileged ancillary state registers.

It also said this in the description of RDASR:

> An rs1 value of 1...14 in an RDASR instruction produces undefined results, but does not cause an illegal_instruction trap.

Maybe later V8 implementations added ASR4, but it wasn't until V9 that ASR4 was standardised as TICK, and on V9 `rd %tick, %rd` is valid assembly.

Thus this needs to be guarded by a V9(/V8+, which is just V9 for 32-bit, as Solaris unfortunately conflated the ISA version with the ABI in use) check, and should then also be able to use `tick` rather than `asr4` in the assembly.

src/aal/aal_sparc.h
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * On Sparc ideally pause instructions ought to be
+     * optimised per Sparc processor but here a version
+     * as least common denominator to avoid numerous ifdef,
+     * reading Conditions Code Register here
+     */
+    static inline void pause()
+    {
+      __asm__ volatile("rd %%ccr, %%g0" ::: "memory");
```
jrtc27, line 41: It's probably worth at least mentioning here that as of OSA2011 there is a PAUSE ASR, intended to be used as:
```
wr %g0, reg_or_imm, %pause
```
where `reg_or_imm` is the number of cycles to pause for, and has:
```
pause reg_or_imm
```
as an assembly alias.

That's obviously not generally usable though for anything below v9e/v8pluse (E because it was the 5th revision to the V9 spec). What made you pick `rd %ccr` above anything else, out of interest? ASR 2 is not specified on V8 (see below) but maybe reading is required to not have any other architecturally-visible side-effects? The specification is not enormously clear.

src/aal/aal_sparc.h
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    static inline void prefetch(void* ptr)
+    {
+#ifdef SNMALLOC_VA_BITS_64
+      __asm__ volatile("prefetch [%0], 0" ::"r"(ptr));
```
jrtc27, line 47: Can use:
```C
      __asm__ volatile("prefetch %0, 0" :: "m"(*(const char*)ptr));
```
in order to support reg+reg and reg+imm addressing. Technically should be `*(const char (*)[])` (according to clang-format there is a space there even for `PointerAlignment: Left`, which makes sense I guess) to indicate it reads an unknown number of bytes, but given it's just a hint with no side-effects (nor actual value-dependence) it doesn't matter, and Clang doesn't seem to like using `[]` in that context, when GCC does and explicitly documents it as the way to express such constraints.

Probably also worth documenting that the 0 is for fun 0 meaning "Prefetch for several reads".

PR #265, opened 2020-12-14
Small fixes from living unusually

nwf:
Riding around on a POWER Linux box brings out corner cases.

### DIFF
--- a/src/mem/chunkmap.h
+++ b/src/mem/chunkmap.h
@@ -10,11 +10,11 @@ using namespace std;
 
 namespace snmalloc
 {
-  enum ChunkMapSuperslabKind
+  enum ChunkMapSuperslabKind : uint8_t
   {
     CMNotOurs = 0,
     CMSuperslab = 1,
-    CMMediumslab = 2
+    CMMediumslab = 2,
 
     /*
      * Values 3 (inclusive) through SUPERSLAB_BITS (exclusive) are as yet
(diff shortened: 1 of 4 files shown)


src/mem/chunkmap.h
```diff
@@ -47,12 +59,14 @@ namespace snmalloc
 // Use flat map is under a single node.
 #  define SNMALLOC_MAX_FLATPAGEMAP_SIZE PAGEMAP_NODE_SIZE
```
mjp41, line 60: Would it be better to change 
```
#  define SNMALLOC_MAX_FLATPAGEMAP_SIZE PAGEMAP_NODE_SIZE
```
to 
```
#  define SNMALLOC_MAX_FLATPAGEMAP_SIZE \
  pal_supports<LazyCommit> ? .... : PAGEMAP_NODE_SIZE
```
Where we choose `...` suitably.

nwf: With default chunk size (i.e., neither `USE_SMALL_CHUNKS` nor `USE_LARGE_CHUNKS`), `SUPERSLAB_BITS` is 20.  With a 48-bit VA, that means a flat chunkmap pagemap is 256 MiB.  Is that a reasonable choice for `...`?

mjp41: Probably, I would say going above 2^32 causes a lot of issues, so below that maybe 1GiB is the max?

nwf: Permitting 512M will break `src/test/func/memory/memory.cc`; we could adjust the test, tho'.

mjp41: Either is fine.  Just document the choice.

mjp41 (approved):
LGTM

PR #266, opened 2020-12-14
CHERI prep: stave off amplification for Remote routing

nwf:
Smuggle the size class in Remote objects, so that we can leave Remote pointers bounded to their underlying allocation and yet forward a Remote message using only local state.  That is, we need to avoid going to the {Alloc,Super,Medium,Meta}slab headers unless we're consuming the Remote message back into our free lists.

### DIFF
--- a/src/mem/alloc.h
+++ b/src/mem/alloc.h
@@ -867,44 +867,33 @@ namespace snmalloc
 
     SNMALLOC_FAST_PATH void handle_dealloc_remote(Remote* p)
     {
-      Superslab* super = Superslab::get(p);
+      if (likely(p->trunc_target_id() == get_trunc_id()))
+      {
+        // Destined for my slabs
+        Superslab* super = Superslab::get(p);
 
 #ifdef CHECK_CLIENT
-      if (p->trunc_target_id() != (super->get_allocator()->trunc_id()))
-        error("Detected memory corruption.  Potential use-after-free");
+        if (p->trunc_target_id() != (super->get_allocator()->trunc_id()))
+          error("Detected memory corruption.  Potential use-after-free");
 #endif
-      if (likely(super->get_kind() == Super))
-      {
-        if (likely(super->get_allocator() == public_state()))
-        {
-          small_dealloc_offseted(super, p, p->sizeclass());
-          return;
-        }
-      }
-      handle_dealloc_remote_slow(p);
-    }
+        // Guard against remote queues that have colliding IDs
+        SNMALLOC_ASSERT(super->get_allocator() == public_state());
 
-    SNMALLOC_SLOW_PATH void handle_dealloc_remote_slow(Remote* p)
-    {
-      Superslab* super = Superslab::get(p);
-      if (likely(super->get_kind() == Medium))
-      {
-        if (likely(super->get_allocator() == public_state()))
+        if (likely(p->sizeclass() < NUM_SMALL_CLASSES))
         {
-          void* start = remove_cache_friendly_offset(p, p->sizeclass());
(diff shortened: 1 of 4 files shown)


src/mem/remoteallocator.h
```diff
@@ -22,16 +22,34 @@ namespace snmalloc
...
+     * algorithm of the Alloc's RemoteCache already ignores the bottom
+     * "initial_shift" bits, which is, in practice, well above 8.  There's a
+     * static_assert() over there that helps ensure this stays true.
+     *
+     * This does mean that we might have message_queues that always collide in
+     * the hash algorithm, if they're within "initial_shift" of each other. Such
+     * pairings will substantially decrease performance and so we prohibit them
+     * and use SNMALLOC_ASSERT to verify that they do not exist in debug builds.
+     */
+    alloc_id_t alloc_id_and_sizeclass;
```
mjp41, line 39: Could you add
```C++
constexpr uintptr_t SIZECLASS_MASK = 0xFF;
```
and use that and its inverse for masking.  This removes the dependence on `ULL` being `uintptr_t`.

mjp41 (approved):
LGTM

src/mem/alloc.h
```diff
@@ -507,9 +507,9 @@ namespace snmalloc
 #endif
     }
 
-    size_t get_id()
+    size_t get_trunc_id()
```
davidchisnall, line 510: Doc comment?

src/mem/remoteallocator.h
```diff
@@ -22,16 +22,34 @@ namespace snmalloc
...
```
davidchisnall, line 38: Can this be a class type?  We have some helpers for bitfield extracts, so we could easily provide an abstract interface for getting the two bits.

src/mem/alloc.h
```diff
@@ -860,56 +848,39 @@ namespace snmalloc
...
+        SNMALLOC_ASSERT(likely(super->get_allocator() == public_state()));
 
-    SNMALLOC_SLOW_PATH void handle_dealloc_remote_slow(Remote* p)
-    {
-      Superslab* super = Superslab::get(p);
-      if (likely(super->get_kind() == Medium))
-      {
-        Mediumslab* slab = Mediumslab::get(p);
-        if (p->target_id() == id())
+        if (likely(super->get_kind() == Super))
```
mjp41, line 869: Can't you use the sizeclass here to check which path. 
```suggestion
if (p->sizeclass() < NUM_SMALL_SIZECLASSES)
```

davidchisnall (commented):
Code looks fine, but it could do with a few comments explaining what a 'truncated' ID is.

src/mem/alloc.h
nwf: Sure, that also works.  One fewer load?

src/mem/alloc.h
```diff
@@ -860,56 +848,39 @@ namespace snmalloc
...
-        if (likely(p->target_id() == id()))
-        {
-          small_dealloc_offseted(super, p, meta.sizeclass);
-          return;
-        }
-      }
-      handle_dealloc_remote_slow(p);
-    }
+        // Guard against remote queues that have colliding IDs
+        SNMALLOC_ASSERT(likely(super->get_allocator() == public_state()));
```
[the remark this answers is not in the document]
nwf: Copy-paste; dropped.

src/mem/remoteallocator.h
mjp41: So this is pretty contained in a single class already adding another layer of abstraction probably wouldn't buy us much.

PR #267, opened 2020-12-16
CHERI prep plumbing

nwf:
This is atop #265 and #266 and contains some other small changes and the next steps in preparation for CHERI / StrictProvenance tracking (with an eye towards MTE, too, but that's quite distant).  It is not yet finished -- I have a pile more changes on my desk to see it through -- but I am curious for CI's, and others', opinions.

To begin, read `docs/StrictProvenance.md` for an overview and some rationale for this proposal.

Then: are the aliases in `src/ds/ptrwrap.h` too ugly as they are?  They depend pretty heavily on inlining and SROA to avoid an indirection, and so I am quite open to suggestions to replace them with something better.

In terms of the AAL, there isn't much new, just exposing `CSetBounds` (`ptrauth_bound`) and `CSetAddress` (`ptrauth_rebound`), basically.  The AddressSpaceManager has grown a very coarse AuthMap and `ptrauth_amplify` method, which gets plumbed up to `Alloc`s via the `LargeAlloc`.

The changes to `Alloc` are just the beginning but should give a good hint as to where things are going.  I'll amend this with some more commits as I de-tangle the hideous mess I've made chasing types.

### DIFF
--- a/src/ds/address.h
+++ b/src/ds/address.h
@@ -1,6 +1,7 @@
 #pragma once
 #include "../pal/pal_consts.h"
 #include "bits.h"
+#include "ptrwrap.h"
 
 #include <cstdint>
 
@@ -24,6 +25,12 @@ namespace snmalloc
     return reinterpret_cast<T*>(reinterpret_cast<char*>(base) + diff);
   }
 
+  template<typename T>
+  inline AuthPtr<T> pointer_offset(AuthPtr<T> base, size_t diff)
+  {
+    return mk_authptr<T>(pointer_offset(base.unsafe_auth_ptr, diff));
+  }
+
   /**
    * Perform pointer arithmetic and return the adjusted pointer.
    */
@@ -40,6 +47,18 @@ namespace snmalloc
   inline address_t address_cast(T* ptr)
   {
     return reinterpret_cast<address_t>(ptr);
+  }
+
+  template<typename T>
+  inline address_t address_cast(AuthPtr<T> a)
+  {
+    return reinterpret_cast<address_t>(a.unsafe_auth_ptr);
+  }
+
+  template<typename T>
+  inline address_t address_cast(FreePtr<T> a)
+  {
+    return reinterpret_cast<address_t>(a.unsafe_free_ptr);
   }
 
   /**
@@ -98,6 +117,12 @@ namespace snmalloc
     }
   }
 
+  template<size_t alignment, typename T = void>
+  inline AuthPtr<T> pointer_align_up(AuthPtr<T> p)
+  {
+    return mk_authptr<T>(pointer_align_up<alignment, T>(p.unsafe_auth_ptr));
+  }
+
   /**
    * Align a pointer down to a dynamically specified granularity, which must be
    * a power of two.
@@ -130,6 +155,12 @@ namespace snmalloc
     return reinterpret_cast<T*>(
       bits::align_up(reinterpret_cast<uintptr_t>(p), alignment));
 #endif
+  }
+
+  template<typename T = void>
+  inline AuthPtr<T> pointer_align_up(AuthPtr<T> p, size_t alignment)
+  {
+    return mk_authptr<T>(pointer_align_up<T>(p.unsafe_auth_ptr, alignment));
   }
 
   /**
--- a/src/ds/ptrwrap.h
+++ b/src/ds/ptrwrap.h
@@ -1,0 +1,142 @@
+#pragma once
+
+namespace snmalloc
+{
+  /**
+   * A convenience type that we can use to annotate internal functions
+   * which return pointers that are headed out to the application or
+   * which have come back.
+   */
+  class ReturnPtr
+  {
+  public:
+    void* unsafe_return_ptr;
+
+    ReturnPtr(const std::nullptr_t n)
+    {
+      this->unsafe_return_ptr = n;
+    }
+
+    inline bool operator==(const ReturnPtr& rhs) const
+    {
+      return this->unsafe_return_ptr == rhs.unsafe_return_ptr;
+    }
+
+    inline bool operator!=(const ReturnPtr& rhs) const
+    {
+      return this->unsafe_return_ptr != rhs.unsafe_return_ptr;
+    }
+  };
+
+  /**
+   * A convenience type used to indicate a pointer suitable for inclusion on a
+   * free list.  The pointer itself is ready to be given to a client (that is,
+   * it has restricted authority on StrictProvenance architectures), but the
+   * contents of the memory are possibly unsafe to disclose.
+   *
+   * Obtain non-null `FreePtr`s using the AAL's `ptrauth_bound` method.
+   */
+  template<typename T>
+  class FreePtr
+  {
+  public:
+    T* unsafe_free_ptr;
+
+    FreePtr(const std::nullptr_t n) : unsafe_free_ptr(n) {}
+
+    template<typename U>
+    explicit FreePtr(FreePtr<U> fp)
+    : unsafe_free_ptr(reinterpret_cast<T*>(fp.unsafe_free_ptr))
+    {}
+
+    inline bool operator==(const FreePtr& rhs) const
+    {
+      return this->unsafe_free_ptr == rhs.unsafe_free_ptr;
+    }
+
+    inline bool operator!=(const FreePtr& rhs) const
+    {
+      return this->unsafe_free_ptr != rhs.unsafe_free_ptr;
+    }
+  };
+
+  /**
+   * A convenience type used to indicate that a pointer has elevated authority
+   * and requires some special care.
+   *
+   * Most non-NULL `AuthPtr`s in the Alloc-ator proper come from amplification;
+   * see the largealloc's `ptrauth_amplify` method.
+   */
+  template<typename T>
+  class AuthPtr
+  {
+  public:
+    T* unsafe_auth_ptr;
+
+    AuthPtr() : unsafe_auth_ptr(nullptr) {}
+
+    AuthPtr(const std::nullptr_t n) : unsafe_auth_ptr(n) {}
+
+    inline bool operator==(const AuthPtr& rhs) const
+    {
+      return this->unsafe_auth_ptr == rhs.unsafe_auth_ptr;
+    }
+
+    inline bool operator!=(const AuthPtr& rhs) const
+    {
+      return this->unsafe_auth_ptr != rhs.unsafe_auth_ptr;
+    }
+
+    inline bool operator<(const AuthPtr& rhs) const
+    {
+      return this->unsafe_auth_ptr < rhs.unsafe_auth_ptr;
+    }
+  };
+
+  /**
+   * While it's always safe to view a pointer as carrying its authority, we want
+   * to be explicit when we're doing so, so this isn't a constructor for the
+   * `AuthPtr` class, which would be an implicit conversion.
+   */
+  template<typename T = void>
+  SNMALLOC_FAST_PATH static AuthPtr<T> mk_authptr(void* p)
+  {
+    return *reinterpret_cast<AuthPtr<T>*>(&p);
+  }
+
+  /**
+   * Occasionally we want to treat an `AuthPtr` as a `FreePtr` with the
+   * knowledge that it isn't actually headed out to the user.
+   */
+  template<typename T, typename U = T>
+  SNMALLOC_FAST_PATH static FreePtr<T> unsafe_mk_freeptr(AuthPtr<U> p)
+  {
+    return *reinterpret_cast<FreePtr<T>*>(&p.unsafe_auth_ptr);
+  }
+
+  template<typename T>
+  SNMALLOC_FAST_PATH static FreePtr<T> unsafe_as_freeptr(ReturnPtr p)
+  {
+    return *reinterpret_cast<FreePtr<T>*>(&p.unsafe_return_ptr);
+  }
+
+  /**
+   * `ReturnPtr`s are `FreePtr`s that have had their contents sanitized; there
+   * is no great way to explain that to the type system, so carefully use this
+   * function after that sanitization.
+   */
+  template<typename T>
+  SNMALLOC_FAST_PATH static ReturnPtr unsafe_mk_returnptr(FreePtr<T> p)
+  {
+    return *reinterpret_cast<ReturnPtr*>(&p.unsafe_free_ptr);
+  }
+
+  /**
+   * `ReturnPtr`s are sometimes given back to us as `void *`.
+   */
+  SNMALLOC_FAST_PATH static ReturnPtr unsafe_as_returnptr(void* p)
+  {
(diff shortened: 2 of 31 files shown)


nwf:
9736b37 is just some fixes for CI; let's see how it feels now.

nwf:
CI fixes and a little more detangling of the wad of changes.

src/ds/address.h
```diff
@@ -98,6 +117,12 @@ namespace snmalloc
     }
   }
 
+  template<size_t alignment, typename T = void>
```
davidchisnall, line 120: Doc comment?

src/ds/address.h
```diff
@@ -132,6 +157,12 @@ namespace snmalloc
 #endif
   }
 
+  template<typename T = void>
```
davidchisnall, line 160: Doc comment?  Specifically, for these, explain why they are defined only on `AuthPtr` and not other pointer types.

Is there a reason that these are free-standing functions and not methods on `AuthPtr`?

src/ds/ptrwrap.h
```diff
@@ -0,0 +1,142 @@
...
+{
+  /**
+   * A convenience type that we can use to annotate internal functions
+   * which return pointers that are headed out to the application or
+   * which have come back.
+   */
+  class ReturnPtr
+  {
+  public:
+    void* unsafe_return_ptr;
```
davidchisnall, line 13: If there are no private members, this should be a `struct`.  That said, can't `unsafe_return_ptr` be a `private` member, with the small handful of functions that are actually allowed to use this declared as `friend`s?  At the moment, if someone accesses the raw pointer value we don't get any kind of warning.

src/ds/ptrwrap.h
```diff
@@ -0,0 +1,142 @@
...
+
+    ReturnPtr(const std::nullptr_t n)
```
davidchisnall, line 15: Don't we need a constructor from `AuthPtr`?

src/ds/ptrwrap.h
davidchisnall: Same question for the two other smart pointers.

src/ds/ptrwrap.h
```diff
@@ -0,0 +1,142 @@
...
+    inline bool operator<(const AuthPtr& rhs) const
+    {
+      return this->unsafe_auth_ptr < rhs.unsafe_auth_ptr;
+    }
+  };
+
+  /**
+   * While it's always safe to view a pointer as carrying its authority, we want
+   * to be explicit when we're doing so, so this isn't a constructor for the
+   * `AuthPtr` class, which would be an implicit conversion.
```
davidchisnall, line 99: Why not just make it an `explicit` constructor then?

src/mem/address_space.h
```diff
@@ -1,10 +1,96 @@
...
 {
+  template<
+    SNMALLOC_CONCEPT(ConceptPAL) PAL,
+    SNMALLOC_CONCEPT(ConceptAAL) AAL,
+    typename PrimAlloc>
+  class AuthMap
+  {
+    static_assert(
+      !aal_supports<StrictProvenance> || pal_supports<AlignedAllocation, PAL>,
+      "StrictProvenance requires platform support for aligned allocation");
```
davidchisnall, line 17: Please add a comment explaining why.

src/mem/address_space.h
```diff
@@ -1,10 +1,96 @@
...
+      default_alloc_size_t>::ptrauth_root_alloc_size;
+
+    static_assert(
+      alloc_size % PAL::page_size == 0,
+      "Provenance root granule size must be a multiple of page size");
+
+    static_assert(
+      !aal_supports<StrictProvenance, AAL> ||
+        (bits::next_pow2_const(alloc_size) == alloc_size),
+      "Provenace root granule size must be a power of two");
```
davidchisnall, line 44: Comment explaining why please.

src/mem/alloc.h
```diff
@@ -330,15 +339,14 @@ namespace snmalloc
         return;
       }
       dealloc_sized_slow(p_auth, p_free, size);
-#endif
     }
 
     SNMALLOC_SLOW_PATH
     void
     dealloc_sized_slow(AuthPtr<void> p_auth, FreePtr<void> p_free, size_t size)
```
mjp41, line 346: Do these register pressure concerns affect the codegen without StrictProvenance? `dealloc_sized_slow` will not be inlined, so the calling of them even in tail calls may affect the codegen for the fast path.

src/mem/address_space.h
```diff
@@ -1,10 +1,96 @@
...
+    {
+      template<typename T>
+      static SNMALLOC_FAST_PATH AuthPtr<T> get(address_t a)
+      {
+        UNUSED(a);
+        return nullptr;
+      }
+    };
+
+    using AuthPagemap = std::conditional_t<
```
davidchisnall, line 63: Please add some comments for things like this - it's a complex expression that it's very easy to misread!

src/mem/address_space.h
```diff
@@ -13,7 +99,7 @@ namespace snmalloc
    * It cannot unreserve memory, so this does not require the
    * usual complexity of a buddy allocator.
    */
-  template<SNMALLOC_CONCEPT(ConceptPAL) PAL>
+  template<SNMALLOC_CONCEPT(ConceptPAL) PAL, typename PrimAlloc>
```
davidchisnall, line 102: Please extend the doc comment to explain what `PrimAlloc` is.

mjp41 (commented):
Overall, this is a really nice refactor that very naturally adds Provenance to snmalloc.  I think I have two concerns,
* Have any of the changes affected the codegen of fast paths without strict provenance.  You are passing more variable, so this can affect the calling convention, and thus might not be optimised away even when not used.
* FreePtr is being put into the space of allocations, but this appears to be unbounded?  This seems like it is very likely to be a place for exploit if we don't have full temporal safety,

src/mem/chunkmap.h
```diff
@@ -172,22 +172,12 @@ namespace snmalloc
...
-    /**
-     * Get the pagemap entry corresponding to a specific address.
+     * Get the pagemap entry corresponding to a specific address inside
+     * a ReturnPtr wrapper.
      */
-    static uint8_t get(void* p)
+    static uint8_t get(ReturnPtr p)
     {
-      return get(address_cast(p));
+      return get(address_cast(p.unsafe_return_ptr));
```
davidchisnall, line 180: I think this might be cleaner if `ReturnPtr` defined an `address_cast` member, then we're not requiring access to the raw pointer.

src/mem/mediumslab.h
```diff
@@ -22,7 +22,7 @@ namespace snmalloc
     uint16_t free;
     uint8_t head;
     uint8_t sizeclass;
-    uint16_t stack[SLAB_COUNT - 1];
+    FreePtr<void> stack[SLAB_COUNT - 1];
```
davidchisnall, line 25: This looks as if it's a big increase in the metadata size for a medium slab (especially on CHERI, where each entry goes from 2 to 16 bytes).  Is it necessary?

mjp41: Good catch, this would push the meta-data size over a page.  I would be careful that there aren't corner cases that are bad here.  The increase is probably not going to affect much in terms of performance or memory usage.

src/mem/alloc.h
```diff
@@ -623,6 +623,19 @@ namespace snmalloc
...
             {
               // Send all slots to the target at the head of the list.
+              /*
+               * This is somewhat dubious: this chunk might be either a
+               * Superslab or a Mediumslab, but we access only the
+               * get_allocator() method of their common parent class,
+               * Allocslab, which reads the allocator field, which should
+               * have common offset.
+               */
+#ifdef __clang__
```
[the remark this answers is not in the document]
nwf: An `AllocSlab::get()` would certainly make sense and at this particular site, we don't need any downcasting.  I'd still want to keep the assert, tho', because while I can't really imagine a C++ compiler deciding not to lay out class members for a single inheritance chain as concatenated structs, I'm sure someone will find some epsilon performance improvement by considering something else.

Non-clang objected that `Superslab` and `Mediumslab` were not "standard layout classes" thanks to their inheritance.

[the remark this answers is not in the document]
mjp41: Oh, so you believe it is possible for upcast actually to change the offset, i.e. 
```C++
   Superslab* p; 
   Allocslab* a = reinterpret_cast<AllocSlab*>(p);
   assert (reinterpret_cast<void*>(p) == reinterpret_cast<void*>(a)); 
```
the assertion could fail?

src/mem/mediumslab.h
nwf: Ah, good catch re: crossing a page boundary.  I had converted these to `FreePtr<void>`s so that we didn't have to `ptrauth_bound` on the allocation path and so that, when the time comes, the version tags would be available implicitly... but
* the `CSetBounds` is probably not very expensive and
* it's might be worth instead having `uint16_t stack[SLAB_COUNT-1]` and `uint8_t stack_memver[SLAB_COUNT-1]`; this will necessitate further AAL methods for memory versioning, but that's probably fine.

If we do decide that `CSetBounds`+`CSetVersion` on the alloc path and `CGetVersion` on free is collectively too pricey, I wonder if it might be worth having a pair of SLAB_COUNT/2-sized stacks, one of which lives on a free page.  That'd keep everything small enough and shouldn't add much code complexity.  But that's a discussion for another PR, really.

src/ds/address.h
nwf: It's not a *good* reason, but, as to why they're freestanding: the AAL defines `address_t` and uses the `*Ptr` wrapping types in its own methods.  So far as I know, I can't declare a member of a class with a "trust me, there'll be a type for you by the time you care" signature.

src/ds/ptrwrap.h
nwf: No, by design: we're trying to ensure that we go through `ptrauth_bound`, which takes an `AuthPtr` to a `FreePtr`, and then `unsafe_mk_returnptr` to convert the `FreePtr` to `ReturnPtr`.  The intention is that most allocation paths actually only need the latter half, as they will have obtained a `FreePtr` directly from a fast free list or such, using any `AuthPtr`s they have just to get to allocator metadata (e.g., using `Superslab::get(AuthPtr)` but then reading out a `FreePtr`).

src/ds/ptrwrap.h
```diff
@@ -0,0 +1,142 @@
...
+  {
+    return *reinterpret_cast<AuthPtr<T>*>(&p);
+  }
+
+  /**
+   * Occasionally we want to treat an `AuthPtr` as a `FreePtr` with the
+   * knowledge that it isn't actually headed out to the user.
+   */
+  template<typename T, typename U = T>
+  SNMALLOC_FAST_PATH static FreePtr<T> unsafe_mk_freeptr(AuthPtr<U> p)
```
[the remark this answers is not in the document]
nwf: Indeed.  Ideally, this (or this kind of cast) would be used only in the non-`StrictProvenance` case as just type-level trickery around a no-op.

I'll go sprinkle commentary at each use site about why I think it's justified there.

src/mem/alloc.h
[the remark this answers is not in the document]
nwf: I... I am just a caveman, and C++ terrifies me, as the line goes, and `reinterpret_cast` does so more than most things in this language.  Having dug around quite a bit in C++20, though, I think that that particular assertion cannot fail:

* `reinterpret_cast<Allocslab*>(p)` is defined (7.6.1.9 clause 7) to be `static_cast<Allocslab*>(static_cast<void*>(p))`.
* `static_cast<void*>(p)` is defined (7.6.1.8 clause 4; 12.4.3.1; 7.3.11 clause 2) to have the same pointer value as `p`.
* `static_cast<Allocslab*>(...)` is also defined to leave the pointer value alone (7.6.1.8 clause 13) because `Superslab` and `Allocslab` objects are not pointer interconvertable (6.8.2).
* `reinterpret_cast<void*>(x)` is `static_cast<void*>(static_cast<void*>(x))` and this leaves the pointer value unchanged by the above and the observation that the outer cast is satisfied by an empty implicit conversion sequence.

Unfortunately, I think the case at hand is subtly different: those conversions are defined to leave the pointer value unchanged but are not necessarily defined to result in sensible pointers.  In particular, I think there is risk that, despite its type, `a` is not actually left pointing at an `Allocslab`, due to the trip through `void*` induced by the `reinterpret_cast`.  Consider, instead:

```c++
Superslab *s;
Allocslab *sa = static_cast<Allocslab*>(sa);
Mediumslab *m = reinterpret_cast<Mediumslab*>(s); // same value as s, as above
Allocslab *ma = static_cast<Allocslab*>(ma);
assert (reinterpret_cast<void*>(sa) == reinterpret_cast<void*>(ma));
```

Here, the `static_cast`s are governed by (I think, eventually) 7.3.11 clause **3**, which lacks any note that the pointer value is unchanged, even if the base class is the first or sole base class.  Pointer interconvertability would save us only if our classes were standard-layout, but they're not, because `Baseslab` has mixed access control (as did `Allocslab` prior to this PR; 11.2 clauses 3.3 and 3.4).  Again, I haven't the foggiest why anyone would actually *use* this seemingly permitted freedom in the standard, but here we are.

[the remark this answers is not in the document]
mjp41: I'm happy for you to leave the `static_assert` in place.  But I think guarding it for clang is negating its utility. It is going to probably be some other C++ compiler that causes the issue, I am pretty certain `offsetof` is in the standard now, so this should be valid on any C++ compiler.

src/ds/address.h
davidchisnall: Please add that to a comment.

src/ds/ptrwrap.h
davidchisnall: It feels like the idiomatic way of doing that would be to remove `unsafe_mk_freeptr`, add a private constructor to `FreePtr` that takes an `AuthPtr` and make the few functions that should call it `friend`s of `FreePtr`.

src/mem/alloc.h
```diff
@@ -473,13 +473,13 @@ namespace snmalloc
...
 
   public:
-    SNMALLOC_FAST_PATH static size_t alloc_size(const void* p)
+    SNMALLOC_FAST_PATH size_t alloc_size(const void* p)
     {
 #ifdef SNMALLOC_PASS_THROUGH
       return external_alloc::malloc_usable_size(const_cast<void*>(p));
 #else
       // This must be called on an external pointer.
-      size_t size = ChunkMap::get(address_cast(p));
```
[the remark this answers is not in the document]
nwf: I, very belatedly, took some time to look at this in more detail.  Clang's able, if both `alloc_size` and `external_pointer` are de-`static`-ed, to generate almost the same code, differing just in register selection, for the `perf/external_pointer` test.  (Oddly, changing just one makes it perform worse, but since we're not doing that...)

PR #269, opened 2021-01-07
Fix the sandbox use case and add a test.

davidchisnall:
Summary of changes:

- Add a new PAL that doesn't allocate memory, which can be used with a
  memory provider that is pre-initialised with a range of memory.
- Slightly refactor the memory provider class so that it has a narrower
  interface with LargeAlloc and is easier to proxy.
- Allow the address space manager and the memory provider to be
  initialised with a range of memory.

This may eventually also remove the need for (or, at least, simplify)
the Open Enclave PAL.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,13 +21,13 @@ option(SNMALLOC_OPTIMISE_FOR_CURRENT_MACHINE "Compile for current machine archit
 set(CACHE_FRIENDLY_OFFSET OFF CACHE STRING "Base offset to place linked-list nodes.")
 set(SNMALLOC_STATIC_LIBRARY_PREFIX "sn_" CACHE STRING "Static library function prefix")
 
-# malloc.h was removed from C in C89 and so FreeBSD's version has a #error and
-# tells you to include stdlib.h instead.  Fortunately, all of the non-portable
-# extensions to malloc are in malloc_np.h on FreeBSD so we can use that to
-# check if we should include malloc.h.
+# malloc.h will error if you include it on FreeBSD, so this test must not
+# unconditionally include it.
 CHECK_C_SOURCE_COMPILES("
 #if __has_include(<malloc_np.h>)
 #include <malloc_np.h>
+#if __has_include(<malloc/malloc.h>)
+#include <malloc/malloc.h>
 #else
 #include <malloc.h>
 #endif
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,8 +21,16 @@ option(SNMALLOC_OPTIMISE_FOR_CURRENT_MACHINE "Compile for current machine archit
 set(CACHE_FRIENDLY_OFFSET OFF CACHE STRING "Base offset to place linked-list nodes.")
 set(SNMALLOC_STATIC_LIBRARY_PREFIX "sn_" CACHE STRING "Static library function prefix")
 
+# malloc.h was removed from C in C89 and so FreeBSD's version has a #error and
+# tells you to include stdlib.h instead.  Fortunately, all of the non-portable
+# extensions to malloc are in malloc_np.h on FreeBSD so we can use that to
+# check if we should include malloc.h.
 CHECK_C_SOURCE_COMPILES("
+#if __has_include(<malloc_np.h>)
+#include <malloc_np.h>
+#else
 #include <malloc.h>
+#endif
 size_t malloc_usable_size(const void* ptr) { return 0; }
 int main() { return 0; }
 " CONST_QUALIFIED_MALLOC_USABLE_SIZE)
@@ -306,6 +314,9 @@ if(NOT DEFINED SNMALLOC_ONLY_HEADER_LIBRARY)
         endif()
         if (${SUPER_SLAB_SIZE} STREQUAL "malloc")
           target_compile_definitions(${TESTNAME} PRIVATE SNMALLOC_PASS_THROUGH)
+        endif()
+        if(CONST_QUALIFIED_MALLOC_USABLE_SIZE)
+          target_compile_definitions(${TESTNAME} PRIVATE -DMALLOC_USABLE_SIZE_QUALIFIER=const)
         endif()
         target_link_libraries(${TESTNAME} snmalloc_lib)
         if (${TEST} MATCHES "release-.*")
--- a/src/mem/address_space.h
+++ b/src/mem/address_space.h
@@ -191,14 +191,14 @@ namespace snmalloc
         if (res == nullptr)
         {
           // Allocation failed ask OS for more memory
-          void* block;
-          size_t block_size;
+          void* block = nullptr;
+          size_t block_size = 0;
           if constexpr (pal_supports<AlignedAllocation, PAL>)
           {
             block_size = PAL::minimum_alloc_size;
             block = PAL::template reserve_aligned<false>(block_size);
           }
-          else
+          else if constexpr (!pal_supports<NoAllocation, PAL>)
           {
             // Need at least 2 times the space to guarantee alignment.
             // Hold lock here as a race could cause additional requests to
--- a/src/mem/address_space.h
+++ b/src/mem/address_space.h
@@ -236,5 +236,21 @@ namespace snmalloc
 
       return res;
     }
+
+    /**
+     * Default constructor.  An address-space manager constructed in this way
+     * does not own any memory at the start and will request any that it needs
+     * from the PAL.
+     */
+    AddressSpaceManager() = default;
+
+    /**
+     * Constructor that pre-initialises the address-space manager with a region
+     * of memory.
+     */
+    AddressSpaceManager(void* base, size_t length)
+    {
+      add_range(base, length);
+    }
   };
 } // namespace snmalloc
--- a/src/mem/largealloc.h
+++ b/src/mem/largealloc.h
@@ -75,9 +75,6 @@ namespace snmalloc
      */
     std::atomic<size_t> peak_memory_used_bytes{0};
 
-  public:
-    using Pal = PAL;
-
     /**
      * Memory current available in large_stacks
      */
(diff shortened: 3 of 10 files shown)


davidchisnall:
This is neccessary for fixing the Verona process isolation code.

src/mem/address_space.h
```diff
@@ -236,5 +236,21 @@ namespace snmalloc
 
       return res;
     }
+
+    /**
+     * Default constructor.  An address-space manager constructed in this way
+     * does not own any memory at the start and will request any that it needs
+     * from the PAL.
+     */
+    AddressSpaceManager() = default;
```
rengolin, line 245: What's the use case for this constructor? How can the AS manager manage the memory if it doesn't add a range?

The `Sandbox` class below reserves the memory in the constructor, I was expecting it to be passed down to the AS manager on its own construction.

src/mem/largealloc.h
```diff
@@ -88,6 +85,51 @@ namespace snmalloc
...
+     */
+    SNMALLOC_FAST_PATH void* pop_large_stack(size_t large_class)
+    {
+      void* p = large_stack[large_class].pop();
+      if (p != nullptr)
+      {
+        const size_t rsize = bits::one_at_bit(SUPERSLAB_BITS) << large_class;
+        available_large_chunks_in_bytes -= rsize;
+      }
+      return 0;
```
rengolin, line 104: Shouldn't this return `p`?

rengolin (commented):
I'm not entirely familiar with snmalloc or its implementation, so my comments could be misleading. Grain of salt and all.

src/mem/address_space.h
davidchisnall: Precisely as the comment says: it requests memory from the PAL.  This is how it works normally in the non-sandbox use case.

src/mem/largealloc.h
davidchisnall: Yes.  I honestly have no idea how this passes any tests as written.

src/mem/address_space.h
rengolin: Right, then why do you also need the second c-tor, with pointer and size?

davidchisnall: For the sandbox use case, where the PAL is not able to allocate any memory and all of the memory that it has access to is provided in the constructor.

src/pal/pal_noalloc.h
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    /**
+     * Reserve memory.
+     *
+     * This should never be called.
+     */
+    static std::pair<void*, size_t> reserve_at_least(size_t) noexcept
+    {
+      error("Out of memory");
```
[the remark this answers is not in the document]
mjp41: I like that idea of a faeture flag for this.

src/test/func/sandbox/sandbox.cc
```diff
@@ -0,0 +1,234 @@
...
+        ptr >= pointer_offset(start, sizeof(SharedState)) &&
+        (pointer_offset(ptr, sz) < top));
+    }
+
+  private:
+    void* alloc_sandbox_heap(size_t sb_size)
+    {
+      if constexpr (pal_supports<AlignedAllocation, DefaultPal>)
+      {
+        return DefaultPal::reserve_aligned<true>(sb_size);
```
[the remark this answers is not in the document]
davidchisnall: In theory, yes, but for this test it will be - I don't think we're going to have a PAL that can't allocate 128MiB.

src/pal/pal_noalloc.h
```diff
@@ -0,0 +1,83 @@
...
+  template<typename ErrorHandler>
+  struct PALNoAlloc
+  {
+    /**
+     * Bitmap of PalFeatures flags indicating the optional features that this
+     * PAL supports.
+     */
+    static constexpr uint64_t pal_features = 0;
+
+    static constexpr size_t page_size = ErrorHandler::page_size;
```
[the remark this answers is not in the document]
davidchisnall: I didn't want to call it a Pal because we don't forward any of the memory ops to it.

src/pal/pal_noalloc.h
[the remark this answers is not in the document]
davidchisnall: The ASM still has this code path, it just shouldn't be reached.  Currently, snmalloc doesn't handle allocation failures at all, this is how the other PALs handle allocation failure.

src/mem/address_space.h
mjp41: Could the NO_ALLOCATION Pal feature be used to disable this constructor?

CMakeLists.txt
```diff
@@ -21,8 +21,16 @@ option(SNMALLOC_OPTIMISE_FOR_CURRENT_MACHINE "Compile for current machine archit
 set(CACHE_FRIENDLY_OFFSET OFF CACHE STRING "Base offset to place linked-list nodes.")
 set(SNMALLOC_STATIC_LIBRARY_PREFIX "sn_" CACHE STRING "Static library function prefix")
 
+# malloc.h was removed from C in C89 and so FreeBSD's version has a #error and
```
mjp41, line 24: Not sure the history lesson is necessary.  Non-standard features aren't in the standard, so putting them in "malloc.h" seems perfectly reasonable.

src/pal/pal_noalloc.h
```diff
@@ -0,0 +1,83 @@
...
+    /**
+     * OS specific function for zeroing memory.
+     *
+     * This just calls bzero - we don't assume that we have access to any
+     * virtual-memory functions.
+     */
+    template<bool page_aligned = false>
+    static void zero(void* p, size_t size) noexcept
+    {
+      bzero(p, size);
```
mjp41, line 70: This would need to be overridable/mixinable for OE to be able to reuse this.  It would be nice to not increase the code size unnecessarily.

src/test/func/sandbox/sandbox.cc
```diff
@@ -0,0 +1,234 @@
...
+      }
+      else
+      {
+        return DefaultPal::reserve_at_least(sb_size).first;
```
mjp41, line 199: This is fine for the test, but it is wasting address space.  Add a comment about ignoring second, so if anyone copies this code, they understand what has happened here.

src/test/func/sandbox/sandbox.cc
```diff
@@ -0,0 +1,234 @@
...
+      shared_state(new (start) SharedState()),
+      state(
+        pointer_offset(start, sizeof(SharedState)),
+        sb_size - sizeof(SharedState)),
+      alloc(state, SNMALLOC_DEFAULT_CHUNKMAP(), &shared_state->queue)
+    {
+      auto* state_proxy = static_cast<MemoryProviderProxy*>(
+        alloc.alloc(sizeof(MemoryProviderProxy)));
+      state_proxy->real_state = &state;
+      internal_alloc =
```
mjp41, line 166: I am not sure about directly allocating the allocator, rather than creating a pool of allocators, which this allocator is part of.  There are some operations that work over the set of allocators.  Mostly for debug, but directly allocating will mean that is not something we can do in the future.

src/mem/largealloc.h
mjp41: You probably would only detect this if you are tracking address space usage, this will not recycle any chunks (i.e. 1MiB ranges) of memory,  So tests that go up and down in usage a lot will leak.  Did it pass CI.  I would expect a Windows test to fail with OOM with this bug.  But on your own machine, probably none of the tests will come close to an issue.

src/test/func/sandbox/sandbox.cc
```diff
@@ -0,0 +1,234 @@
...
+{
+  static const size_t sb_size = 128 * 1024 * 1024;
+
+  // Check that we can create two sandboxes
+  Sandbox sb1(sb_size);
+  Sandbox sb2(sb_size);
+
+  auto check = [](Sandbox& sb, auto& alloc, size_t sz) {
+    void* ptr = alloc.alloc(sz);
+    assert(sb.is_in_sandbox_heap(ptr, sz));
```
mjp41, line 215: SNMALLOC_CHECK rather than `assert`.  

FYI.  We use our own `assert` as it is not defined in OE.

src/mem/address_space.h
davidchisnall: Not very easily.  This ends up deleting a constructor that's used in a load of other places and propagates up into LargeAlloc and its `make` function.

src/pal/pal_noalloc.h
davidchisnall: Or the OE version could just implement `snmalloc::bzero` as a wrapper around `oe_memset_s`.

src/pal/pal_noalloc.h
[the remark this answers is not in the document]
davidchisnall: I've added a `NoAllocation` flag and guarded access to the reserve-family functions on it.  This function is now gone.

src/pal/pal_noalloc.h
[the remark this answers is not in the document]
davidchisnall: I've modified this to use the Aal's one, so now this just needs to implement the two error-related things.

src/test/func/sandbox/sandbox.cc
davidchisnall: In the real implementation, we'll set up all of the alloc pool machinery inside the sandbox, this test is just checking that we can construct both kinds of allocator.  I've added a comment to this effect.

Note that in an environment with multiple trust domains we will never be able to enumerate *all* allocators.

src/test/func/sandbox/sandbox.cc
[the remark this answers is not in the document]
davidchisnall: Not all PALs provide `minimum_alloc_size`.  We should probably add that to the concept...

src/mem/address_space.h
rengolin: I see, one for the external alloc and one for the internal alloc. Makes sense, thanks!

rengolin:
In [GCC 8 test](https://dev.azure.com/snmalloc/snmalloc/_build/results?buildId=1238&view=logs&j=6bf0e35d-1288-5ee7-ea08-282b8034ee38&t=571c5d16-f275-579a-504f-3e0ea384f09b&l=180):

```
../src/mem/allocstats.h:199:22: error: array subscript 18446744073709551601 is above array bounds of 'size_t [28]' {aka 'long unsigned int [28]'} [-Werror=array-bounds]
       large_pop_count[sc]++;
```

Uninitialised?

src/pal/pal_noalloc.h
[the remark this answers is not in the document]
mjp41: >The ASM still has this code path, it just shouldn't be reached. Currently, snmalloc doesn't handle allocation failures at all, this is how the other PALs handle allocation failure.

Returning `nullptr` should work for allocation failures.  We have tests that check for OOM as that is requried behaviour on OE.  If there are code paths that cannot handle a null return from the Pal, then that is a problem.

CMakeLists.txt
mjp41: I am happy with the test fix, just not the history lesson in the comments.  The standard does not require the compiler to fail if you include "malloc.h" in C89 or later. I think FreeBSDs choice is a perfectly valid one, but it doesn't need documenting in detail here.

While fixing you might want to fix Apple as well, which uses
```C++
#include <malloc/malloc.h>
```
The file [external_alloc.h](https://github.com/microsoft/snmalloc/blob/master/src/mem/external_alloc.h) goes through various ways of including `malloc_usable_size`.

src/test/func/sandbox/sandbox.cc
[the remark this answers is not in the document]
nwf: AFAIK `minimum_alloc_size` is required by AlignedAllocation PALs.  Sorry to have left it out of the concept; I can catch that in my next round of miscellany if you don't want to do it here.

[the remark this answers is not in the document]
mjp41: Yep, it uses it here:
https://github.com/microsoft/snmalloc/blob/4837c8248912f3dda3df84e1b1345bcb4c29d2ab/src/mem/address_space.h#L181-L185

CMakeLists.txt
davidchisnall: I don't really want to lose the comment because, until we get FreeBSD in CI, someone is likely to look at the code say 'oh, you can simplify this by unconditionally including `<malloc.h>`' or 'We should just include `<malloc.h>` if it exists and see that CI is happy.

mjp41:
```c++
# `malloc_usable_size` is an extension to the C runtime.
# This test code needs to find the relevant header. 
# Below we list the platforms and the relevant header: 
#   Linux            malloc.h
#   FreeBSD       malloc_np.h
#   Apple           malloc/malloc.h
```
We don't need to say why a particular platform made their choice the way they did.  It does not affect `snmalloc` functionality.

mjp41: I realised Apple has `malloc_size` so what I said there is incorrect.  

We could add a comment like:
```
# Including malloc.h is an error on FreeBSD.
```

mjp41: Thank you for the change the new comment states clearly what an snmalloc dev needs to know.

davidchisnall:
CI is happy!  Thus ends today's game of whack-a-mole.

nwf (approved):

mjp41 (approved):
LGTM, thanks for all the changes.

PR #270, opened 2021-01-08
Slab metadata vs. the C++ standard

nwf:
Inheritance with non-static members is not considered Standard Layout, and so
the compiler is nominally free to shuffle things around in all kinds of unusual
ways.  Instead of inheritance, rely on pointer interconvertibility with first
members of classes.

Document what's going on with static_asserts and type_traits.  The actual test
we want, std::is_pointer_interconvertible_with_class(Class::member), is not
defined until C++20 and is at least not yet in Clang.  Gate it behind the
appropriate feature test macro.

-----

See also the conversation at https://github.com/microsoft/snmalloc/pull/267#discussion_r552456019 .  I could go either way on this being included or not, and it's certainly severable from the rest of the CHERI work, so here 'tis as a separate PR.

Asking clang its opinions of the structure layouts before and after yields no significant difference; only the names of things change:
* Before:
```
         0 | class snmalloc::Mediumslab
         0 |   class snmalloc::Allocslab (base)
         0 |     class snmalloc::Baseslab (base)
         0 |       enum snmalloc::SlabKind kind
         8 |     struct snmalloc::RemoteAllocator * allocator
        64 |   class snmalloc::Mediumslab * next
        72 |   class snmalloc::Mediumslab * prev
        80 |   uint16_t free
        82 |   uint8_t head
        83 |   uint8_t sizeclass
        84 |   uint16_t [63] stack
           | [sizeof=256, dsize=210, align=64,
           |  nvsize=210, nvalign=64]

         0 | class snmalloc::Superslab
         0 |   class snmalloc::Allocslab (base)
         0 |     class snmalloc::Baseslab (base)
         0 |       enum snmalloc::SlabKind kind
         8 |     struct snmalloc::RemoteAllocator * allocator
        64 |   class snmalloc::Superslab * next
        72 |   class snmalloc::Superslab * prev
        80 |   class snmalloc::Mod<64, unsigned char> head
        80 |     unsigned char value
        82 |   uint16_t used
        88 |   class snmalloc::ModArray<64, class snmalloc::Metaslab> meta
        88 |     class snmalloc::Metaslab [64] array
           | [sizeof=1152, dsize=1112, align=64,
           |  nvsize=1112, nvalign=64]
```
* After:
```
         0 | class snmalloc::Mediumslab
         0 |   class snmalloc::Allocslab allocslab
         0 |     class snmalloc::Baseslab base
         0 |       enum snmalloc::SlabKind kind
         8 |     struct snmalloc::RemoteAllocator * allocator
        64 |   class snmalloc::Mediumslab * next
        72 |   class snmalloc::Mediumslab * prev
        80 |   uint16_t free
        82 |   uint8_t head
        83 |   uint8_t sizeclass
        84 |   uint16_t [63] stack
           | [sizeof=256, dsize=210, align=64,
           |  nvsize=210, nvalign=64]

         0 | class snmalloc::Superslab
         0 |   class snmalloc::Allocslab allocslab
         0 |     class snmalloc::Baseslab base
         0 |       enum snmalloc::SlabKind kind
         8 |     struct snmalloc::RemoteAllocator * allocator
        64 |   class snmalloc::Superslab * next
        72 |   class snmalloc::Superslab * prev
        80 |   class snmalloc::Mod<64, unsigned char> head
        80 |     unsigned char value
        82 |   uint16_t used
        88 |   class snmalloc::ModArray<64, class snmalloc::Metaslab> meta
        88 |     class snmalloc::Metaslab [64] array
           | [sizeof=1152, dsize=1112, align=64,
           |  nvsize=1112, nvalign=64]
```


mjp41:
So I am not sure about this still.  I find it hard to believe that a compiler could layout out the classes differently and not break everything.  By Standard layout do you mean

https://en.cppreference.com/w/cpp/named_req/StandardLayoutType

as this says

> All non-static data members have the same access control

and

> Has no non-static data members of reference type

So a lot of sensible code would be non-standard layout even without inheritance. Is this the thing you are worrying about.

nwf:
Yes, that's what I mean.  I, too, am quite surprised by the degrees of freedom the C++ standard permits here; presumably early C++ compilers did many things and the standards committee wants all of them to still be within spec.

PR #271, opened 2021-01-13
Small tweaks to make the sandbox mode happy.

davidchisnall:
Allow replacing the type of GlobalVirtual at compile time.
Make that type a friend of Pool so that it can allocate things.


mjp41 (approved):
LGTM

PR #278, opened 2021-02-08
Mac M1 fix proposal.

devnexen:
The actual code works fine on the usual Mac Intel however, some failures
 occurs with some unit tests on the ARM h/w when zero'ing page range.
While at it, using MADV_FREE_REUSABLE which is Darwin's specific but helps
command line tools to report better memory accounting.

### DIFF
--- a/src/pal/pal_apple.h
+++ b/src/pal/pal_apple.h
@@ -35,6 +35,21 @@ namespace snmalloc
      * descriptor for the mapping.
      */
     static constexpr int anonymous_memory_fd = VM_MAKE_TAG(PALAnonID);
+
+    /**
+     * Note: The root's implementation works fine on Intel
+     * however mprotect/PROT_NONE fails on ARM
+     * especially since the 11.2 release (seems known issue
+     * spotted in various projects; might be a temporary fix).
+     */
+    template<bool page_aligned = false>
+    static void zero(void* p, size_t size) noexcept
+    {
+      if constexpr (Aal::aal_name != ARM)
+        PALBSD::zero(p, size);
+      else
+        ::bzero(p, size);
+    }
   };
 } // namespace snmalloc
 #endif


src/pal/pal_apple.h
```diff
@@ -35,6 +36,31 @@ namespace snmalloc
...
+    /*
+     * Note : The root's implementation works fine on Intel
+     * however mprotect/PROT_NONE fails on ARM especially
+     * since the 11.2 release (seems known issue spotted
+     * in various projects ; might be a temporary fix).
+     */
+    template<bool page_aligned = false>
+    static void zero(void* p, size_t size) noexcept
+    {
+      ::bzero(p, size);
```
davidchisnall, line 49: If this is required only on AArch64 (is it only AArch64 macOS, or iOS derivatives as well?) then can we guard it on a `constexpr if` check that the AAL is for AArch64, not x86?

src/pal/pal_apple.h
```diff
@@ -35,6 +36,31 @@ namespace snmalloc
...
+    /*
+     * MADV_FREE_REUSABLE has the same expected effect than
+     * the more common MADV_FREE however better fit to Darwin's
+     * internal ie all code relying on task_info reports much
+     * more accurate resident memory.
+     */
+    static void notify_not_using(void* p, size_t size) noexcept
+    {
+      SNMALLOC_ASSERT(is_aligned_block<PALBSD::page_size>(p, size));
+      while (madvise(p, size, MADV_FREE_REUSABLE) == -1 && errno == EAGAIN)
```
davidchisnall, line 61: I don't believe this is correct.  `MADV_FREE_REUSABLE` is completely undocumented (and therefore Apple reserves the right to remove it or modify its semantics in minor version updates) but the comments in Apple's open-source libmalloc indicate that you must `madvise` with `MADV_FREE_REUSE` before reuse.  We would need a `notify_using` implementation that does this.  I don't know what the performance differences are between `MADV_FREE` and the pair of `MADV_FREE_REUSEABLE` + `MADV_FREE_REUSE`, if it's significant then we might want to provide both options to allow users to tune for lower memory or higher performance.

davidchisnall (changes requested):
Thanks for working on this, a couple of comments inline.

src/pal/pal_apple.h
devnexen: fair enough. yes Mac M1 is "only" a super iOs device anyway.

src/pal/pal_apple.h
devnexen: Yes this doesd even appear on the man page. however it s not related at all to performance (it s mostly the same) but memory accounting "faithfulness" but fair point I can give up this part.

davidchisnall: From what I can tell, it's not just about accounting, it has different semantics.  The traditional `MADV_FREE` has to track the dirty bit to tell whether you've used the page and so can require an IPI and TLB shoot-down to make the page read-only in the map before it's actually removed.  This version can reclaim the page immediately but requires a system call (which explicitly synchronises any thread that might be about to use the page) before it can be reused.  

It looks like it's a good fit for snmalloc, but we do need the corresponding `notify_using` function in the PAL.

devnexen: Might be better in another PR as here it s mainly to fix some unit tests crashing whereas this is improvement.

davidchisnall (approved):

src/pal/pal_apple.h
davidchisnall: Sounds good, thanks!

PR #279, opened 2021-02-09
Remove an undefined behaviour

mjp41:
The initialisation had an awkward sequence that used undefined
behaviour. This fixes this.


nwf:
LGTM

nwf (approved):

PR #280, opened 2021-02-09
Second batch for Mac M1 (mainly) changes proposal to make the whole

devnexen:
work more realibly, in both mono and multi thread contexts.

### DIFF
--- a/src/pal/pal_apple.h
+++ b/src/pal/pal_apple.h
@@ -3,6 +3,7 @@
 #ifdef __APPLE__
 #  include "pal_bsd.h"
 
+#  include <errno.h>
 #  include <mach/vm_statistics.h>
 #  include <utility>
 
@@ -49,6 +50,37 @@ namespace snmalloc
         PALBSD::zero(p, size);
       else
         ::bzero(p, size);
+    }
+
+    /**
+     * Overriding here to mark the page as reusable
+     * rolling it as much as necessary.
+     * As above, the x86 h/w worked alright without this change
+     * however now large allocations work better and more reliably
+     * with on ARM, not to mention better RSS number accuracy
+     * for tools based on task_info API.
+     */
+    static void notify_not_using(void* p, size_t size) noexcept
+    {
+      SNMALLOC_ASSERT(is_aligned_block<PALBSD::page_size>(p, size));
+      while (madvise(p, size, MADV_FREE_REUSABLE) == -1 && errno == EAGAIN)
+        ;
+    }
+
+    /**
+     * same remark as above but we need to mark the page as REUSE
+     * first
+     */
+    template<ZeroMem zero_mem>
+    static void notify_using(void* p, size_t size) noexcept
+    {
+      SNMALLOC_ASSERT(
+        is_aligned_block<PALBSD::page_size>(p, size) || (zero_mem == NoZero));
+      while (madvise(p, size, MADV_FREE_REUSE) == -1 && errno == EAGAIN)
+        ;
+
+      if constexpr (zero_mem == YesZero)
+        zero<true>(p, size);
     }
   };
 } // namespace snmalloc


src/pal/pal_apple.h
```diff
@@ -50,6 +52,40 @@ namespace snmalloc
...
+    /**
+     * same remark as above but we need to mark the page as REUSE
+     * first
+     */
+    template<ZeroMem zero_mem>
+    static void notify_using(void* p, size_t size) noexcept
+    {
+      SNMALLOC_ASSERT(
+        is_aligned_block<PALBSD::page_size>(p, size) || (zero_mem == NoZero));
+      os_unfair_lock lck = OS_UNFAIR_LOCK_INIT;
```
mjp41, line 80: Not sure if there are some crazy macro tricks here, but should it be static?
```suggestion
      static os_unfair_lock lck = OS_UNFAIR_LOCK_INIT;
```

src/pal/pal_apple.h
```diff
@@ -50,6 +52,40 @@ namespace snmalloc
...
+     * rolling it as much as necessary.
+     * As above, the x86 h/w worked alright without this change
+     * however now large allocations work better and more reliably
+     * with on ARM, not to mention better RSS number accuracy
+     * for tools based on task_info API.
+     */
+    static void notify_not_using(void* p, size_t size) noexcept
+    {
+      SNMALLOC_ASSERT(is_aligned_block<PALBSD::page_size>(p, size));
+      while (madvise(p, size, MADV_FREE_REUSABLE) == -1 && errno == EAGAIN)
```
mjp41, line 67: Does this need to be guarded by the same lock at in `notify_using`?

src/pal/pal_apple.h
```diff
@@ -50,6 +52,40 @@ namespace snmalloc
...
+     * first
+     */
+    template<ZeroMem zero_mem>
+    static void notify_using(void* p, size_t size) noexcept
+    {
+      SNMALLOC_ASSERT(
+        is_aligned_block<PALBSD::page_size>(p, size) || (zero_mem == NoZero));
+      os_unfair_lock lck = OS_UNFAIR_LOCK_INIT;
+      os_unfair_lock_lock(&lck);
+      while (madvise(p, size, MADV_FREE_REUSE) == -1 && errno == EAGAIN)
```
mjp41, line 82: The client of `snmalloc` could call `madvise(_,_, MADV_FREE_REUSE_`, and it wouldn't be guarded by a lock.  Is this okay?

mjp41 (commented):
I don't know these APIs, but could you clarify a little for me please. Thanks

src/pal/pal_apple.h
devnexen: true

src/pal/pal_apple.h
devnexen: yes in my local rebase it got lost indeed.

src/pal/pal_apple.h
devnexen: yes that depends of the usage, in fact I am trying to be prudent here as I still learn with the Mac M1 specifics maybe at some point it can be changed.

src/pal/pal_apple.h
davidchisnall: Why does this need protecting by a lock?  Do concurrent calls to this `madvise` call fail?

src/pal/pal_apple.h
devnexen: No it does not.

src/pal/pal_apple.h
devnexen: No it does not. As said I was prudent, maybe a bit too much finally. I can remove those if you wish ?

davidchisnall: If the call won't fail due to concurrent uses, I don't think there's any point in protecting it with a lock.

src/pal/pal_apple.h
```diff
@@ -3,7 +3,9 @@
 #ifdef __APPLE__
 #  include "pal_bsd.h"
 
+#  include <errno.h>
 #  include <mach/vm_statistics.h>
+#  include <os/lock.h>
```
mjp41, line 8: Unused header now?

mjp41 (approved):
LGTM

davidchisnall (approved):

PR #281, opened 2021-02-10
Make LowMemoryNotification object allocated

mjp41:
This makes a separate allocation for the callback object.  This makes
it easier for different callbacks to be used.

### DIFF
--- a/src/mem/largealloc.h
+++ b/src/mem/largealloc.h
@@ -57,7 +57,7 @@ namespace snmalloc
   // global state of the allocator.  This is currently stored in the memory
   // provider, so we add this in.
   template<SNMALLOC_CONCEPT(ConceptPAL) PAL>
-  class MemoryProviderStateMixin : public PalNotificationObject
+  class MemoryProviderStateMixin
   {
     /**
      * Simple flag for checking if another instance of lazy-decommit is
@@ -153,8 +153,10 @@ namespace snmalloc
       // Register this allocator for low-memory call-backs
       if constexpr (pal_supports<LowMemoryNotification, PAL>)
       {
-        allocated->PalNotificationObject::pal_notify = &(allocated->process);
-        PAL::register_for_low_memory_callback(allocated);
+        auto callback =
+          allocated->template alloc_chunk<LowMemoryNotificationObject, 1>(
+            allocated);
+        PAL::register_for_low_memory_callback(callback);
       }
 
       return allocated;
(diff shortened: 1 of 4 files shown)


mjp41:
For #272 I need some time/timer bits in the Pal, and this is a minor tidying to make that cleaner.  Making the memory_provider inherit from the callback class makes it hard to be notified of two things uniformly.

src/mem/largealloc.h
```diff
@@ -207,15 +209,26 @@ namespace snmalloc
...
-    }
+      MemoryProviderStateMixin<PAL>* memory_provider;
+
+      /***
+       * Method for callback object to perform lazy decommit.
+       */
+      static void process(PalNotificationObject* p)
+      {
+        // Unsafe downcast here. Don't want vtable and RTTI.
+        auto self = reinterpret_cast<LowMemoryNotificationObject*>(p);
```
nwf, line 222: Do you anticipate `PalNotificationObject`s that don't need to refer to the `memory_provider`?  If not, maybe just move the field up into the parent type?  Or maybe templatize `PalNotificationObject` on the type of state it's carrying?

src/mem/largealloc.h
```diff
@@ -153,8 +153,10 @@ namespace snmalloc
       // Register this allocator for low-memory call-backs
       if constexpr (pal_supports<LowMemoryNotification, PAL>)
       {
-        allocated->PalNotificationObject::pal_notify = &(allocated->process);
-        PAL::register_for_low_memory_callback(allocated);
+        auto callback =
+          allocated->template alloc_chunk<LowMemoryNotificationObject, 1>(
+            allocated);
+        PAL::register_for_low_memory_callback(callback);
```
davidchisnall, line 159: Does this need to be a dynamically allocated thing?  Can we make it a field and use the `std::conditional` and no_unique_address trick to make it not use any space if not used?

mjp41: No unique address is a C++20 thing.  @nwf found issues when we tried to move to C++20 with the instantiation of the pagemap causing compile times to be insanely large.

mjp41: The main issue is it is a little complex for the construction to refer to the enclosing type by pointer.  So it would be much less nice code when embedded.

src/mem/largealloc.h
mjp41: Not easily.  As the Pal doesn't know about memory providers.  We need the list to carry types of this, so really we need existential types.  In C++ that really means virtual, but generally we have tried to stay away from C++ vtables and RTTI in the allocator.

src/mem/largealloc.h
mjp41: I could manually combine the allocation a few lines higher:
https://github.com/microsoft/snmalloc/blob/ee470c535e258d93590488a4e1266bbe24dc90f1/src/mem/largealloc.h#L142-L145

This could modify the size to account for other allocations. However, variable allocation sizes like that seem to regularly be related to security vulnerabilities.  So I am not keen to introduce the anti-pattern.

mjp41: Okay, I have modified the address_space to allow for allocations below a power of two, and the address space will retain the extra.  Thus is you ask for 
  1. 1500 bytes
  2. 16 bytes
  3. 16 bytes 

Then it will place the 16 bytes in the gap between 1500 - 2048. This makes it cheaper to allocate multiple objects for clarity, while not incurring the rounding costs for the allocations.

mjp41: @davidchisnall thoughts?  What are you reasons for not wanting an allocation?

PR #283, opened 2021-02-19
Change behaviour of USE_POSIX_COMMIT_CHECKS

mjp41:
The previous setting applied USE_POSIX_COMMIT_CHECKS to snmalloc if it
was a non-release build.  This caused issues in CCF virtual mode, as it
was being built in RelWithDebInfo.

This commit changes the flag to be applied less, but for tests to always
apply the setting independent of build type.

This means that when snmalloc is being used as a library, it will be
off, unless explicitly requested.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,9 +35,10 @@ size_t malloc_usable_size(const void* ptr) { return 0; }
 int main() { return 0; }
 " CONST_QUALIFIED_MALLOC_USABLE_SIZE)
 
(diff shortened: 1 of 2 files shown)


nwf:
While you're in the neighborhood, https://github.com/microsoft/snmalloc/blob/ee470c535e258d93590488a4e1266bbe24dc90f1/src/pal/pal_bsd.h#L34 overrides `notify_not_using` but doesn't have code handling `USE_POSIX_COMMIT_CHECKS`, despite that `notify_using` isn't overridden and will `mprotect` things if that flag is set.

davidchisnall (approved):

achamayou (approved):

PR #284, opened 2021-02-20
POSIX_COMMIT_CHECKS adding for apple.

devnexen:


mjp41:
@devnexen thanks for thinking about this.  You also need the equivalent to 

https://github.com/microsoft/snmalloc/blob/master/src/pal/pal_bsd.h#L42-L44

to simulate actually removing the pages from being accessible. This is actually more important than the memset.

devnexen:
Yes I know but unfortunately func-malloc* tests abort with it.

mjp41:
Does it do that if we disable calling the `MADV_FREE_REUSABLE`, when we have the commit checks on?  And remove the converse `MADV_FREE_REUSE` in that mode too?

devnexen:
Unfortunately still does not still work (on Mac M1/ARM it should work on Mac Intel).

devnexen:
I propose doing it only for Mac Intel (it works as expected) if that s fine for you ?

devnexen:
With actual code both work under Mac Intel and ARM

mjp41 (changes requested):
LGTM to me, but Clang format is unhappy
https://dev.azure.com/snmalloc/snmalloc/_build/results?buildId=1308&view=logs&j=fb683405-d979-52da-6de9-2541dff429a6&t=f66a877f-3a0c-5131-ec8c-e97d54dcc0fe

@davidchisnall, any comments?

mjp41 (approved):

PR #285, opened 2021-02-23
Systematize validity checks on dealloc paths

nwf:
More work coming off the bottom of the #267 effort.  This one in particular was motivated by a desire to have some tests currently (and still, even after this PR) gated behind `CHECK_CLIENT` on in production without doing the full `check_size` call to `alloc_size` which, as written, would have had to go through much of this logic (and, eventually, amplification) redundantly.

While in the neighbourhood, expand a comment about the dance we're doing to include not just the happy path of a correct `dealloc`ation but also to consider double-free or wild-free cases.

This will change the errors reported for firing `CHECK_CLIENT` checks, but otherwise should not have any behavioural impact.


mjp41 (commented):
Sorry didn't get to review properly.  Minor nit from a quick skim.

src/mem/alloc.h
```diff
@@ -325,51 +294,63 @@ namespace snmalloc
       return external_alloc::free(p);
 #else
 
-      uint8_t size = chunkmap().get(address_cast(p));
+      uint8_t cmsk = chunkmap().get(address_cast(p));
```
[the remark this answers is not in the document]
nwf: It perhaps says something that that looks fine to me... >_>

src/mem/alloc.h
```diff
@@ -214,35 +214,6 @@ namespace snmalloc
...
-    /**
-     * Checks the allocation at `p` could have been validly allocated with
-     * a size of `size`.
-     */
-    void check_size(void* p, size_t size)
-    {
-#if defined(CHECK_CLIENT)
-      auto asize = alloc_size(p);
-      auto asc = size_to_sizeclass(asize);
-      if (size_to_sizeclass(size) != asc)
```
mjp41, line 226: This functionality appears to be lost with this change.  You are no longer checking that deallocations are the correct size class just whether they are the correct size kind (large, ,medium or small).

mjp41 (changes requested):

src/mem/alloc.h
mjp41: Also, I now wonder if this code works for large classes.  In particular, does 
```
free(malloc(SUPERSLAB_SIZE+1))
```
work with `CHECK_CLIENT`.

nwf: I surely wasn't intending to remove this check!  The `{small,medium}_dealloc_checked_chunkmap` and `large_dealloc_unchecked` functions are comparing the sizeclass derived from the provided size against the sizeclass in the chunkmap (for large allocs) or slab headers (for the rest).  What'd I miss?

mjp41: You're right.  I do wonder about adding tests for this.  Having a Custom Pal that intercepts the Error just, and then calling with a bunch of correct and incorrect deallocations.  Can be part of a separate PR. I am happy with this one.

mjp41 (approved):

PR #286, opened 2021-02-24
Add -Wconversion to clang builds

nwf:
MSVC has strong opinions on implicit conversions as used in CI, while Clang both
locally and in CI has weaker opinions.  In an effort to avoid subsequent
roundtrips through CI, make clang more strict.  Adding -Wconversion definitely
increases the strength of clang's opinions, apparently to include frowning on
some that even MSVC considers OK, so go make explicit the current implicit
behavior.


davidchisnall (commented):
I think we have a helper called `one_at_bit` or something for setting bit *n* of a `size_t`.  It would probably be cleaner to use that than to repeatedly `static_cast` one to a `size_t`

src/test/func/first_operation/first_operation.cc
```diff
@@ -194,18 +194,18 @@ int main(int, char**)
...
-    f(7ULL << exp);
-    f((1ULL << exp) + 1);
-    f((3ULL << exp) + 1);
-    f((5ULL << exp) + 1);
-    f((7ULL << exp) + 1);
-    f((1ULL << exp) - 1);
-    f((3ULL << exp) - 1);
-    f((5ULL << exp) - 1);
-    f((7ULL << exp) - 1);
+    f(static_cast<size_t>(1) << exp);
```
mjp41, line 197: This makes me very sad.  Perhaps, define a local lambda to do this? Something like:
```C++
auto shifted = [exp](size_t v) { v << exp };
f(shifted(1));
f(shifted(3));
```
If that doesn't cause any warnings.

mjp41 (approved):

PR #288, opened 2021-03-02
Reordering pal_noalloc due to the memset call in. caught with solaris…

devnexen:
… build.

### DIFF
--- a/src/pal/pal_noalloc.h
+++ b/src/pal/pal_noalloc.h
@@ -1,4 +1,9 @@
+#ifndef SNMALLOC_PAL_NO_ALLOC_H
+#define SNMALLOC_PAL_NO_ALLOC_H
+
 #pragma once
+
+#include <cstring>
 
 namespace snmalloc
 {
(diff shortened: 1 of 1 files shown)


mjp41:
Thanks for spotting this.  I would prefer to fix this by adding 
```
#include <cstring>
```
to `pal_noalloc.h`.  That file needs `memset`.  Forcing clangformat to include in the right order always seems a little wrong to me.

@davidchisnall,  do you have a preference here?

devnexen:
I thought doing this primarly but I thought it was not there for a reason.

mjp41:
Hence, why I pinged @davidchisnall :-)

devnexen:
We can eventually as third alternative rename to pal_zeroalloc.h :-)

mjp41:
LOL

davidchisnall:
I prefer headers to include everything that they need and have working guards against multiple inclusion.  I obviously failed here, sorry!

mjp41 (approved):

Issue #289, opened 2021-03-02
Extension of snmalloc for Pyoxidizer (python embedded in rust)


ryancinsight:
Hello again @mjp41, I started looking over the source code and found it should be simple enough to extend the options to rust and python for our use case and it looks like I should be able to obtain everything from src/override/malloc.cc. However, can you explain the "SNMALLOC_NAME_MANGLE(realloc)", i.e. what is the extern name that I am calling from rust here? I did some simple tests moving code from malloc.cc to rust.cc and naming directly sn_realloc, sn_free, sn_malloc for use in rust, which works fine, but I'm sure you would rather not have duplicates in the code but I am unfamiliar with Name_mangle definitions.

Best

mjp41:
So this lets you define the name you want to call.  

For instance,
https://github.com/microsoft/snmalloc/blob/a1fc509a65cc68a8746ce21681e51dcb1fde531a/src/test/func/malloc/malloc.cc#L4-L5

Is building snmalloc, but with all the standard C things prefixed with `our_`.   E.g 

https://github.com/microsoft/snmalloc/blob/a1fc509a65cc68a8746ce21681e51dcb1fde531a/src/test/func/malloc/malloc.cc#L92

https://github.com/microsoft/snmalloc/blob/a1fc509a65cc68a8746ce21681e51dcb1fde531a/src/test/func/malloc/malloc.cc#L99

https://github.com/microsoft/snmalloc/blob/a1fc509a65cc68a8746ce21681e51dcb1fde531a/src/test/func/malloc/malloc.cc#L96

Probably, you want to do something like 

```C++
#define SNMALLOC_NAME_MANGLE(a) sn_##a
```

Then you will have all the `snmalloc` functions, but prefixed with `sn_`, so they don't conflict with the ones in libc.

Probably, the easiest thing to do is add
```C++
#define SNMALLOC_NAME_MANGLE(a) sn_##a
#include "malloc.cc"
```
inside `rust.cc`, and then expose the symbols inside `snmalloc-sys`:

https://github.com/SchrodingerZhu/snmalloc-rs/blob/master/snmalloc-sys/src/lib.rs

ryancinsight:
This worked perfectly, only issue was slowalloc.h was being called more than once so I had to remove direct call. I tested rust crate symbols locally and will send to @SchrodingerZhu  crate with added tests once this is reviewed.

mjp41:
[...]

PR #290, opened 2021-03-02
Add rust calls without allocation tracking

ryancinsight:
This is needed for the pyoxidizer project to extend allocation to the embedded python interface without the allocation tracking overhead in typical rust interface. This is for issue #289

### DIFF
--- a/src/override/rust.cc
+++ b/src/override/rust.cc
@@ -1,4 +1,4 @@
-#define SNMALLOC_NAME_MANGLE(a) snmalloc_##a
+#define SNMALLOC_NAME_MANGLE(a) sn_##a
 #include "../snmalloc.h"
 #include "malloc.cc"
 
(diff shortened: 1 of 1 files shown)


src/override/rust.cc
```diff
@@ -36,3 +37,23 @@ rust_realloc(void* ptr, size_t alignment, size_t old_size, size_t new_size)
   }
   return p;
 }
+
+extern "C" SNMALLOC_EXPORT void* sn_malloc(size_t size)
+{
+  return snmalloc_malloc(size);
```
mjp41, line 43: `snmalloc_malloc` will already be `extern "C"` and have `SNMALLOC_EXPORT` applied to it.  So I am not sure why you have the level of indirection?

src/override/rust.cc
```diff
@@ -1,5 +1,6 @@
-#include "../mem/slowalloc.h"
+#define SNMALLOC_NAME_MANGLE(a) snmalloc_##a
 #include "../snmalloc.h"
```
mjp41, line 2: With `malloc.cc` included then this header will be included already.

src/override/rust.cc
ryancinsight: Ah, that makes sense, I'll correct it now.

src/override/rust.cc
ryancinsight: I went ahead and removed ../snmalloc.h too then since malloc.cc has it.

mjp41 (approved):
LGTM

PR #293, opened 2021-03-08
Fix for Mac OS X 10.14

mjp41:
The dllist was able to call delete during a destructor if a template
flag was set.  This flag is never set in snmalloc, and was included
to enable reuse in another project.  This was triggering an error on
older mac builds.

This PR calls a templated function when the DLList is destructed.  Hence
other projects can specify the `delete` behaviour if required.

### DIFF
--- a/src/ds/dllist.h
+++ b/src/ds/dllist.h
@@ -1,5 +1,7 @@
 #pragma once
 
+#include "helpers.h"
+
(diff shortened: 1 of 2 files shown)


mjp41:
Good suggestions, I have moved `ignore` to `helpers.h`.

PR #294, opened 2021-03-12
Move link out of object space.

mjp41:
The link object was previously stored in a disused object.  This is
good for reducing meta-data, but if we want to reduce the meta-data
corruption potential, then this is not a good design choice.

This commit moves it into the Metaslab.

The overall change improves performance slightly in a couple of cases.

### DIFF
--- a/src/ds/cdllist.h
+++ b/src/ds/cdllist.h
@@ -110,6 +110,17 @@ namespace snmalloc
 #endif
     }
 
+    /**
+     * Nulls the previous pointer
+     *
+     * The Meta-slab uses nullptr in prev to mean that it is not part of a
+     * size class list.
+     **/
+    void null_prev()
+    {
+      prev = nullptr;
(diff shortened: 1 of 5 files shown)


mjp41:
@nwf this might make the CHERI changes slightly easier as it is one less bit to store in an exposed location.

nwf (approved):
Seems reasonable.  If I have it right, this replaces a `uint16_t` in each `Metaslab` with a two `void*`, but padding rules mean that it's probably just the two new `void*` and some more internal padding now.

mjp41:
> Seems reasonable. If I have it right, this replaces a uint16_t in each Metaslab with a two void*, but padding rules mean that it's probably just the two new void* and some more internal padding now.

Yep, I think it goes from 16 bytes on 64bit non-CHERI architectures to 32 bytes per slab, so still a very small overhead.

PR #295, opened 2021-03-13
202103 CHERI prep plumbing

nwf:
Take.. I dunno, five? six?  This supersedes #267.  I have not yet actually seen this version through to completion -- that is, I haven't glued the CHERI AAL and CheriBSD PAL onto it, yet, but it's been sufficiently slow going that I don't want to delay any longer before getting initial reviews.

The most salient feature is that there's just one pointer wrapper type, now.  An `AuthPtr<T, bounds>` points at `T` and has an indicated static approximation of its bounds.  Those come in several flavours: arena (awesome power, use carefully), slab (generally 1M or 16M chunks, but perhaps larger as part of large allocations), and alloc (a single allocation), and additionally carry indication of whether the `VMMAP` permission has been stripped (more generally, any "platform constraints" the PAL wants to apply).  In the future, I expect to enlarge the taxonomy with a notion of "has had its contents zeroed" as well.

Despite this building on my box, I am quite, quite certain that CI will be unhappy.  This is also going to conflict with #294, which should go first.  Expect minor changes as the CHERIfication actually happens, but hopefully the static typing has caught most of it.

### DIFF
--- a/docs/StrictProvenance.md
+++ b/docs/StrictProvenance.md
@@ -1,0 +1,229 @@
+# StrictProvenance Architectures
+
+To aid support of novel architectures, such as CHERI, which explicitly track pointer *provenance* and *bounds*, `snmalloc` makes heavy use of a `CapPtr<T, B>` wrapper type around `T*` values.
+You can view the annotation `B` on a `CapPtr<T, B>` as characterising the set of operations that are supported on this pointer, such as
+
+* address arithmetic within a certain range (e.g, a `Superslab` chunk)
+* requesting manipulation of the virtual memory mappings
+
+Most architectures and platforms cannot enforce these restrictions, but CHERI enables software to constrain its future use of particular pointers and `snmalloc` imposes strong constraints on its *client(s)* use of memory it manages.
+The remainder of this document...
+
+* gives a "quick start" guide,
+* provides a summary of the constraints imposed on clients,
+* motivates and introduces the internal `ArenaMap` structure and the `capptr_amplify` function, and
+* describes the `StrictProvenance` `capptr_*` functions provided by the Architecture Abstraction Layer (AAL) and Platform Abstraction Layer (PAL).
+
+## Preface
+
+The `CapPtr<T, B>` and `capptr_*` primitives and derived functions are intended to guide developers in useful directions; they are not security mechanisms in and of themselves.
+For non-CHERI architectures, the whole edifice crumbles in the face of an overzealous `reinterpret_cast<>` or `unsafe_capptr` member access.
+On CHERI, these are likely to elicit capability violations, but may not if all subsequent access happen to be within bounds.
+
+## Quick Start Guide
+
+### How do I safely get an ordinary pointer to reveal to the client?
+
+If you are...
+
+* Adding an interface like `external_pointer`, and so you have a `CapPtr<T, CBAllocE>`, `e`, whose bounds you want to *inherit* when revealing some other `CapPtr` `p`, use `capptr_rebound(e, p)` to obtain another `CapPtr<T, CBAllocE>` with address from `p`, then go to the last step here.
+
+* Otherwise, if your object is...
+
+  * an entire `SUPERSLAB_SIZE` chunk or bigger, you should have in hand a `CapPtr<T, CBChunk>` from the large allocator.  Use `capptr_export` to make a `CapPtr<T, CBChunkE>`, then use `capptr_chunk_is_alloc` to convert that to a `CapPtr<T, CBAllocE>`, and then proceed.  (If, instead, you find yourself holding a `CapPtr<T, CBChunkD>`, use `capptr_chunk_from_chunkd` first.)
+
+  * of size `sz` and smaller than such a chunk,
+
+    * and have a `CapPtr<T, CBChunkE> p` in hand, use `Aal::capptr_bound<T, CBAllocE>(p, sz)` to get a `CapPtr<T, CBAllocE>`, and then proceed.
+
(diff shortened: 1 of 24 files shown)


src/ds/ptrwrap.h
```diff
@@ -0,0 +1,227 @@
...
+    T* unsafe_auth_ptr;
+
+    /**
+     * nullptr is constructable at any bounds type
+     */
+    AuthPtr(const std::nullptr_t n) : unsafe_auth_ptr(n) {}
+
+    AuthPtr() : unsafe_auth_ptr(nullptr) {}
+
+    explicit AuthPtr(T* p) : unsafe_auth_ptr(p) {}
```
nwf, line 66: Why does MSVC think my explicit constructor isn't reachable? :(

mjp41:
@nwf, I think you should pull the first 13 commits out as a separate PR.  So they can be integrated.  I think it would be good to get a lot of the refactoring in sooner, so you will experience less churn.

nwf:
Specifically 13?  It seems more natural to split at 15 (i.e., `NFC: pointer_offset* functions always return void*`)?

mjp41:
I can't count. ;)

src/ds/ptrwrap.h
nwf: Aaaargh.  So, something about the `sandbox`'s use of an allocator that never needs initialization results in MSVC getting unhappy.  In particular, it seems to correctly realize that the `large_alloc()` `NeedsInitialization` code path cannot happen, but something about that path's need to unwrap and rewrap the `AuthPtr<>` (as/from a `void*` to pass through the `InitThreadAllocator` callback) results in it generating this warning that the `AuthPtr` ctor itself is unreachable. :(

I've added the `#pragma warning(push)`-and-friends dance and documentation around the constructor itself (sigh) as part of 53c30a6342.

docs/StrictProvenance.md
```diff
@@ -0,0 +1,126 @@
...
+We take the position that free list entries should be suitable for return, i.e., with authority bounded to their backing slab entry.
+(However, the *contents* of free memory may be dangerous to expose to the user and require clearing prior to handing out.)
+This means that allocation fast paths are unaffected by the requirement to bound return pointers, but that deallocation paths may need to amplify twice, once on receipt of the pointer from the application and again on receipt of the pointer from another `Alloc`.
+
+## Annotated Pointer Wrapper
+
+We introduce `AuthPtr<T, ptrauth_bounds>` as an alias for `T*` that is statically tagged with a summary of the architectural bounds data.
+At the moment, we introduce six possible annotations, though the taxonomy is imperfect:
+
+* bounded only to an underlying arena allocation without platform constraints, `PAArena`;
```
mjp41, line 90: Arena not introduced at this point.

docs/StrictProvenance.md
```diff
@@ -0,0 +1,126 @@
+# StrictProvenance Architectures
```
mjp41, line 1: Can you add a summary at the start of the file that says everything in paragraph.  Perhaps, something like:
  
>  To support novel architectures, such as CHERI, `snmalloc` uses an `AuthPtr` wrapper class that tracks the way in which this pointer can be used.
>  You can view the annotation on `AuthPtr` as specifying the type of operations that are supported by this pointer such as 
> *  address arithmetic within a certain range, i.e. slab.
> *  if virtual memory mappings maybe modified
>
> This restrictions are not imposed on most platforms, but on CHERI they can be enforced by hardware.  The rest of the document...

mjp41: I would also perhaps put the more background material later in the document, and focus on what someone modifying snmalloc needs to know at the start, which is kind of the opposite way round to you have it.

Who is this document for, and why are they reading it?  My imagination that most people reading md files on github: are trying to work out how code works, or how to change it.

nwf:
This spin has several things going for it:
* a new overview document, geared towards "how do I do the thing"
* `PASlab` renamed to `PAChunk`
* some new machinery around `PAChunkD`, letting us avoid needing to set bounds on pointers that aren't giving rise to things headed out the door (except on debug builds, where bounds setting is probably still a good plan)
* it's running on CHERI with additional changes merely in `src/aal` and `src/pal` (and disabling the `oe` tests in `CMakeLists.txt`)

The bottom-most commits in this PR are, at present, also replicated in #308 .

nwf:
CI failure is a timeout of one of the agents.  I'll leave this be and will trigger a re-run either upon being given the OK or given changes to make. :)

mjp41 (approved):

mjp41:
LGTM.

src/ds/address.h
```diff
@@ -24,6 +25,13 @@ namespace snmalloc
     return reinterpret_cast<U*>(reinterpret_cast<char*>(base) + diff);
   }
 
+  template<enum ptrauth_bounds bounds, typename T>
```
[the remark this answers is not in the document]
nwf: Oh, that seems reasonable, yes.

src/mem/remoteallocator.h
```diff
@@ -51,6 +51,15 @@ namespace snmalloc
     {
       return alloc_id_and_sizeclass & SIZECLASS_MASK;
     }
+
+    /** Zero out a Remote tracking structure, return pointer to object base */
+    template<capptr_bounds B>
+    SNMALLOC_FAST_PATH static CapPtr<void, B>
+    clear(CapPtr<Remote, B> self, sizeclass_t sizeclass)
+    {
+      pal_zero<Pal>(self, sizeof(Remote));
```
mjp41, line 60: Is using the default `Pal` the right choice here?

PR #296, opened 2021-03-15
202103 Pre-CHERIfication refactoring

nwf:
Carves off the first 15 commits of #295 for separate review as requested.

### DIFF
--- a/src/ds/address.h
+++ b/src/ds/address.h
@@ -18,19 +18,19 @@ namespace snmalloc
   /**
    * Perform pointer arithmetic and return the adjusted pointer.
    */
-  template<typename T>
-  inline T* pointer_offset(T* base, size_t diff)
+  template<typename U = void, typename T>
+  inline U* pointer_offset(T* base, size_t diff)
   {
-    return reinterpret_cast<T*>(reinterpret_cast<char*>(base) + diff);
+    return reinterpret_cast<U*>(reinterpret_cast<char*>(base) + diff);
   }
 
   /**
    * Perform pointer arithmetic and return the adjusted pointer.
    */
-  template<typename T>
-  inline T* pointer_offset_signed(T* base, ptrdiff_t diff)
+  template<typename U = void, typename T>
+  inline U* pointer_offset_signed(T* base, ptrdiff_t diff)
   {
-    return reinterpret_cast<T*>(reinterpret_cast<char*>(base) + diff);
+    return reinterpret_cast<U*>(reinterpret_cast<char*>(base) + diff);
   }
 
   /**
--- a/src/mem/mediumslab.h
+++ b/src/mem/mediumslab.h
@@ -82,13 +82,13 @@ namespace snmalloc
     }
 
     template<ZeroMem zero_mem, SNMALLOC_CONCEPT(ConceptPAL) PAL>
-    void* alloc(size_t size)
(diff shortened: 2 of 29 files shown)


src/ds/address.h
```diff
@@ -19,16 +19,16 @@ namespace snmalloc
    * Perform pointer arithmetic and return the adjusted pointer.
    */
   template<typename T>
```
davidchisnall, line 21: Maybe a default value of `void` for T?

mjp41, line 21: How about the following:
```C++
  template <typename T = void>
  inline T* pointer_offset(void* base, size_t diff)
  {
    return reinterpret_cast<T*>(reinterpret_cast<char*>(base) + diff);
  }
```
I think this might be slightly cleaner,

mjp41 (commented):
LGTM, I would prefer to get #294 in first. As the two will conflict slightly.

mjp41 (approved):

src/ds/address.h
nwf: With `AuthPtr`, we're going to lose the ability to implicitly convert to `void*`.  How would you feel about

```c++
  template<typename U = void, typename T>
  inline U* pointer_offset(T* base, size_t diff)
  {
    return reinterpret_cast<U*>(reinterpret_cast<char*>(base) + diff);
  }
```

src/mem/mediumslab.h
```diff
@@ -82,13 +82,13 @@ namespace snmalloc
     }
 
     template<ZeroMem zero_mem, SNMALLOC_CONCEPT(ConceptPAL) PAL>
-    void* alloc(size_t size)
+    static void* alloc(Mediumslab* self, size_t size)
```
davidchisnall, line 85: It would be good to have a comment explaining why this is a `static` method with an explicit `this`.  Same comment on the next ones.

src/mem/pagemap.h
```diff
@@ -53,7 +54,11 @@ namespace snmalloc
    * static ChunkMap object, which encapsulates knowledge about the
    * pagemap's parametric type T.
    */
-  template<size_t GRANULARITY_BITS, typename T, T default_content>
+  template<
```
davidchisnall, line 57: Please can you update the comment to explain what the template parameters are?

src/pal/pal_posix.h
```diff
@@ -196,6 +197,14 @@ namespace snmalloc
...
         if (r != MAP_FAILED)
           return;
+
+        /*
+         * We're falling back to zeroing the memory ourselves, which is not
+         * stellar, but correct.  However, mmap() has left errno nonzero in an
+         * effort to explain that MAP_FAILED.  Reset it here so subsequent
+         * checks don't see a stale value.
+         */
+        errno = 0;
```
davidchisnall, line 207: This is not actually POSIXly correct: the spec for `errno` says that it should not be modified by things that don't explicitly set it.  We should be getting `errno` before calling the thing that can set errno and resetting it later.  It would be quite nice to capture this in an RAII thing (capture errno in constructor, reset errno in destructor).

src/ds/address.h
mjp41: I forget that you can put a default not on the last one in C++.  Yes, that is good.

mjp41 (approved):

PR #297, opened 2021-03-17
More pre-CHERIfication refactoring

nwf:
Sorry; these four should have been part of #296 but I neglected to include them.  Hopefully similarly without objection.


mjp41 (approved):
All looks fine to me.

nwf:
Slightly different commits this time; one down, two pulled from atop to here.

PR #298, opened 2021-03-17
Pointer encoding

mjp41:
This uses the approach of signing the per slab free list pointers with a very simple reversible function. If you have 
```
  prev -> curr -> next
```
It effectively stores the value of `next` in `curr` but encoding using `prev`, to form a sequence that will detect corruption and double-free.

This is currently only on in Debug, or in Release if the `CHECK_CLIENT` flag is set. 

There are still other aspects that need improving
* Randomisation of keys
* Doesn't do anything about cross thread deallocations
* Only implemented for small size classes

The `CHECK_CLIENT` flag enables checking for the start of object, which still needs performance analysis, but without that enabled, this change has minimal regression to Release perf.


CMakeLists.txt
```diff
@@ -79,7 +79,7 @@ macro(clangformat_targets)
     message(WARNING "Not generating clangformat target, no clang-format tool found")
   else ()
     message(STATUS "Generating clangformat target using ${CLANG_FORMAT}")
-    file(GLOB_RECURSE ALL_SOURCE_FILES *.cc *.h *.hh)
+    file(GLOB_RECURSE ALL_SOURCE_FILES src/*.cc src/*.h src/*.hh)
```
nwf, line 82: Not that I object, but why the change?

src/mem/alloc.h
```diff
@@ -47,7 +47,7 @@ namespace snmalloc
   class FastFreeLists
   {
   protected:
-    FreeListHead small_fast_free_lists[NUM_SMALL_CLASSES];
+    FreeListIter small_fast_free_lists[NUM_SMALL_CLASSES];
```
nwf, line 50: Nice way to emphasize that these are populated all at once and consumed piecemeal.

src/mem/allocconfig.h
```diff
@@ -10,6 +10,17 @@ namespace snmalloc
 #  define CHECK_CLIENT
 #endif
 
+  SNMALLOC_FAST_PATH void check_client(bool test, const char* const str)
```
nwf, line 13: This seems to rely on inlining and DCE to avoid evaluating the test expression in non-`CHECK_CLIENT` builds.  Does `SNMALLOC_FAST_PATH` ensure that all that takes place?  Maybe we should be shouting (that is, using a macro, perhaps `SNMALLOC_CHECK_CLIENT(,)` like `SNMALLOC_ASSERT`) to be sure the compiler understands? ;)

Either way is, to be clear, much nicer than all the clutter from the `#ifdef CHECK_CLIENT`s.

src/mem/alloc.h
```diff
@@ -747,35 +747,29 @@ namespace snmalloc
...
         {
-          auto n = Metaslab::follow_next(prev);
+          auto curr = small_fast_free_lists[i].take();
 
-          Superslab* super = Superslab::get(prev);
-          Slab* slab = Metaslab::get_slab(prev);
-          small_dealloc_offseted_inner(super, slab, prev, i);
-
-          prev = n;
+          Superslab* super = Superslab::get(curr);
```
nwf, line 770: Not necessarily required for this PR, but looking ahead: it might be nice, while in the neighborhood, to somehow articulate that all objects in a `small_fast_free_list` are in the same `Slab`.  With the CHERI work, we're going to be storing allocation-bounded pointers in the `small_fast_free_list` and, because I was lazy here, we end up amplifying for each one to get its associated `Superslab` and `Slab`, but we should just need to do that once.  It's not a huge deal because this is just in the debugging path, tho'.

src/mem/metaslab.h
```diff
@@ -3,30 +3,13 @@
...
 
 namespace snmalloc
 {
   class Slab;
 
-  /**
-   * Free objects within each slab point directly to the next (contrast
-   * SlabLink, which chain different Slabs of the same sizeclass together).
-   */
-  struct SlabNext
```
nwf, line 16: Hm.  Sorry to have churned the code base by introducing this only for it to be removed.

src/mem/freelist.h
```diff
@@ -0,0 +1,297 @@
+#pragma once
+/**
+ * This file encapsulates the in disused object free lists
+ * that are used per slab.
```
nwf, line 4: Perhaps "small-object Slab" not just "slab" to emphasize that this is not used for `Mediumslab`s or `Largeslab`s?

src/mem/freelist.h
```diff
@@ -0,0 +1,297 @@
...
+
+  static inline bool different_slab(void* p1, void* p2)
+  {
+    return different_slab(address_cast(p1), address_cast(p2));
+  }
+
+  /**
+   * Free objects within each slab point directly to the next.
+   * The next_object pointer can be encoded to detect
+   * corruption caused by writes in a UAF or a double free.
```
nwf, line 53: Perhaps add commentary, assuming it's accurate, that `FreeObject`s are cache-friendly-offsetted within objects and so overlap `Remote`s?  Since we're eventually going to want (a configuration option) to zero on the `dealloc` side, it's important to note where we might deliberately put nonzeros and have to clear them on the `alloc` side.

src/mem/freelist.h
```diff
@@ -0,0 +1,297 @@
...
+  class FreeObject
+  {
+    FreeObject* next_object;
+
+    static FreeObject* encode(uintptr_t local_key, FreeObject* next_object)
+    {
+#ifdef CHECK_CLIENT
+      if constexpr (aal_supports<IntegerPointers>)
+      {
+        // Simple symmetric encoding
```
nwf, line 64:
```suggestion
        // Simple involutional encoding.  The bottom half of each word is multiplied by
        // a function of both global and local keys (the latter, in practice, being the
        // address of the previous list entry) and the resulting word's top half is XORed
        // into the pointer value before it is stored.
```

src/mem/freelist.h
```diff
@@ -0,0 +1,297 @@
...
+        next_object = reinterpret_cast<FreeObject*>(next);
+      }
+#else
+      UNUSED(local_key);
+#endif
+      return next_object;
+    }
+
+  public:
+    static FreeObject* make(void* p)
```
nwf, line 79: Come to think of it, should we be applying cache-friendly offsets _here_ (and removing below) so that the `Allocator` functions can just pass around `void *`s that always point to the start of the object?

src/mem/freelist.h
```diff
@@ -0,0 +1,297 @@
...
+    void add(void* n)
+    {
+      SNMALLOC_ASSERT(!different_slab(end.get_curr(), n));
+      FreeObject* next = FreeObject::make(n);
+      end.set_next_and_move(next);
+    }
+
+    /**
+     * Adds a terminator at the end of a free list,
+     * but doesn't not close the builder.  Thus new elements
```
nwf, line 266:
```suggestion
     * but does not close the builder.  Thus new elements
```

src/mem/freelist.h
```diff
@@ -0,0 +1,297 @@
...
+  };
+
+  /**
+   * Used to build a free list in object space.
+   *
+   * Checks signing of pointers
+   */
+  class FreeListBuilder : public FreeListIter
+  {
+    FreeObjectCursor end;
```
nwf, line 240: If we just upcast, this won't go away even though (I suspect?) we never downcast from an `Iter` to a `Builder`.  Something something move constructor?

src/mem/freelist.h
```diff
@@ -0,0 +1,297 @@
...
+#include <iostream>
+
+namespace snmalloc
+{
+#ifdef CHECK_CLIENT
+  /**
+   * The key that is used to encode free list pointers.
+   * This should be randomised at startup in the future.
+   */
+  inline static uintptr_t global_key =
```
nwf, line 22:
```suggestion
  inline static address_t global_key =
```
here and below?

src/mem/metaslab.h
```diff
@@ -90,57 +73,25 @@ namespace snmalloc
...
-
     bool valid_head()
     {
       size_t size = sizeclass_to_size(sizeclass);
-      size_t slab_end = (address_cast(head) | ~SLAB_MASK) + 1;
-      uintptr_t allocation_start =
-        remove_cache_friendly_offset(address_cast(head), sizeclass);
+      auto h = address_cast(free_queue.peek());
+      size_t slab_end = (h | ~SLAB_MASK) + 1;
+      uintptr_t allocation_start = remove_cache_friendly_offset(h, sizeclass);
```
nwf, line 94:
```suggestion
      address_t slab_end = (h | ~SLAB_MASK) + 1;
      address_t allocation_start = remove_cache_friendly_offset(h, sizeclass);
```

src/mem/slab.h
```diff
@@ -121,12 +128,21 @@ namespace snmalloc
         return Superslab::NoSlabReturn;
       }
 
-      // Remove from the sizeclass list and dealloc on the superslab.
+      size_t count = 1;
+      // Check free list is well-formed.
```
nwf, line 132: Expand on "well-formed"?  On `IntegerPointer` architectures with `CHECK_CLIENT`, WHP, we'll catch tampering with the list as well as cyclicity due to the conflicting previous addresses for the earliest node participating in the cycle.

nwf (changes requested):
Looks good; some scattered minor changes requested, but nothing fundamental.

CMakeLists.txt
mjp41: I have about 20 different build directories inside, and this was searching all of them looking for something to format, and the WSL perf was annoying.

src/mem/metaslab.h
mjp41: Don't worry, it was actually useful the churn gave me the places I need to change it for this.

src/mem/freelist.h
mjp41: The only issue is that they use allocator local state to introduce the offsets.

src/mem/slab.h
mjp41: Probably should `ifdef` this code.

src/mem/freelist.h
mjp41: It is so assignment from a builder to an iter "works". But I should perhaps codify that more carefully.

CMakeLists.txt
davidchisnall: Putting your build directory outside your source directory avoids a lot of that and also means that you can use a faster FS for the builds (e.g. with WSL2, put the sources in the share FS but the build directories in a Linux native FS).  I typically do builds from a directory on a ZFS FS that has sync disabled (if the machine crashes during a build and I lose some files in my build directory, I don't care).

I'm not sure this glob is actually correct.  Does this find files in `src/{whatever}/*.h`?

mjp41: It is correct:

```
Examples of recursive globbing include:

/dir/*.py  - match all python files in /dir and subdirectories
```
https://cmake.org/cmake/help/latest/command/file.html#filesystem

src/mem/alloc.h
mjp41: So that is checked in the CHECK_CLIENT version for Integer pointers.  I think we can ensure `take` is doing some checking elsewhere?

src/mem/freelist.h
mjp41: Made inheritance private, and refactored slightly.

src/mem/alloc.h
nwf: I'm content to leave it not explicitly checked on CHERI, since we'll blow up pretty fast if we take something out of bounds.  How'd you feel about code like
```
  if (!small_fast_free_lists[i].empty()) {
    auto first = small_fast_free_lists[i].peek(); // wrap with large_allocator.ptrauth_amplify() once AuthPtr<>s land
    auto slab = Metaslab::get_slab(first);
    auto super = Superslab::get(first);
    while (!small_fast_free_lists[i].empty()) {
      // ...
```

mjp41: Done

src/mem/freelist.h
mjp41: Made inheritance non-public.

src/mem/freelist.h
mjp41: Leaving as can be part of future PR. If we decide to keep cache-friendly offsets.

src/mem/freelist.h
mjp41: Added some comments

nwf (changes requested):
Looks good to me; can you fix my email address in the one commit message?

mjp41:
> Looks good to me; can you fix my email address in the one commit message?

What to?

nwf:
> > Looks good to me; can you fix my email address in the one commit message?
> 
> What to?

Please use my `@microsoft.com` hat.

PR #299, opened 2021-03-19
Better reciprocal division and modulus.

mjp41:
The previous reciprocal division branch on the prime that the sizeclass
was constructed from. All sizeclasses can be represented as
  2^n * {1,3,5,7}
This lead to a very small table, but some work to calculate the
appropriate shifts and multiplications to implement reciprocal division.

This commit uses a completely uniform representation for every
sizeclass using a lookup table.  Due to the precise ranges that we query
the modulus and rounding on, we can do this much more efficiently.

The func-release-rounding exhaustively tests all the queries we are
interested in.

### DIFF
--- a/src/ds/helpers.h
+++ b/src/ds/helpers.h
@@ -73,18 +73,26 @@ namespace snmalloc
   template<size_t length, typename T>
   class ModArray
   {
+    /**
+     * Align the elements, so that access is cheaper.
+     */
+    struct alignas(bits::next_pow2_const(sizeof(T))) TWrap
+    {
+      T v;
+    };
+
     static constexpr size_t rlength = bits::next_pow2_const(length);
-    T array[rlength];
+    TWrap array[rlength];
 
   public:
     constexpr const T& operator[](const size_t i) const
     {
-      return array[i & (rlength - 1)];
+      return array[i & (rlength - 1)].v;
     }
 
     constexpr T& operator[](const size_t i)
     {
-      return array[i & (rlength - 1)];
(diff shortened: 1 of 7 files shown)


src/ds/helpers.h
```diff
@@ -73,18 +73,26 @@ namespace snmalloc
...
+     * Align the elements, so that access is cheaper.
+     */
+    struct alignas(bits::next_pow2_const(sizeof(T))) TWrap
+    {
+      T v;
+    };
+
     static constexpr size_t rlength = bits::next_pow2_const(length);
-    T array[rlength];
+    TWrap array[rlength];
```
nwf, line 85: Glancing around, it seems like this is only going to change the `ModArray<SLAB_COUNT, Metaslab> meta` field of a `Superslab` from allocating each `Metaslab` 56 bytes (53 + 3 padding on a 64-bit pointer architecture) to 64.  Everything else looks like it's already a power of two size (`size_t`, `uint16_t`).  Is that right?

nwf (approved):
LGTM; just one clarifying question this time 'round.

src/ds/helpers.h
mjp41: Correct.  The codegen for Metaslab lookup was getting a bit messy due to a multiple by 7, which was hindering a bunch of consolidating optimisations.  @plietar suggested this approach, and it seems much more principled than the
```C++
  void* dummy;
```
I was planning to add as a field to `Metaslab`.

PR #300, opened 2021-03-19
Remove old features

mjp41:
There are a few features that are no longer used this PR removes them.


nwf (approved):
LGTM; if CI's happy with it, I think I am, too.

davidchisnall:
Happy with the histogram going on - that brought in an iostream dependency, which made the kernel version hard to write.  I'm not sure what `allow_reserve` actually did - I don't see anything in the removed code that used the template parameter to do anything other than pass it down.  At some point, I'd like to do an implementation of [`malloc(9)`](https://www.freebsd.org/cgi/man.cgi?query=malloc&apropos=0&sektion=9&manpath=FreeBSD+12.2-RELEASE+and+Ports&arch=default&format=html) on top of snmalloc and the `M_NOWAIT` flag requires a mechanism to avoid calling into the PAL, which may block.

mjp41:
The intent was the same as `M_NOWAIT`.  Calling into the CLR Pal required you to setup various context to enable thread suspend, which is expensive, and required very infrequently.  This was originally implemented to allow the call to happen without those costs, and then backoff if we needed to request more memory from the platform.  As it wasn't tested its functionality had gotten broken at some point.

PR #301, opened 2021-03-22
Remove allocated field from Metaslab

mjp41:
The metaslab contains a field specifying how many elements have been
allocated.  As the code has evolved this field has now always become
the maximum capacity of the slab for the sizeclass.

This commit looks up this value based on the sizeclass, and removes the
field from the slab's metadata.


nwf (approved):
LGTM

davidchisnall (approved):

PR #302, opened 2021-03-23
Bug fix for superslab meta-data

mjp41:
If the metadata is made larger, then the calculation of which classes can use the "short slab" was incorrect.  In particular, if the meta data grows above half the size of a slab, then things go very wrong.  This alters that calculation.

### DIFF
--- a/src/mem/slab.h
+++ b/src/mem/slab.h
@@ -32,8 +32,7 @@ namespace snmalloc
     static SNMALLOC_FAST_PATH void
     alloc_new_list(void*& bumpptr, FreeListIter& fast_free_list, size_t rsize)
     {
-      void* slab_end =
-        pointer_align_up<SLAB_SIZE>(pointer_offset(bumpptr, 1));
+      void* slab_end = pointer_align_up<SLAB_SIZE>(pointer_offset(bumpptr, 1));
 
       FreeListBuilder b;
       SNMALLOC_ASSERT(b.empty());
--- a/src/mem/slab.h
+++ b/src/mem/slab.h
@@ -33,7 +33,7 @@ namespace snmalloc
     alloc_new_list(void*& bumpptr, FreeListIter& fast_free_list, size_t rsize)
     {
       void* slab_end =
-        pointer_align_up<SLAB_SIZE>(pointer_offset(bumpptr, rsize));
+        pointer_align_up<SLAB_SIZE>(pointer_offset(bumpptr, 1));
 
       FreeListBuilder b;
       SNMALLOC_ASSERT(b.empty());
--- a/src/mem/superslab.h
+++ b/src/mem/superslab.h
@@ -5,6 +5,7 @@
 #include "metaslab.h"
 
 #include <new>
+#include <iostream>
 
 namespace snmalloc
 {
@@ -73,8 +74,10 @@ namespace snmalloc
 
     static bool is_short_sizeclass(sizeclass_t sizeclass)
     {
-      constexpr sizeclass_t h = size_to_sizeclass_const(sizeof(Superslab));
-      return sizeclass <= h;
+      static_assert(SLAB_SIZE > sizeof(Superslab), "Meta data requires this.");
+      // Note that h will be the next size class above the space, so use < in comparison.
+      constexpr sizeclass_t h = size_to_sizeclass_const(SLAB_SIZE - sizeof(Superslab));
+      return sizeclass < h;
     }
 
     void init(RemoteAllocator* alloc)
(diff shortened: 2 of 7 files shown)


src/mem/superslab.h
```diff
@@ -73,8 +74,10 @@ namespace snmalloc
 
     static bool is_short_sizeclass(sizeclass_t sizeclass)
```
nwf, line 75: It might be good to have some overview of the whole "short slab" scheme, say, at the top of `superslab.h`.  Just spitballing some text:

```
Superslabs are, to first approximation, a `CHUNK_SIZE`-sized and -aligned
region of address space, internally composed of a header (a `Superslab`
structure) followed by an array of `Slab`s, each `SLAB_SIZE`-sized and
-aligned.  Each active `Slab` holds an array of identically sized allocations
strung on an invasive free list, which is lazily constructed from a
bump-pointer allocator (see `Metaslab::alloc_new_list`).

In order to minimize overheads, Slab metadata is held externally, in `Metaslab`
structures; all `Metaslab`s for the Slabs within a Superslab are densely packed
within the `Superslab` structure itself.  Moreover, as the `Superslab`
structure is typically much smaller than `SLAB_SIZE`, a "short Slab" is
overlaid with the `Superslab`.  This short Slab can hold only allocations that
are smaller than the `SLAB_SIZE - sizeof(Superslab)` bytes; see
`Superslab::is_short_sizeclass`.  The Metaslab state for a short slabs is
constructed in a way that avoids branches on fast paths; effectively, the
object slots that overlay the `Superslab` at the start are omitted from
consideration.
```

src/mem/slab.h
```diff
@@ -33,7 +33,7 @@ namespace snmalloc
     alloc_new_list(void*& bumpptr, FreeListIter& fast_free_list, size_t rsize)
     {
       void* slab_end =
-        pointer_align_up<SLAB_SIZE>(pointer_offset(bumpptr, rsize));
+        pointer_align_up<SLAB_SIZE>(pointer_offset(bumpptr, 1));
```
nwf, line 36: This isn't actually a bug (with the correction below in place), yes?  If the precondition for `alloc_new_list` is that `bumpptr` is at least `rsize` before the end of the slab (and I think that's the case?), then these will both produce the same `slab_end`, right?

src/mem/superslab.h
```diff
@@ -73,8 +74,10 @@ namespace snmalloc
...
     {
-      constexpr sizeclass_t h = size_to_sizeclass_const(sizeof(Superslab));
-      return sizeclass <= h;
+      static_assert(SLAB_SIZE > sizeof(Superslab), "Meta data requires this.");
+      // Note that h will be the next size class above the space, so use < in comparison.
```
nwf, line 78:
```suggestion
      /*
       * size_to_sizeclass_const rounds *up* and returns the smallest class that
       * could contain (and so may be larger than) the free space available for the
       * short slab.  While we could detect the exact fit case and compare `<= h`
       * therein, it's simpler to just treat this class as a strict upper bound and
       * only permit strictly smaller classes in short slabs.
       */
```

src/test/measuretime.h
```diff
@@ -3,12 +3,19 @@
...
+#include <sstream>
 
-#define DO_TIME(name, code) \
-  { \
-    auto start__ = std::chrono::high_resolution_clock::now(); \
-    code auto finish__ = std::chrono::high_resolution_clock::now(); \
-    auto diff__ = finish__ - start__; \
-    std::cout << name << ": " << std::setw(12) << diff__.count() << " ns" \
-              << std::endl; \
+class MeasureTime : public std::stringstream
```
nwf, line 8: I'm not quite sure how to feel about this.  I think... it's cute, but it seems very odd to both extend `std::stringstream` and have such striking side effects in the destructor.  Still, it does make for rather EDSL-esque use.

nwf (commented):
LGTM, but maybe take this as an opportunity for more commentary?

src/mem/slab.h
mjp41: Yeah, this isn't needed.  It was part of my trying to chase down the but, but I thought it was perhaps nicer with 1, rather than `rsize`

src/test/measuretime.h
mjp41: I was trying to minimise the change from the original macro.

src/mem/slab.h
nwf: I'm happy to have it use `1` rather than `rsize`, I just wanted to check my understanding

nwf (approved):

PR #305, opened 2021-03-23
FreeObject: hide harder from the compiler, tweak types

nwf:
On CHERI, the compiler will always issue a warning for
`reinterpret_cast<T*>(vaddr_t)` and similar expressions, and of course, if the
compiler can see far enough into the types, the presence of `if constexpr` will
not save us.  Therefore, lift the conditional out to two definitions of
`FreeObject::encode` using `std::enable_if_t` to gate which is used.

While in the neighborhood, use `address_t` rather than `uintptr_t`; the former
is the latter on all non-CHERI platforms, but is `vaddr_t` rather than
`uintcap_t` on CHERI.

### DIFF
--- a/src/mem/freelist.h
+++ b/src/mem/freelist.h
@@ -68,29 +68,45 @@ namespace snmalloc
   {
     FreeObject* reference;
 
-  public:
-    static inline FreeObject*
-    encode(uint16_t local_key, FreeObject* next_object)
-    {
-#ifdef CHECK_CLIENT
-      if constexpr (aal_supports<IntegerPointers>)
-      {
-        // Simple involutional encoding.  The bottom half of each word is
-        // multiplied by a function of both global and local keys (the latter,
-        // in practice, being the address of the previous list entry) and the
-        // resulting word's top part is XORed into the pointer value before it
-        // is stored.
-        auto next = address_cast(next_object);
-        constexpr address_t MASK = bits::one_at_bit(PRESERVE_BOTTOM_BITS) - 1;
-        // Mix in local_key
-        address_t key = (local_key + 1) * global_key;
-        next ^= (((next & MASK) + 1) * key) &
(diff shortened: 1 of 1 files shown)


jrtc27:
Perhaps you want to move to ptraddr_t from vaddr_t? The latter is a deprecated CheriBSDism.

jrtc27:
Oh, you don't actually use vaddr_t anywhere in snmalloc, just the commit message here.

nwf:
Oh, but good call.  The `address_t = vaddr_t` bit is still pending in my CHERIfication tree and it'd be best to fix that before I forget.

mjp41:
I think this is going to have some interesting interactions with #304.

src/mem/freelist.h
```diff
@@ -59,29 +59,44 @@ namespace snmalloc
   {
     FreeObject* next_object;
 
-    static FreeObject* encode(uintptr_t local_key, FreeObject* next_object)
+    /**
+     * On architectures which use IntegerPointers, we can obfuscate our free
+     * lists and use this to drive some probabilistic checks for integrity
+     */
+    template<typename AAL = Aal, typename T = FreeObject>
+    static std::enable_if_t<aal_supports<IntegerPointers, AAL>, T*>
```
mjp41, line 67: Can you make this
```
static std::enable_if_t<aal_supports<IntegerPointers, AAL> && CHECK_CLIENT, T*>
```
?

Then you can remove the `ifdef` inside, and use the second encode for that case.

mjp41 (changes requested):
I  changed the `uintptr_t` to address in another PR, as I was touching this already.  I think it is just the hiding left.

mjp41 (approved):
LGTM, just a little conflict with my merged changes.

mjp41 (approved):

PR #306, opened 2021-03-24
Threshold freelist wakeup

mjp41:
When a slab has been fully allocated, then we no longer
check it has entries until something returns an allocation to this slab.
However, it is possible that only a single allocation is available, and
then we can end up frequently on the slow path.

This change only considers free lists that cover at least 1/8 of a slab.
This means that we will hit the slow path less frequently.  This also
means that the randomisation changes will have more entropy: with a
single element free list there is only one order.

For large small sizes it can still be a single element, as 1/8 is of the
slab capacity is below 1.

### DIFF
--- a/src/mem/freelist.h
+++ b/src/mem/freelist.h
@@ -47,12 +47,12 @@ namespace snmalloc
 #endif
   }
 
-  static inline bool different_slab(uintptr_t p1, uintptr_t p2)
+  static inline bool different_slab(address_t p1, address_t p2)
   {
     return ((p1 ^ p2) >= SLAB_SIZE);
   }
 
-  static inline bool different_slab(uintptr_t p1, void* p2)
+  static inline bool different_slab(address_t p1, void* p2)
   {
     return different_slab(p1, address_cast(p2));
   }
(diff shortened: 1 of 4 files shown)


src/mem/allocconfig.h
```diff
@@ -4,12 +4,6 @@
 
 namespace snmalloc
 {
-// The CHECK_CLIENT macro is used to turn on minimal checking of the client
-// calling the API correctly.
-#if !defined(NDEBUG) && !defined(CHECK_CLIENT)
-#  define CHECK_CLIENT
-#endif
-
```
nwf, line 12: This hunk seems either unrelated or mistaken, I amn't sure which.  Do you mean for `CHECK_CLIENT` to be completely determined by `cmake` now?

nwf (changes requested):
Mostly looks fine, save for the one hunk (which is also fine if it's intentional, but perhaps should be in its own commit within this PR).

I do wonder, though, if we want to bound the number of slabs we're keeping in this suspended animation state?  Absent any such check, in pathological cases, this might increase resident size by 12% or so?

src/mem/allocconfig.h
mjp41: It should probably be another PR.  I really want to make 

*   Secure for users of snmalloc
*   Debug for developers for snmalloc

These are different and switching on checking in Debug using encrypted pointers is not the easiest to debug issues.  We build all the tests with the checked flag on.  But everything checked is checking is a user error, not a snmalloc error.

mjp41:
> I do wonder, though, if we want to bound the number of slabs we're keeping in this suspended animation state? Absent any such check, in pathological cases, this might increase resident size by 12% or so?

Yeah. I am not sure I am happy with that worst case either.  Really I want to have enough entropy that the inside a slab reordering of allocations is sufficient to provide randomisation.  Open to alternatives.  I am also not sure what can be done with the fast path to adapt.

mjp41:
I've added a max trigger ratio of 32 elements, this will reduce the worst case for most sizeclasses.  But not remove it entirely.

src/mem/freelist.h
```diff
@@ -352,6 +351,34 @@ namespace snmalloc
 #endif
     }
 
+    /**
+     *  Calculates the length of the queue.
+     *  This is O(n) as it walks the queue.
+     *  If this is needed in a none debug setting then
```
sylvanc, line 357: `none debug` -> `non-debug`

sylvanc (approved):

PR #307, opened 2021-03-25
Provide a source of entropy

mjp41:
The randomisation was previously using constants.  This calls an external source
of entropy, and uses this to seed a PRNG for future needs.

The entropy is threaded into the various freelist construction and read routines.

- [x]  Need to add Pal support for Entropy, with fall back to `std::random_device`.


src/mem/entropy.h
```diff
@@ -0,0 +1,93 @@
+#include <cstdint>
+#include <random>
+#include <type_traits>
+
+#include "../ds/address.h"
+
+namespace snmalloc
+{
+  template<typename Pal>
```
nwf, line 9: I've been trying to use `PAL` for template variables and `Pal` for the global, just to avoid shadowing.

src/mem/alloc.h
```diff
@@ -683,6 +684,11 @@ namespace snmalloc
       if (isFake)
         return;
 
+      // Entropy must be first, so that all data-structures can use the key
+      // it generates.
+      // This must occur before any freelists are constructed.
+      entropy.init<typename MemoryProvider::Pal>();
```
nwf, line 690: We are probably inconsistent about referring to `MemoryProvider::Pal` vs. just the `Pal` that's in scope.  Most often, those are the same, but it might be good to examine all the places we've defaulted `<typename PAL = Pal>`?  Doesn't have to be this PR; maybe just raise an issue so we don't forget?

src/mem/entropy.h
```diff
@@ -0,0 +1,93 @@
...
+
+  template<typename Pal>
+  std::enable_if_t<!pal_supports<Entropy, Pal>, uint64_t> get_entropy64()
+  {
+    std::random_device rd;
+    uint64_t a = rd();
+    return (a << 32) ^ rd();
+  }
+
+  class LocalEntropy
```
nwf, line 23: Maybe split into a class for `bit_source` and `constant_key` and one for `local_key` and `local_counter`?  That'd make the use elsewhere in the tree a little more obvious, I think.  (e.g., `get_next` is used only to populate `constant_key` and `bit_source` in `init()` here.)

src/mem/entropy.h
mjp41: Good catch

src/mem/entropy.h
mjp41: I agree and disagree on this. 

`refresh_bits` calls `get_next` so uses most of it.  

I think `get_next` might be useful in other parts, as we could generate new keys for each free list.  Leaving this open makes the refactoring easier at the moment.

mjp41:
@achamayou @anakrish I have been adding some security features to `snmalloc`.  These require a source of entropy, what do you recommend I use on OE?

achamayou:
@mjp41 https://github.com/openenclave/openenclave/blob/57c224925df6df7e65bfdd81dd8cd78fc06c8be6/include/openenclave/internal/entropy.h#L38 for seeding a PRNG, or https://github.com/openenclave/openenclave/blob/57c224925df6df7e65bfdd81dd8cd78fc06c8be6/include/openenclave/internal/random.h if you want randomness directly.

mjp41:
@devnexen this is quite cross cutting on various platforms.  Hopefully, I haven't broken one you care about.

devnexen:
I had a quick look at your changes, most of systems nowadays has getentropy coming originally from openbsd, freebsd has it since 12.x I believe, illumos ... I think only netbsd does not have it but I ll test shortly your branch.

devnexen:
so far only netbsd fails due to getentropy call absent. macos, freebsd ... all fine.

src/pal/pal_posix.h
```diff
@@ -264,5 +270,31 @@ namespace snmalloc
...
+      }
+      else if constexpr (OS::get_entropy64 != get_entropy64)
+      {
+        // Derived Pal has provided a custom definition.
+        return OS::get_entropy64();
+      }
+      else
+      {
+        uint64_t result;
+        if (getentropy(&result, sizeof(result)) != 0)
```
devnexen, line 294: genuine question, would arc4random_buf would work as fallback for unsupported systems since it s aiming userland allocation only anyway ?

mjp41: Anything is better than nothing.  I am not an expert on APIs for random, so I'll look at that one.

mjp41: My current plan was to use 
```
std::random_device
```
as the back-up.  Would that be okay on the platforms that aren't currently working.  So if I just disable the Pal feature flag like I have on Haiku for those platforms.

devnexen: fair enough. ;-)

devnexen: also was just thinking but that depends how far back you re willing to support older oses versions. getentropy was introduced at various timing, so depends if you stick to constexpr approach vs cmake+ifdef'ries

SchrodingerZhu:
https://android.googlesource.com/platform/bionic/+/master/libc/bionic/getentropy.cpp

do not know whether android is happy w/ this. should be fine?

SchrodingerZhu:
> so far only netbsd fails due to getentropy call absent, dragonflybsd has getrandom but not getentropy counterpart. macos, freebsd ... all fine.

Is arc4random_buf an alternative？

devnexen:
> > so far only netbsd fails due to getentropy call absent, dragonflybsd has getrandom but not getentropy counterpart. macos, freebsd ... all fine.
> 
> Is arc4random_buf an alternative？

Not entirely, purpose wise.

devnexen:
> https://android.googlesource.com/platform/bionic/+/master/libc/bionic/getentropy.cpp
> 
> do not know whether android is happy w/ this. should be fine?

Not sure it worth it in this case here, `std::random_device` usage sounds reasonable but not my call :-)

mjp41:
> so far only netbsd fails due to getentropy call absent, dragonflybsd has getrandom but not getentropy counterpart. macos, freebsd ... all fine.

So I have switched to `getrandom` as it covers one additional platform. I have made NetBSD fall back to `std::random_device".  With this PR I am mostly aiming to get to a working everywhere, and will improve other platforms and support going to older versions with future PRs.

src/pal/pal_posix.h
mjp41: I am not keen to go to far down the ifdef'ries, but will if strong demand.

devnexen:
too bad openbsd does not have getrandom :-)

mjp41:
> too bad openbsd does not have getrandom :-)

Doh!

mjp41:
> > too bad openbsd does not have getrandom :-)
> 
> Doh!

@devnexen I have changed to give netbsd a custom pal calling `getrandom` for Entropy, and moved back to `getentropy` on most platforms.  So currently only DragonflyBSD and Haiku have fallen back to `std::random_device`.  Hopefully, this is an acceptable start point, that doesn't break anything, and we can improve, if required, in the future.

src/pal/pal_open_enclave.h
```diff
@@ -66,6 +67,17 @@ namespace snmalloc
...
       oe_memset_s(p, size, 0, size);
     }
+
+    /**
+     * Source of Entropy
+     */
+    static uint64_t get_entropy64()
+    {
+      uint64_t result = 0;
+      if (oe_random(&result, sizeof(result)) != 0)
```
achamayou, line 77:
```suggestion
      if (oe_random(&result, sizeof(result)) != OE_OK)
```

achamayou (approved):

src/pal/pal_open_enclave.h
mjp41: Do we have OE_OK in scope when we build snmalloc for OE?  I'll add it to the simulatation layer, but if you could check it works when you build that would be great.

achamayou: If we have oe_random, then we have oe_result_t. If we're missing one or the other, I'll add them.

src/pal/pal_posix.h
```diff
@@ -264,5 +270,31 @@ namespace snmalloc
...
+     * This is a default that works on many POSIX platforms.
+     */
+    static uint64_t get_entropy64()
+    {
+      if constexpr (!pal_supports<Entropy, OS>)
+      {
+        // Derived Pal does not provide entropy.
+        return 0;
+      }
+      else if constexpr (OS::get_entropy64 != get_entropy64)
```
nwf, line 286: I'm sure I'm missing something, but doesn't `OS` inherit from us?  That is, if this condition is true we won't be here in the first place?

mjp41: OS could override.  In which case the overriding definition will take over that is
  `OS::get_entropy64`
will be the override, and 
  `get_entropy 64`
will be the non-overriden definition. In my experiments, this did the right thing, but I may be wrong.  And this is not tested by CI, so extra care is required.

nwf: Sorry, what I mean is: if `Pal` is `OS<Pal_Posix>` then `Pal::get_entropy64` will be `OS::get_entropy64` not `Pal_Posix::get_entropy64`, and so we won't even get to this `if constexpr` to consider it...  unless we refer to `Pal_Posix` directly somewhere (except to extend from it)?

mjp41: I think inside that branch of the `constexpr` we could put anything that compiles on any platform.

This definition is useful, if any of the Pal_Posix layer requires Entropy itself.  Let's say we add our own ASLR to the Pal_Posix, then that can use the system supplied Entropy, which may be overridden by the platform.  Does that make sense?

mjp41: @nwf is this okay with you?

PR #309, opened 2021-04-06
Further Randomisation

mjp41:
This PR implements the full randomisation of a small slab of allocations.  This uses an algorithm to build a random cycle in a slab, and then encodes that into the free list.  This provides a lot of entropy in the layout within a slab. 

The PR also refactors the FreeListBuilder, so it can be more compact and not introduce randomisation. This PR reestablishes the non-CHECK_CLIENT performance that was impacted by some of the randomisation code in previous PRs.

### DIFF
--- a/src/ds/bits.h
+++ b/src/ds/bits.h
@@ -55,6 +55,7 @@ namespace snmalloc
 
     SNMALLOC_FAST_PATH size_t clz(size_t x)
     {
+      SNMALLOC_ASSERT(x != 0); // Calling with 0 is UB on some implementations
 #if defined(_MSC_VER)
 #  ifdef USE_LZCNT
 #    ifdef SNMALLOC_VA_BITS_64
@@ -148,6 +149,8 @@ namespace snmalloc
 
     inline size_t ctz(size_t x)
     {
+      SNMALLOC_ASSERT(x != 0); // Calling with 0 is UB on some implementations
+
 #if defined(_MSC_VER)
 #  ifdef SNMALLOC_VA_BITS_64
       return _tzcnt_u64(x);
@@ -229,7 +232,9 @@ namespace snmalloc
     inline size_t next_pow2_bits(size_t x)
     {
       // Correct for numbers [1..MAX_SIZE].
-      // Returns 64 for 0. Approximately 2 cycles.
+      if (x == 1)
+        return 0;
+
       return BITS - clz(x - 1);
     }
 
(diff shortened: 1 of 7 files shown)


src/ds/bits.h
```diff
@@ -228,6 +231,9 @@ namespace snmalloc
 
     inline size_t next_pow2_bits(size_t x)
     {
+      if (x == 1)
+        return 0;
+
       // Correct for numbers [1..MAX_SIZE].
       // Returns 64 for 0. Approximately 2 cycles.
```
nwf, line 238: Stale comment?

src/mem/slab.h
```diff
@@ -37,24 +37,51 @@ namespace snmalloc
...
+      // is described as outside-in without citation on Wikipedia, appears to be
+      // Folklore algorithm.
+      PreAllocObject* curr = pointer_offset<PreAllocObject>(bumpptr, 0);
+      curr->next = curr;
+      uint16_t count = 1;
+      for (PreAllocObject* p = pointer_offset<PreAllocObject>(bumpptr, rsize); p < slab_end;
+           p = pointer_offset<PreAllocObject>(p, rsize))
+      {
+        size_t insert_index = entropy.sample(count);
+        p->next = std::exchange(pointer_offset<PreAllocObject>(bumpptr, insert_index * rsize)->next, p);
```
nwf, line 63: A `make clangformat` may be in order?

src/mem/freelist.h
```diff
@@ -392,53 +411,56 @@ namespace snmalloc
...
-        prev[1] = HEAD_KEY;
-        curr[1] = initial_key(h) & 0xffff;
-#endif
-        end[1] = &(head[1]);
+        // If second list is non-empty, perform append.
+        if (end[1] != &head[1])
+        {
+          end[1]->store(nullptr, get_prev(1), entropy);
+
+          // Append 1 to 0
```
nwf, line 424: Possibly worth refactoring out to a method for appending these encoded lists?  If you think it won't get used anywhere else I'm happy to leave it in place, tho'.

nwf (changes requested):

src/mem/freelist.h
mjp41: Generally I'm in favour of doing things like this, but I am not sure this will add much value as almost all the code is about appending, or depends on precisely what the append does.

src/mem/slab.h
mjp41: The last thing I changed ;)

nwf (approved):

PR #310, opened 2021-04-06
Correct PAL Concept

nwf:
The test for NoAllocation/AlignedAllocation/neither was broken and this was
undetected until the sandbox demo came into being with its NoAllocation PAL
instance.  (Sadly, this remained undetected for so long because we can't
routinely build with C++20 because std::atomic_t<>'s initialization rules
changed in ways that make us require oodles and oodles of RAM at compile time.
This patch has been validated with -j1 on a machine with lots of RAM.)

Separately, it's possible that we end up here without `using namespace std`,
which seems like overkill.  So just use `std::size_t` ourselves within the PAL
Concept definition.


mjp41 (approved):
LGTM.

PR #312, opened 2021-04-12
netbsd build fix.

devnexen:
getentropy not being guarded, providing prototype even tough we end up
 using the C++ api version.

### DIFF
--- a/src/pal/pal_netbsd.h
+++ b/src/pal/pal_netbsd.h
@@ -1,6 +1,18 @@
 #pragma once
 
 #ifdef __NetBSD__
+#  include <cassert>
+
+namespace snmalloc
+{
+  template<typename... Args>
+  int getentropy(Args...)
+  {
+    assert(0 && "Unreachable path");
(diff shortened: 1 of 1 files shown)


mjp41 (approved):
LGTM.  Thanks for fixing.

davidchisnall:
This looks fine.  I'd originally suggested that NetBSD should do something like this:

```c++
namespace snmalloc
{
  int getentropy(void*, size_t)
  {
    assert(0 && "should not be used");
  }
}
```

This avoids declaring a function in the global namespace, which might trip up later things that use SFINAE to detect its presence.

I think you might be able to get away with this:

```c++
namespace snmalloc
{
  template<typename... Args>
  int getentropy(Args...)
  {
    static_assert(false, "This implementation should never be called");
  }
}
```
I believe that this won't be instantiated (and the static assert won't fire) unless the superclass method is called.

mjp41 (approved):

PR #313, opened 2021-04-15
CapPtr taxonomy changes

nwf:
This PR shims a conversion from `enum` to a multi-parameter template type for `capptr_bounds` underneath an equivalent to #311 .  This may be more cleanly extensible as we add dimensions (the second commit is an excellent worked example; I anticipate adding boolean zeroed and MTE-colourised dimensions in the near future).  Doubtless some of it highlights my ignorance of the loftier points of C++; commentary more than welcome (and @mjp41 retains permission to complain, as he said he might prior to my undertaking this refactoring ;) ).

There's a `concept` of bounds here, too, but CI won't save us from ourselves as we don't/can't yet use C++20 there.  I've build-tested with C++17 and 20 in both debug and release flavours, though, so hopefully haven't missed something there.

### DIFF
--- a/src/ds/ptrwrap.h
+++ b/src/ds/ptrwrap.h
@@ -38,59 +38,111 @@ namespace snmalloc
    * *E forms are "exported" and have had platform constraints applied.  That
    * means, for example, on CheriBSD, that they have had their VMMAP permission
    * stripped.
-   *
-   * Yes, I wish the start-of-comment characters were aligned below as well.
-   * I blame clang format.
    */
-  enum capptr_bounds
+
+  namespace capptr_bounds
   {
-    /*                Spatial  Notes                                      */
-    CBArena, /*        Arena                                              */
-    CBChunkD, /*       Arena   Chunk-bounded in debug; internal use only! */
-    CBChunk, /*        Chunk                                              */
-    CBChunkE, /*       Chunk   (+ platform constraints)                   */
-    CBAlloc, /*        Alloc                                              */
-    CBAllocE /*        Alloc   (+ platform constraints)                   */
+    /*
+     * Bounds dimensions are sorted so that < reflects authority.  For example,
+     * spatial::Alloc is less capable than spatial::Arena.
+     */
+    enum class spatial
+    {
+      Alloc,
+      Chunk,
+      ChunkD,
+      Arena
+    };
+    enum class platform
+    {
+      Exported,
+      High
+    };
+
+    template<spatial S, platform P>
+    struct t
+    {
+      static constexpr enum spatial spatial = S;
+      static constexpr enum platform platform = P;
+
+      template<enum spatial SO>
+      using with_spatial = t<SO, P>;
+
+      template<enum platform PO>
+      using with_platform = t<S, PO>;
+
+      /*
+       * The dimensions here are not used completely orthogonally.  In
+       * particular, high-authority spatial bounds imply high-authority in other
+       * dimensions and unchecked pointers must be annotated as tightly bounded.
+       */
+      static_assert(!(S == spatial::Arena) || (P == platform::High));
+    };
+
+    using CBArena = t<spatial::Arena, platform::High>;
+    using CBChunk = t<spatial::Chunk, platform::High>;
+    using CBChunkD = t<spatial::ChunkD, platform::High>;
+    using CBChunkE = t<spatial::Chunk, platform::Exported>;
+    using CBAlloc = t<spatial::Alloc, platform::High>;
+    using CBAllocE = t<spatial::Alloc, platform::Exported>;
+
+    // clang-format off
+#ifdef __cpp_concepts
+    template<typename T>
+    concept c = requires()
+    {
+      { T::spatial } ->ConceptSame<const spatial>;
+      { T::platform } ->ConceptSame<const platform>;
+    };
+#endif
+    // clang-format on
   };
+
+  /*
+   * Defining these above and then `using` them to hide the namespace label like
+   * this is still shorter than defining these completely out here, since there
+   * isn't a notion of local namespace imports.
+   */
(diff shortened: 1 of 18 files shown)


src/ds/ptrwrap.h
```diff
@@ -38,67 +38,138 @@ namespace snmalloc
...
+    };
+    enum class platform
+    {
+      Exported,
+      High
+    };
+    enum class wild
+    {
+      Unchecked,
+      Checked
```
mjp41, line 64: We have to be careful about changing from `Unchecked` to `Checked` in particular, we could end up encoding a "time of check to time of use" attack into the API design?

src/ds/ptrwrap.h
```diff
@@ -38,67 +38,138 @@ namespace snmalloc
...
-    CBAllocE /*        Alloc   (+ platform constraints)                   */
+    /*
+     * Bounds dimensions are sorted so that < reflects authority.  For example,
+     * spatial::Alloc is less capable than spatial::Arena.
+     */
+    enum class spatial
+    {
+      Alloc,
+      Chunk,
+      ChunkD,
```
mjp41, line 53: You have dropped the description on the `D` postfix.

src/ds/ptrwrap.h
```diff
@@ -38,67 +38,138 @@ namespace snmalloc
...
+       * The dimensions here are not used completely orthogonally.  In
+       * particular, high-authority spatial bounds imply high-authority in other
+       * dimensions and unchecked pointers must be annotated as tightly bounded.
+       */
+      static_assert(
+        !(S == spatial::Arena) || (P == platform::High && W == wild::Checked));
+      static_assert(
+        !(W == wild::Unchecked) ||
+        (S == spatial::Alloc && P == platform::Exported));
+    };
```
mjp41, line 93: So I was wondering if 

> Alloc => Exported

If the Alloc is does not cover all the pages it is part of, then we don't want to give out `High`, but if it is 4KiB aligned and sized, then it is okay to give out the ability to call `mprotect` to the client.  But the allocator would have be clear the `mprotect` on return?

src/ds/ptrwrap.h
```diff
@@ -38,67 +38,138 @@ namespace snmalloc
...
+      static constexpr enum wild wild = W;
+
+      template<enum spatial SO>
+      using with_spatial = t<SO, P, W>;
+
+      template<enum platform PO>
+      using with_platform = t<S, PO, W>;
+
+      template<enum wild WO>
+      using with_wild = t<S, P, WO>;
```
mjp41, line 81: Nice SML records into C++ templates ;-)

mjp41 (approved):
LGTM  Seems clean.  Generally it is easier to read as the order of aspects of the capability is easier to follow, than the previous sets of capability.  Doesn't seem to impact far either.  So I am in favour of this change.

src/ds/ptrwrap.h
nwf: It is not currently the case that `spatial::Alloc` implies `platform::Exported`.  In particular, when we are reconstructing internal pointer(s) from the `wild::Unchecked` pointer given to `dealloc` by the client, we do not apply platform constraints before placing messages on the `Remote` queues.  We certainly could, but because `Mediumslab`s do not actually use the pointer given to them except as an address, it seemed excessive.  However, in thinking about that more, perhaps that's not true: if it's worth bounding `Remote`s to `spatial::Alloc` (to thwart UAF as a vector to access allocator metadata, e.g.), it's probably worth bounding them to `platform::Exported` as well.  Probably merits additional thought.

nwf: It might actually make sense to have AAL-directed policies here... on CHERI+MTE, any bounding we do of `Remote` pointers or other bits of in-band metadata will be out of paranoia about bugs in `snmalloc` itself, as we will ensure that the client cannot read the band.  OTOH, for just CHERI, it probably is worth having the spatial bounds in place as a kind of lightweight, partial defence against UAF?  For non-CHERI, of course, all this is just type-level fun and games.

src/ds/ptrwrap.h
nwf: Yes, and this is more subtle than I'd like.  A `wild::Checked` pointer just means that we have validated, on CHERI, that the capability is tagged and that it points at its base (and nothing on non-CHERI).  It doesn't mean that it's within an allocation arena or corresponds to an allocated object.

In the final taxonomy, I expect pointers from the client to be
- `wild::Unchecked`
- `platform::Exported`
- `temporal::Monochrome`
- `spatial::Alloc`
- `zero::Arbitrary`
and then we will check (and change)...
- capability attributes (convert to `wild::Checked`)
- that the pointer points into an arena (via amplification, giving us a `wild::Checked`, `platform::High`, `temporal::Rainbow`, `spatial::Arena`, `zero::Arbitrary` analogue to the client pointer)
- that we have been told the right size (for size-parametric `dealloc`s; see `*_dealloc_unchecked` and `*_dealloc_checked_chunkmap`).
- that, in fact, it points to the start of an object (by determining the sizeclass and checking `is_start_of_object` or analogue; see `*_dealloc_checked_sizeclass`; at this point we can convert down to a `spatial::Alloc` pointer again)
- that the MTE colours aligned (using our `temporal::Rainbow` rights to atomically clock the 0th granule of the object from the provided colour to the next; see placeholders in `*_dealloc_start`; at this point, we can convert down to `temporal::Monochrome` and `zero::Zeroed` when we `bzero_with_color()` the memory)

Phew!  And then we can thread the thing on a `Remote` queue or hand it off to the backing {`Slab`, `Mediumslab`, `LargeAlloc`} for dealing with.

nwf:
Closing; going to redo this on the `snmalloc2` branch.

PR #315, opened 2021-04-19
Fix static-sized alloc paths, add compilation test

nwf:
The CapPtr refactoring was largely compiler guided; unfortunately, it turns out
that the static-sized alloc function is not well exercised.  As such, some type
errors and unnecessary unsafety lurked behind missing template instantiation.

Correct those and add calls to the test harness to make sure we always generate
at least one instance of each small/medium/large case.  While here, it doesn't
hurt to make sure that we call all three possible dealloc() flavors as well.
This will, if nothing else, force instantiation of the static-sized dealloc
template as well.

### DIFF
--- a/src/mem/alloc.h
+++ b/src/mem/alloc.h
@@ -139,18 +139,18 @@ namespace snmalloc
 
       if constexpr (sizeclass < NUM_SMALL_CLASSES)
       {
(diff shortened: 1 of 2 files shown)


mjp41 (approved):
LGTM

PR #316, opened 2021-04-20
Deal with C++20 atomics

nwf:
Alright, so... as mentioned, C++20 does away with trivial initializers for `std::atomic<T>` types (presumably for the benefit of platforms that need some more exotic implementation internally).  As such, our `GlobalPagemap`s full of `std::atomic<T>`s get global ctors created for them... that run in undefined order, and so perhaps _after_ some other ctors that dive in to snmalloc and set some values in our pagemaps.  That's rude.  Moreover, at least clang's generator ctors take human-perceptable amounts of time to zero all that .bss, which is just adding insult to injury.

Fortunately, C++20 introduces `std::atomic_ref<T>` for an atomic reference to non-atomic state, so long as some simple rules are obeyed (referent must outlive reference; no concurrent atomic and nonatomic access; and no concurrent access to subobjects, even if via atomic).  (See http://open-std.org/jtc1/sc22/wg21/docs/papers/2020/n4861.pdf section 31.7 for details.)  Our pagemaps use atomics in rather boring, and so amenable, ways: the tables are basically immortal; we _always_ access the internal table atomically; and, if there are subobjects in `T` (there aren't), we don't know about them.

So, introduce a `snmalloc::TrivialInitAtomic<T>` type that gives the desired behaviour on both standards.  On C++17 and earlier, this uses `std::atomic<T>` directly for storage and takes references to them as `std::atomic<T>&`.  On C++20 (and later, one assumes), it uses `T` for storage and takes references using `std::atomic_ref<T>`.

This has the side-effect of working around clang's excessive memory use when compiling a global `std::atomic<T> [LOTS]` in C++20 mode, and so might make it reasonable for more people to build `snmalloc` that way (and benefit? from the `concept`... goodness? that comes with it).  `gcc`, interestingly, did not suffer that particular pathology, but its concept error messages are, IMHO/IME a little less useful for understanding when something has gone amiss.

I've tested this on my desk with {`clang`,`gcc`} x {C++17, C++20}.  Let's see what CI thinks.


src/ds/helpers.h
```diff
@@ -156,4 +158,69 @@ namespace snmalloc
...
+   * became harder to accomplish in C++20.  Fortunately, our rules for accessing
+   * these are at least as strong as those required by C++20's atomic_ref:
+   *
+   *   * The objects outlive any references to them
+   *
+   *   * We always access the objects through references (though we'd be allowed
+   *     to access them without if we knew there weren't other references)
+   *
+   *   * We don't access sub-objects at all, much less concurrently through
+   *     other references.
```
mjp41, line 173: Is this property true of the recursive version used on platforms without lazy commit?

mjp41 (changes requested):
Thanks for working out a route forward for C++20.  I think we probably should have tests for C++20 in CI.  Could you add
* A CMake option to build with C++20
* A build task with this option set

davidchisnall:
Skimming the spec, `std::atomic_ref` appears to have condition-variable behaviour (wait + notify).  Is it guaranteed to be the same size as the underlying type?  libc++ doesn't seem to have an implementation yet, does libstdc++?

src/ds/helpers.h
nwf: AIUI, yes: in the recursive case, the objects in question are either a pointer to another page or a `T` on a page; we never, say, claim atomic access to the entire array on either variety of page.

nwf:
`std::atomic<T>` also has `wait` and `notify` behavior in C++20; it's really all very... exciting.  I suspect that that is part of the motivation to drop the trivial initializer.  The good news is that on C++20 our storage is just an array of `T` again and we don't care about the size of the ephmeral `std::atomic_ref<T>`.

It looks like `libc++`'s implementation is WIP at https://reviews.llvm.org/D72240 ; irritatingly, this means that we're still in the `std::atomic<T>[]` storage case on CHERI even when we're building for C++20, so probably for the moment we need to restrict to C++17 for CHERI builds.  `libstdc++`'s implementation is in tree as of https://github.com/gcc-mirror/gcc/commit/a90fe12c80a13ffed9d5c775703788275d8b7c42 .

nwf:
Now that CI is updated as per #317, let's tickle it again and see if we pass on the clang-10 C++20 CI target.

mjp41 (approved):
LGTM.  Thanks for the CI work.

PR #319, opened 2021-04-29
Refactorings for encoding remote deallocation next pointers

mjp41:
The remote deallocation next pointers are not protected in the same way that free list pointers are. This PR provides refactorings to make it easier to add those changes.

### DIFF
--- a/src/mem/alloc.h
+++ b/src/mem/alloc.h
@@ -916,18 +916,29 @@ namespace snmalloc
       if (likely(sizeclass < NUM_SMALL_CLASSES))
       {
         SNMALLOC_ASSERT(super->get_kind() == Super);
+        check_client(
+          super->get_kind() == Super,
+          "Heap Corruption: Sizeclass of remote dealloc corrupt.");
         auto slab =
           Metaslab::get_slab(Aal::capptr_rebound(super.as_void(), p_offseted));
+        check_client(
+          super->get_meta(slab)->sizeclass() == sizeclass,
+          "Heap Corruption: Sizeclass of remote dealloc corrupt.");
         small_dealloc_offseted(
           super, slab, FreeObject::make(p_offseted), sizeclass);
       }
       else
       {
-        SNMALLOC_ASSERT(super->get_kind() == Medium);
+        auto medium = super.template as_reinterpret<Mediumslab>();
+        SNMALLOC_ASSERT(medium->get_kind() == Medium);
+        check_client(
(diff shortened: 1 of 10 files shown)


src/mem/remoteallocator.h
```diff
@@ -22,34 +24,39 @@ namespace snmalloc
...
 
-    alloc_id_t trunc_target_id()
+    /**
+     * Return allocator for this object.  This may perform amplification.
+     */
+    template<typename LargeAlloc>
+    static alloc_id_t trunc_target_id(CapPtr<Remote, CBAlloc> r, LargeAlloc* large_allocator)
     {
-      return alloc_id_and_sizeclass & ~SIZECLASS_MASK;
+      // Rederive allocator id.
```
nwf, line 51: This might be expensive on CHERI, where `capptr_amplify` has to do some maths and a load.  The reason for the `alloc_id_and_sizeclass` cache was so that we could get to the "Merely routing" branch without needing to `capptr_amplify`.  If it turns out that that is having real performance impact on non-CHERI, then probably we should either have the AAL manage the contents of `Remote`s or we should, here, switch on `aal_supports<StrictProvenance>` (or some other condition?).

mjp41: I tried to leave the relevant arguments around so that either could be done for this reason.  Yeah, just having an actual capability/pointer to the allocator seems a big security risk, so having an encoded version/hash or something would be better.

src/mem/alloc.h
```diff
@@ -948,7 +823,7 @@ namespace snmalloc
         return;
 
       stats().remote_post();
-      remote.post(&large_allocator, get_trunc_id());
+      remote.post<Allocator>(this, get_trunc_id());
```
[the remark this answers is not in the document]
mjp41: The problem is the type of `large_allocator` also depends on the template.  So I think what you are proposing could be done with two template parameters.  But then I worry that it would be harder to follow.  But I didn't like making it a friend.

src/mem/remoteallocator.h
nwf: I suppose I'm content to have this land and I'll investigate the performance impact on my CHERIfication branch.

mjp41: It now keeps both implementations, if an ifdef for which to use.

mjp41:
@nwf I added a couple more commits on to this, that I have been working on as part of the larger refactoring.

Issue #328, opened 2021-05-20
Compilation failure with glibc 2.17 - no `getentropy` available


mfelsche:
Hi, thanks first of all for this great allocator! :) We at tremor love it!

Yesterday we tried to compile it on our CI system to create a release and it failed.

The CI system is this slightly outdated centos docker image: https://github.com/tremor-rs/tremor-runtime/blob/main/packaging/builder-images/Dockerfile.x86_64-unknown-linux-gnu

It contains GCC 9.3.1 and glibc 2.17 and we use it to compile our Rust application including the `snmalloc-rs` binding which points to: https://github.com/microsoft/snmalloc/tree/e3a7eab78923e84cb2301797590583279880cf6c

Unfortunately this old glibc doesnt contain `getentropy` yet, which is included since 2.25.

it fails with the following error message:

```
 /root/.cargo/registry/src/github.com-1ecc6299db9ec823/snmalloc-sys-0.2.27/snmalloc/src/override/../mem/../mem/../pal/pal_posix.h:294:23: error: 'getentropy' was not declared in this scope; did you mean 'get_entropy64'?
    294 |         if (getentropy(&result, sizeof(result)) != 0)
        |             ~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~
        |             get_entropy64

```

I dont know if it makes sense to provide a fix for such an ancient libc, but wanted to report anyways.

---

I have attached a lengthy and verbose compilation log, maybe this will shed some light on what is happening:

```
error: failed to run custom build command for `snmalloc-sys v0.2.27`

Caused by:
  process didn't exit successfully: `/snmalloc_test/target/debug/build/snmalloc-sys-6b891a7b439974c7/build-script-build` (exit code: 101)
  --- stdout
[...]

PR #329, opened 2021-05-20
No getentropy on ancient libc

mfelsche:
Old libc versions (e.g. glibc < 2.25) do not provide a `getentropy` function.

As discussed in #328 I added a CMake compile check and exposed a flag `LIBC_HAS_GETENTROPY` and put the getentropy into `#ifdef`s, while also disabling the `Entropy` PAL feature.

Keeping this as a draft for now, as I didnt check on every platform. On a linux system with libc 2.17 it compiles fine with this change, where it previously broke.

Should I verify with some test that the PAL is actually providing some good entropy?


mjp41:
> Should I verify with some test that the PAL is actually providing some good entropy?

Don't worry about that, yet.  I haven't done that on any platforms.  This is really work in progress, so I will add testing for that at a later point.

mjp41:
@devnexen could you check you are happy with this?

CMakeLists.txt
```diff
@@ -34,6 +34,19 @@ size_t malloc_usable_size(const void* ptr) { return 0; }
 int main() { return 0; }
 " CONST_QUALIFIED_MALLOC_USABLE_SIZE)
 
+# older libcs might not have getentropy, e.g. it appeared in gliobc 2.25
+# so we need to fallback if we cannot compile this
+CHECK_C_SOURCE_COMPILES("
+#if __has_include(<sys/random.h>)
```
mjp41, line 40: Could you add a
```
#include <unistd.h>
```
I seem to remember some platforms include it in there.

mfelsche: good call! :) done that.

mjp41:
So it doesn't seem to find `getentropy` on Mac, but I thought it was using that previously.  I don't know how it worked if it didn't

https://dev.azure.com/snmalloc/snmalloc/_build/results?buildId=1533&view=logs&j=b19f5f93-56f8-5dd1-9094-af7dea069984&t=3c6ba7a3-8a56-50c1-9cac-eac388bd8352&l=25

mfelsche:
It seems it needs to include `<unistd.h>` before `sys/random.h`. Seems only to be an issue with my snippet.

mfelsche:
Yay!

```
Performing Test LIBC_HAS_GETENTROPY - Success
```

https://dev.azure.com/snmalloc/snmalloc/_build/results?buildId=1534&view=logs&jobId=b19f5f93-56f8-5dd1-9094-af7dea069984&j=b19f5f93-56f8-5dd1-9094-af7dea069984&t=3c6ba7a3-8a56-50c1-9cac-eac388bd8352

src/pal/pal_netbsd.h
```diff
@@ -35,12 +23,10 @@ namespace snmalloc
...
      *
-     * We disable the default implementation of randomness on NetBSD as it
-     * does not have the getentropy call. This will currently fallback to
-     * C++ libraries std::random_device.
+     * As NetBSD does not have the getentropy call, get_entropy64 will
+     * currently fallback to C++ libraries std::random_device.
      */
-    static constexpr uint64_t pal_features =
-      PALBSD_Aligned::pal_features & ~Entropy;
+    static constexpr uint64_t pal_features = PALBSD_Aligned::pal_featuresy;
```
devnexen, line 29: typo here `pal_featuresy` => `pal_features`, typical command line editor mistake most likely :)

devnexen:
overall looks good to me even tough less "elegant" than the current version but definitively safer.

src/pal/pal_netbsd.h
mfelsche: you got me.

mfelsche:
If there is anything more elegant you could think of to solve compilation on those older libcs, i am all ears :)
Supporting such a ton of platforms is really hard to 1.) get right and 2.) do elegantly, so i really admire how it is done here.

I still need to hear back from our legal team regarding the CLA, sorry for the delay here.

devnexen:
I honestly don t think you really can do better still readable enough :)

CMakeLists.txt
```diff
@@ -160,6 +176,10 @@ if(USE_POSIX_COMMIT_CHECKS)
   target_compile_definitions(snmalloc_lib INTERFACE -DUSE_POSIX_COMMIT_CHECKS)
 endif()
 
+if(LIBC_HAS_GETENTROPY)
+  target_compile_definitions(snmalloc_lib INTERFACE -DLIBC_HAS_GETENTROPY)
```
davidchisnall, line 180: Please prefix snmalloc macros with `SNMALLOC_`.  All of our macros appear in the global namespace for consumers and so shouldn't start with an identifier that isn't ours.

src/pal/pal_posix.h
```diff
@@ -117,9 +121,14 @@ namespace snmalloc
...
-     * POSIX systems are assumed to support lazy commit.
+     * POSIX systems are assumed to support lazy commit. The build system checks
+     * getentropy is available, only then this PAL supports Entropy.
      */
-    static constexpr uint64_t pal_features = LazyCommit | Entropy;
+    static constexpr uint64_t pal_features = LazyCommit
+#ifdef LIBC_HAS_GETENTROPY
+      | Entropy
+#endif
+      ;
```
davidchisnall, line 131: This could be simplified by defining the `SNMALLOC_PLATFORM_HAS_GETENTROPY` to either `0` or `Entropy` in the cmake.  Then this entire line becomes:

```suggestion
    static constexpr uint64_t pal_features = LazyCommit | SNMALLOC_PLATFORM_HAS_GETENTROPY;
```

That said, it makes me a little bit nervous to have a global configuration affecting code that's shared among platform configurations.  We support having multiple PAL instantiations in a single program and I can see some confusions arising from this, though those corner cases are not very likely and shouldn't block this.

src/pal/pal_posix.h
```diff
@@ -19,6 +19,10 @@
 #  include <unistd.h>
 #endif
 
+#ifndef LIBC_HAS_GETENTROPY
+#  include <cassert>
+#endif
```
davidchisnall, line 24: We aren't actually checking a dynamic condition here, so this should not be necessary.

src/pal/pal_posix.h
```diff
@@ -290,10 +299,15 @@ namespace snmalloc
...
+#ifdef LIBC_HAS_GETENTROPY
         uint64_t result;
         if (getentropy(&result, sizeof(result)) != 0)
           error("Failed to get system randomness");
         return result;
+#else
+        assert(0 && "Unreachable path");
+        return 0;
+#endif
       }
```
davidchisnall, line 311:
```suggestion
      else
      {
#ifdef LIBC_HAS_GETENTROPY
        uint64_t result;
        if (getentropy(&result, sizeof(result)) != 0)
          error("Failed to get system randomness");
        return result;
#endif
      }
      error("Entropy requested on platform that does not provide entropy");
```

mfelsche:
It seems the clang-tidy check is erroring on unrelated parts of the codebase. Should I be worried about them?

mjp41:
Is it broken because the Macro has not been defined for the clangtidy CI task?

https://github.com/microsoft/snmalloc/blob/e77a5d9c588171a3bcafce3f27b526bb06138e24/azure-pipelines.yml#L343-L345

Perhaps, we could add an 
```C++
#ifndef SNMALLOC_PLATFORM_HAS_GETENTROPY
```
to set a default, if the CMake has not given one.  This would be useful for projects not building using the CMake.

mfelsche:
CLA signed! :tada:

Anything else I can do to bring this forward? :)

mjp41 (approved):
LGTM.  If you are happy remove, can you remove the draft status.  

I'll give @devnexen and @davidchisnall a chance to comment, but I think you have address both of their existing comments.  So I'll merge if I don't get objections.

src/pal/pal_posix.h
```diff
@@ -19,6 +19,11 @@
 #  include <unistd.h>
 #endif
 
+// default value for builds not using CMake
+#ifndef SNMALLOC_PLATFORM_HAS_GETENTROPY
+#  define SNMALLOC_PLATFORM_HAS_GETENTROPY 0
+#endif
```
davidchisnall, line 25: It might be nice to add a `#warning` here, something like:
```suggestion
#ifndef SNMALLOC_PLATFORM_HAS_GETENTROPY
#  define SNMALLOC_PLATFORM_HAS_GETENTROPY 0
#  warning snmalloc not using entropy source. To silence this warning please define SNMALLOC_PLATFORM_HAS_GETENTROPY to 0, or to Entropy if your platform implements getentropy().
#endif
```

davidchisnall:
The net changes here look like less code and more comments, which is my favourite kind of commit!

davidchisnall:
Don't forget to remove the 'draft' from this, if you're happy with it.

mjp41:
Clang-tidy has broken due to the warning introduction.

PR #330, opened 2021-05-21
ptrwrap: hold a uintptr_t, not a T*, internally

nwf:
This should make it easier to work with CapPtr<>s in constexprs

Let's see what CI thinks of it.


src/ds/address.h
```diff
@@ -29,7 +29,7 @@ namespace snmalloc
   inline CapPtr<void, bounds>
   pointer_offset(CapPtr<T, bounds> base, size_t diff)
```
mjp41, line 30: Can these functions be made constexpr too?

src/ds/ptrwrap.h
```diff
@@ -101,14 +101,16 @@ namespace snmalloc
...
   {
-    T* unsafe_capptr;
+    uintptr_t unsafe_capptr;
 
     /**
      * nullptr is implicitly constructable at any bounds type
      */
-    CapPtr(const std::nullptr_t n) : unsafe_capptr(n) {}
+    constexpr SNMALLOC_FAST_PATH CapPtr(const std::nullptr_t n)
+    : unsafe_capptr(reinterpret_cast<uintptr_t>(n))
```
mjp41, line 110: I'm surprised this reinterpret_cast is allowed in a constexpr.

mjp41 (approved):
LGTM

src/ds/ptrwrap.h
nwf: So am I... oh.

This is C++ templating biting us once again: that `reinterpret_cast` isn't actually instantiated as a `constexpr` anywhere and so nobody notices that there's a spec violation lurking in the template AST.  Greeeeeat.

But that answers your other question: no, `pointer_offset` cannot be usefully made `constexpr` (though you'll get no errors if you do, it just won't be useful).

nwf:
Adjustment in light of my newfound `constexpr` understanding.  Please audit somewhat carefully, and let's see what CI thinks.

src/ds/ptrwrap.h
```diff
@@ -101,14 +101,20 @@ namespace snmalloc
   template<typename T, capptr_bounds bounds>
   struct CapPtr
   {
-    T* unsafe_capptr;
+    uintptr_t unsafe_capptr;
```
davidchisnall, line 104: I'm a little bit nervous about this from a non-CHERI codegen perspective.  Casting a pointer to and from an integer type adds it to the may-alias-anything set.  Did you check whether this affected codegen for x86?

PR #332, opened 2021-05-28
Gated the mcx16 flag to C++ compilation

ihaller:
This change enables the SNMALLOC_ONLY_HEADER_LIBRARY mode to work with projects that mix nasm assembly files with their C++ sources.


mjp41 (approved):
LGTM

PR #333, opened 2021-05-29
Fix still non getentropy platform builds

devnexen:


devnexen:
sorry here we go again but it fails on netbsd (at least).

mjp41:
@devnexen, sorry I am confused by your NetBSD comment.  Is this PR fixing NetBSD?  Or is NetBSD broken even with this change?

I prefer the guarded `| SNMALLOC_PLATFORM_HAS_GETENTROPY` you are proposing, but I don't see what this fixes? Could you give some extra guidance on your proposed change.

devnexen:
No pb :-) Yes it is for fixing NetBSD build, see SNMALLOC_PLATFORM_HAS_GETENTROPY was always defined regardless (`0` or `Entropy`) thus went back to square one. Hopefully it will solve for all :-)

CMakeLists.txt
```diff
@@ -178,8 +178,6 @@ endif()
 
 if(SNMALLOC_PLATFORM_HAS_GETENTROPY)
   target_compile_definitions(snmalloc_lib INTERFACE -DSNMALLOC_PLATFORM_HAS_GETENTROPY=Entropy)
```
mjp41, line 180:
```suggestion
  target_compile_definitions(snmalloc_lib INTERFACE -DSNMALLOC_PLATFORM_HAS_GETENTROPY)
```

src/pal/pal_posix.h
```diff
@@ -126,8 +120,11 @@ namespace snmalloc
      * POSIX systems are assumed to support lazy commit. The build system checks
      * getentropy is available, only then this PAL supports Entropy.
      */
-    static constexpr uint64_t pal_features =
-      LazyCommit | SNMALLOC_PLATFORM_HAS_GETENTROPY;
+    static constexpr uint64_t pal_features = LazyCommit
+#if defined(SNMALLOC_PLATFORM_HAS_GETENTROPY)
+      | SNMALLOC_PLATFORM_HAS_GETENTROPY
```
mjp41, line 125:
```suggestion
      | Entropy
```

mjp41 (changes requested):
Okay, I see
https://github.com/microsoft/snmalloc/blob/aefd149f0ace9307b2fef2e8d803472d2b5a3622/src/pal/pal_posix.h#L301-L306
is still being compiled causing the error.  I think given this change, I would sooner we not expose `Entropy` element of the `enum` to `CMake`.

src/pal/pal_posix.h
devnexen: ok makes sense was not sure if you wanted to keep `Entropy`

mjp41 (approved):
LGTM

PR #336, opened 2021-06-12
Rewrite Apple PAL using native APIs

amari:
This is a rewrite of the Apple PAL that implements the `AlignedAllocation`, `Entropy`, and `LazyCommit` features through native Apple APIs.

It adds a dependency on `Security.framework` via `SecRandomCopyBytes`. Apple actively discourages use of `getentropy` and the symbol is [not allowed on the App Store](https://developer.apple.com/forums/thread/675396).

# TODO / Discussion

## `USE_POSIX_COMMIT_CHECKS`
I removed support for `USE_POSIX_COMMIT_CHECKS` because the POSIX memory management APIs are no longer used but I'd like to avoid a regression. I need some context about what the `memset` and `mprotect` calls in `pal_bsd.h` and `pal_posix.h` are supposed to accomplish. Is it supposed to emulate Windows' decommit behavior? If so, why do the pages need to be decommitted? What are the invariants?

I hope to use the mach APIs to implement the behavior if possible and prevent any M1-related bugs:

- #278 
- #280 
- #284

## Can the `size` parameter of `reserve_aligned` be greater than `4GiB`?
There are Apple equivalents of `ReclaimVirtualMemory` and `OfferVirtualMemory`, but they only work on regions that are `4GiB` and smaller. (These APIs have existed since macOS 10.6 and are present in the iOS SDK.) 

The `4GiB` limit is a hard limit in the XNU kernel even on 64-bit platforms.

## ~~`LowMemoryNotification`~~

~~I have a working implementation behind the feature `SNMALLOC_APPLE_LOWMEMORYNOTIFICATION`. I'll push the change after we resolve `USE_POSIX_COMMIT_CHECKS`.~~

### DIFF
--- a/src/pal/pal.h
+++ b/src/pal/pal.h
@@ -120,11 +120,9 @@ namespace snmalloc
   // define `PAGE_SIZE` as a macro.  We don't use `PAGE_SIZE` as our variable
   // name, to avoid conflicts, but if we do see a macro definition then check
   // that our value matches the platform's expected value.
-#if defined(__APPLE__) && defined(PAGE_MAX_SIZE)
-  static_assert(
-    PAGE_MAX_SIZE == OS_PAGE_SIZE,
-    "Page size from system header does not match snmalloc config page size.");
-#elif defined(PAGE_SIZE)
+  // On macOS 11, system headers (mach/i386/vm_param.h and mach/arm/vm_param.h)
+  // define `PAGE_SIZE` as an extern making it unsuitable for this assertion.
+#if !defined(__APPLE__) && defined(PAGE_SIZE)
   static_assert(
     PAGE_SIZE == OS_PAGE_SIZE,
     "Page size from system header does not match snmalloc config page size.");
--- a/src/pal/pal.h
+++ b/src/pal/pal.h
@@ -120,7 +120,11 @@ namespace snmalloc
   // define `PAGE_SIZE` as a macro.  We don't use `PAGE_SIZE` as our variable
   // name, to avoid conflicts, but if we do see a macro definition then check
   // that our value matches the platform's expected value.
-#ifdef PAGE_SIZE
+#if defined(__APPLE__) && defined(PAGE_MAX_SIZE)
+  static_assert(
+    PAGE_MAX_SIZE == OS_PAGE_SIZE,
+    "Page size from system header does not match snmalloc config page size.");
+#elif defined(PAGE_SIZE)
   static_assert(
     PAGE_SIZE == OS_PAGE_SIZE,
     "Page size from system header does not match snmalloc config page size.");
--- a/src/pal/pal.h
+++ b/src/pal/pal.h
@@ -117,15 +117,18 @@ namespace snmalloc
     "The smallest architectural page size must divide OS_PAGE_SIZE");
 
   // Some system headers (e.g. Linux' sys/user.h, FreeBSD's machine/param.h)
-  // define `PAGE_SIZE` as a macro.  We don't use `PAGE_SIZE` as our variable
-  // name, to avoid conflicts, but if we do see a macro definition then check
-  // that our value matches the platform's expected value.
-  // On macOS 11, system headers (mach/i386/vm_param.h and mach/arm/vm_param.h)
-  // define `PAGE_SIZE` as an extern making it unsuitable for this assertion.
-#if !defined(__APPLE__) && defined(PAGE_SIZE)
+  // define `PAGE_SIZE` as a macro, while others (e.g. macOS 11's
+  // mach/machine/vm_param.h) define `PAGE_SIZE` as an extern. We don't use
+  // `PAGE_SIZE` as our variable name, to avoid conflicts, but if we do see a
+  // macro definition evaluates to a constant then check that our value matches
+  // the platform's expected value.
+#ifdef PAGE_SIZE
   static_assert(
-    PAGE_SIZE == OS_PAGE_SIZE,
+#  if __has_builtin(__builtin_constant_p)
+    !__builtin_constant_p(PAGE_SIZE) || (PAGE_SIZE == OS_PAGE_SIZE),
+#  else
+    true,
+#  endif
     "Page size from system header does not match snmalloc config page size.");
 #endif
-
 } // namespace snmalloc
--- a/src/pal/pal_apple.h
+++ b/src/pal/pal_apple.h
@@ -19,7 +19,7 @@ namespace snmalloc
   /**
    * PAL implementation for Apple systems (macOS, iOS, watchOS, tvOS...).
    */
-  template<int PALAnonID = PALAnonDefaultID>
+  template<uint8_t PALAnonID = PALAnonDefaultID>
   class PALApple : public PALBSD<PALApple<>>
   {
   public:
@@ -71,10 +71,12 @@ namespace snmalloc
      */
 
     // Encoded memory tag passed to `mmap`.
-    static constexpr int anonymous_memory_fd = VM_MAKE_TAG(PALAnonID);
+    static constexpr int anonymous_memory_fd =
+      int(VM_MAKE_TAG(uint32_t(PALAnonID)));
 
     // Encoded memory tag passed to `mach_vm_map`.
-    static constexpr int default_mach_vm_map_flags = VM_MAKE_TAG(PALAnonID);
+    static constexpr int default_mach_vm_map_flags =
+      int(VM_MAKE_TAG(uint32_t(PALAnonID)));
 
     /**
      * Notify platform that we will not be using these pages.
--- a/src/pal/pal_apple.h
+++ b/src/pal/pal_apple.h
@@ -5,6 +5,7 @@
 #  include "pal_bsd.h"
 
 #  include <CommonCrypto/CommonRandom.h>
+#  include <errno.h>
 #  include <mach/mach_init.h>
 #  include <mach/mach_vm.h>
 #  include <mach/vm_statistics.h>
@@ -93,7 +94,10 @@ namespace snmalloc
       // otherwise it's an error.
       //
       // `mach_vm_behavior_set` is observably slower in benchmarks.
-      madvise(p, size, MADV_FREE_REUSABLE);
+      //
+      // macOS 11 Big Sur may behave in an undocumented manner.
+      while (madvise(p, size, MADV_FREE_REUSABLE) == -1 && errno == EAGAIN)
+        ;
 
 #  ifdef USE_POSIX_COMMIT_CHECKS
       // This must occur after `MADV_FREE_REUSABLE`.
@@ -139,7 +143,10 @@ namespace snmalloc
       // otherwise it's an error.
       //
       // `mach_vm_behavior_set` is observably slower in benchmarks.
-      madvise(p, size, MADV_FREE_REUSE);
+      //
+      // macOS 11 Big Sur may behave in an undocumented manner.
+      while (madvise(p, size, MADV_FREE_REUSE) == -1 && errno == EAGAIN)
+        ;
 
       if constexpr (zero_mem == YesZero)
       {
--- a/src/pal/pal_apple.h
+++ b/src/pal/pal_apple.h
@@ -154,6 +154,9 @@ namespace snmalloc
       }
     }
 
+    // Apple's `mmap` doesn't support user-specified alignment and only
+    // guarantees mappings are aligned to the system page size, so we use
+    // `mach_vm_map` instead.
     template<bool committed>
     static void* reserve_aligned(size_t size) noexcept
     {
@@ -198,8 +201,26 @@ namespace snmalloc
     /**
      * Source of Entropy
      *
-     * Apple platforms do not have a public getentropy implementation, so use
-     * CCRandomGenerateBytes instead.
+     * Apple platforms have a working `getentropy(2)` implementation.
+     * However, it is not allowed on the App Store and Apple actively
+     * discourages its use. The substitutes `arc4random_buf(3)`,
+     * `CCRandomGenerateBytes`, and `SecRandomCopyBytes` are recommended
+     * instead.
+     *
+     * `CCRandomGenerateBytes` was selected because:
+     *
+     * 1. The implementation of `arc4random_buf(3)` differs from its
+     * documentation. It is documented to never fail, yet its'
+     * implementation can fail silently: it calls the function
+     * `ccrng_generate`, but ignores the error case.
+     * `CCRandomGenerateBytes` is built on the same function, but can return an
+     * error code in case of failure. See:
+     *      https://opensource.apple.com/source/Libc/Libc-1439.40.11/gen/FreeBSD/arc4random.c.auto.html
+     *      https://opensource.apple.com/source/CommonCrypto/CommonCrypto-60061/include/CommonRandom.h.auto.html
+     *
+     * 2. `SecRandomCopyBytes` introduces a dependency on `Security.framework`.
+     * `CCRandomGenerateBytes` introduces no new dependencies.
+     *
      */
     static uint64_t get_entropy64()
     {
--- a/src/pal/pal_apple.h
+++ b/src/pal/pal_apple.h
@@ -4,15 +4,14 @@
 
 #  include "../ds/address.h"
 
-#  include <Security/Security.h>
-#  include <errno.h>
+#  include <CommonCrypto/CommonRandom.h>
 #  include <execinfo.h>
 #  include <mach/mach_init.h>
 #  include <mach/mach_vm.h>
-#  include <mach/vm_region.h>
 #  include <mach/vm_statistics.h>
 #  include <mach/vm_types.h>
-#  include <utility>
+#  include <stdio.h>
+#  include <unistd.h>
 
 extern "C" int puts(const char* str);
 
@@ -170,17 +169,15 @@ namespace snmalloc
     /**
      * Source of Entropy
      *
-     * Apple platforms do not have a getentropy implementation, so use
-     * SecRandomCopyBytes instead.
+     * Apple platforms do not have a public getentropy implementation, so use
+     * CCRandomGenerateBytes instead.
      */
     static uint64_t get_entropy64()
     {
       uint64_t result;
       if (
-        SecRandomCopyBytes(
-          kSecRandomDefault,
-          sizeof(result),
-          reinterpret_cast<void*>(&result)) != errSecSuccess)
+        CCRandomGenerateBytes(
+          reinterpret_cast<void*>(&result), sizeof(result)) != kCCSuccess)
       {
         error("Failed to get system randomness");
       }
--- a/src/pal/pal_apple.h
+++ b/src/pal/pal_apple.h
@@ -5,7 +5,6 @@
 #  include "pal_bsd.h"
 
 #  include <CommonCrypto/CommonRandom.h>
-#  include <execinfo.h>
 #  include <mach/mach_init.h>
 #  include <mach/mach_vm.h>
 #  include <mach/vm_statistics.h>
@@ -14,8 +13,6 @@
 #  include <stdlib.h>
 #  include <sys/mman.h>
 #  include <unistd.h>
-
-extern "C" int puts(const char* str);
 
 namespace snmalloc
 {
@@ -39,17 +36,6 @@ namespace snmalloc
     static constexpr int anonymous_memory_fd = VM_MAKE_TAG(PALAnonID);
 
     static constexpr int default_mach_vm_map_flags = VM_MAKE_TAG(PALAnonID);
-
-    static void print_stack_trace()
-    {
-      constexpr int SIZE = 1024;
-      void* buffer[SIZE];
-      auto nptrs = backtrace(buffer, SIZE);
-      fflush(stdout);
-      backtrace_symbols_fd(buffer, nptrs, STDOUT_FILENO);
-      puts("");
-      fflush(stdout);
-    }
 
     /**
      * Notify platform that we will not be using these pages.
--- a/src/pal/pal_apple.h
+++ b/src/pal/pal_apple.h
@@ -156,8 +156,7 @@ namespace snmalloc
       // mask has least-significant bits set
       mach_vm_offset_t mask = size - 1;
 
-      int flags =
-        VM_FLAGS_ANYWHERE | VM_FLAGS_RANDOM_ADDR | default_mach_vm_map_flags;
+      int flags = VM_FLAGS_ANYWHERE | default_mach_vm_map_flags;
 
       // must be initialized to 0 or addr is interepreted as a lower-bound.
       mach_vm_address_t addr = 0;
--- a/src/pal/pal_apple.h
+++ b/src/pal/pal_apple.h
@@ -12,6 +12,7 @@
 #  include <mach/vm_statistics.h>
 #  include <mach/vm_types.h>
 #  include <stdio.h>
+#  include <stdlib.h>
 #  include <unistd.h>
 
 extern "C" int puts(const char* str);
--- a/src/pal/pal_apple.h
+++ b/src/pal/pal_apple.h
@@ -33,8 +33,47 @@ namespace snmalloc
 
     static constexpr size_t minimum_alloc_size = page_size;
 
+    /*
+     * Memory Tag
+     *
+     * A memory tag is an 8-bit value that denotes auxillary "type information"
+     * of a vm region. This tag can be used for marking memory for profiling and
+     * debugging, or instructing the kernel to perform tag-specific behavior.
+     * (E.g. VM_MEMORY_MALLOC{_*} is reused by default, unless it is no longer
+     * in its "original state". See `vm_map_entry_is_reusable` in
+     * `osfmk/vm/vm_map.c` for more details of this behavior.)
+     *
+     * Memory tags are encoded using `VM_MAKE_TAG(tag_value)`, and can be passed
+     * to the kernel by either `mmap` or `mach_vm_map`:
+     * 1. `fd` argument of `mmap`.
+     * 2. `flags` argument of `mach_vm_map`.
+     *
+     * There are currently 4 categories of memory tags:
+     *
+     * 1. Reserved: [0, 39]. Typically used for Apple libraries and services.
+     * Use may trigger undocumented kernel-based behavior.
+     *
+     * 2. Defined "placeholders": [39, 98]. Typically used for Apple libraries
+     * and services.
+     *
+     * 3. Undefined "placeholders": [99, 239]. Unallocated by Apple. Typically
+     * used for libraries. (E.g. LLVM sanitizers use 99.)
+     *
+     * 4. Application specific: [240, 255]
+     *
+     * See <mach/vm_statistics.h> for more details about memory tags and their
+     * uses.
+     *
+     * In the future, we may switch our default memory tag from "category 4" to
+     * "category 3", thereby affording us a "well-known" memory tag that can be
+     * easily identified in tools such as vmmap(1) or Instruments.
+     *
+     */
+
+    // Encoded memory tag passed to `mmap`.
     static constexpr int anonymous_memory_fd = VM_MAKE_TAG(PALAnonID);
 
+    // Encoded memory tag passed to `mach_vm_map`.
     static constexpr int default_mach_vm_map_flags = VM_MAKE_TAG(PALAnonID);
 
     /**
@@ -57,7 +96,7 @@ namespace snmalloc
 #  ifdef USE_POSIX_COMMIT_CHECKS
       // This must occur after `MADV_FREE_REUSABLE`.
       //
-      // `mach_protect` is observably slower in benchmarks.
+      // `mach_vm_protect` is observably slower in benchmarks.
       mprotect(p, size, PROT_NONE);
 #  endif
     }
--- a/src/pal/pal_apple.h
+++ b/src/pal/pal_apple.h
@@ -1,11 +1,20 @@
 #pragma once
 
 #ifdef __APPLE__
-#  include "pal_bsd.h"
 
+#  include "../ds/address.h"
+
+#  include <Security/Security.h>
 #  include <errno.h>
+#  include <execinfo.h>
+#  include <mach/mach_init.h>
+#  include <mach/mach_vm.h>
+#  include <mach/vm_region.h>
 #  include <mach/vm_statistics.h>
+#  include <mach/vm_types.h>
 #  include <utility>
+
+extern "C" int puts(const char* str);
 
 namespace snmalloc
 {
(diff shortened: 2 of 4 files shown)


mjp41:
@amari thanks for looking at this.  

> USE_POSIX_COMMIT_CHECKS

This was primarily introduced to stop people introducing Windows bugs on other platforms.  For memory not to be part of the Working Set (Resident set), we need to Decommit the reserved memory.  Any access to decommitted memory will cause a fault. Adding the `mprotect` was done to ensure any memory that is accessed has notified the PAL is will be using it.

For any memory access, the most recent call of `reserve`, `pal_notify_using` and `pal_notify_not_using` that covers that region of memory was either `pal_notify_using` or `reserve` with the `committed` flag set.

We do not require linearity, we can repeatedly call `pal_notify_using` or `pal_notify_not_using` and that is not considered a problem, i.e. the following is not considered an error:

````
  pal_notify_using(p, size);
  pal_notify_using(p, size);
```

I believe the code could be made to restrict this if required for the Apple PAL.  But it would require some additional work to validate all the paths, and add cross platform testing.

> Can the size parameter of reserve_aligned be greater than 4GiB?

Yes.  If someone asks snmalloc for more than 4GiB of memory, then we ask for more than 4GiB.

> LowMemoryNotification

Awesome


@davidchisnall, @ryancinsight, @devnexen  do you have any thoughts on moving away from the POSIX view of Apple to the native APIs?  This seems a good reason for the PAL, but do you have any compatibility concerns?

amari:
@mjp41. There was a formatting error. I don't think that the contributors were pinged.

devnexen:
Just tried your branch on Mac M1 and just does not build. Fails on

`pal.h: 125:5:static_assert expression PAGE_SIZE == OS_PAGE_SIZE ... #define PAGE_SIZE vm_page_size`

devnexen:
I m not against removing getentropy support here, seems justifiable to avoid the app store issue, however I m less keen about the vm api revamp. but we ll see I m not against neither.

amari:
@devnexen. Are you compiling using Rosetta? Otherwise `vm_page_size` should be `16KiB`. See [Determining CPU page size](https://www.wwdcnotes.com/notes/wwdc20/10214/).

devnexen:
Did not use rosetta. By the way I retried in my older mac intel same issue. Is there any prerequisite ?

amari:
I found this. I don't think externs play nice with `static_assert`. Before, it was comparing `vm_page_size` with itself and therefore the comparison optimized away.

> Declaration of `vm_page_size` in `/usr/include/mach/vm_page_size.h`
```
extern  vm_size_t       vm_page_size;
```

> Before macOS 11

`/usr/include/mach/i386/vm_param.h`
```
#define PAGE_SIZE  I386_PGBYTES // 4KiB
```
> macOS 11 and Later

`/usr/include/mach/i386/vm_param.h`
```
#define PAGE_SIZE  vm_page_size // extern
```
`/usr/include/mach/arm/vm_param.h`
```
#define PAGE_SIZE  vm_page_size // extern
```

Also, If we use `PAGE_SIZE` we lose any benefit of `page_size` being a constexpr. The replacement is `PAGE_MAX_SIZE`.

src/pal/pal.h
```diff
@@ -120,7 +120,11 @@ namespace snmalloc
   // define `PAGE_SIZE` as a macro.  We don't use `PAGE_SIZE` as our variable
   // name, to avoid conflicts, but if we do see a macro definition then check
   // that our value matches the platform's expected value.
-#ifdef PAGE_SIZE
+#if defined(__APPLE__) && defined(PAGE_MAX_SIZE)
+  static_assert(
+    PAGE_MAX_SIZE == OS_PAGE_SIZE,
+    "Page size from system header does not match snmalloc config page size.");
+#else if defined(PAGE_SIZE)
```
devnexen, line 127: elif

devnexen:
More important
`pal/pal.h:124:3: error: static_assert failed due to requirement '(1 << 14) == OS_PAGE_SIZE' "Page size from system header does not match snmalloc config page size."
`

devnexen:
Not sure for Intel about `PAGE_MAX_SIZE` tough, the build fails, my Mac M1 I can t use it now but should work.

amari:
> More important
> `pal/pal.h:124:3: error: static_assert failed due to requirement '(1 << 14) == OS_PAGE_SIZE' "Page size from system header does not match snmalloc config page size." `

Okay. Apple misrepesents the page size on macOS 11+. On x86_64 `PAGE_MAX_SIZE` is defined as `1 << 14`, which is simply untrue. The solution is to eliminate the `static_assert`. There's no way to get the page size at compile time.

devnexen:
Might be still needed for non apple tough ? Just curious is your gear the ARM64 flavor ?

devnexen:
> > More important
> > `pal/pal.h:124:3: error: static_assert failed due to requirement '(1 << 14) == OS_PAGE_SIZE' "Page size from system header does not match snmalloc config page size." `
> 
> @devnexen Okay. Apple misrepesents the page size on macOS 11+. On x86_64 `PAGE_MAX_SIZE` is defined as `1 << 14`, which is simply untrue. 

Or maybe PAGE_MAX_SIZE is not really what it appears.

amari:
> > > More important
> > > `pal/pal.h:124:3: error: static_assert failed due to requirement '(1 << 14) == OS_PAGE_SIZE' "Page size from system header does not match snmalloc config page size." `
> > 
> > 
> > @devnexen Okay. Apple misrepesents the page size on macOS 11+. On x86_64 `PAGE_MAX_SIZE` is defined as `1 << 14`, which is simply untrue.
> 
> Or maybe PAGE_MAX_SIZE is not really what it appears. May means more arch agnostic max value rather than arch specific, to be fair most of the time apple specifics in general are a bit of opaque :-)

I'm using Intel. Between 10.15 and 11, Apple changed the definition of PAGE_MAX_SHIFT from 12 to 14 for Intel builds. 

I think that the assertion should be skipped when `__APPLE__` is defined b/c there is only one kernel build, unlike Linux and the BSDs where the page size can be build dependent.

devnexen:
> > > > More important
> > > > `pal/pal.h:124:3: error: static_assert failed due to requirement '(1 << 14) == OS_PAGE_SIZE' "Page size from system header does not match snmalloc config page size." `
> > > 
> > > 
> > > @devnexen Okay. Apple misrepesents the page size on macOS 11+. On x86_64 `PAGE_MAX_SIZE` is defined as `1 << 14`, which is simply untrue.
> > 
> > 
> > Or maybe PAGE_MAX_SIZE is not really what it appears. May means more arch agnostic max value rather than arch specific, to be fair most of the time apple specifics in general are a bit of opaque :-)
> 
> I'm using Intel. Between 10.15 and 11, Apple changed the definition of PAGE_MAX_SHIFT from 12 to 14 for Intel builds.
> 
That s right, also same value for ARM thus my earlier assumption.
Last commit builds, and benchmarks outputs for master and this branch seems more or less on part.

ryancinsight:
> @amari thanks for looking at this.
> 
> > USE_POSIX_COMMIT_CHECKS
> 
> This was primarily introduced to stop people introducing Windows bugs on other platforms. For memory not to be part of the Working Set (Resident set), we need to Decommit the reserved memory. Any access to decommitted memory will cause a fault. Adding the `mprotect` was done to ensure any memory that is accessed has notified the PAL is will be using it.
> 
> For any memory access, the most recent call of `reserve`, `pal_notify_using` and `pal_notify_not_using` that covers that region of memory was either `pal_notify_using` or `reserve` with the `committed` flag set.
> 
> We do not require linearity, we can repeatedly call `pal_notify_using` or `pal_notify_not_using` and that is not considered a problem, i.e. the following is not considered an error:
> 
> ```
>   pal_notify_using(p, size);
>   pal_notify_using(p, size);
> ```
> 
> I believe the code could be made to restrict this if required for the Apple PAL. But it would require some additional work to validate all the paths, and add cross platform testing.
> 
> > Can the size parameter of reserve_aligned be greater than 4GiB?
> 
> Yes. If someone asks snmalloc for more than 4GiB of memory, then we ask for more than 4GiB.
> 
> > LowMemoryNotification
> 
> Awesome
> 
> @davidchisnall, @ryancinsight, @devnexen do you have any thoughts on moving away from the POSIX view of Apple to the native APIs? This seems a good reason for the PAL, but do you have any compatibility concerns?

I can't say I know enough about apple to forsee issues here, I don't currently use one either. That said, I would have thought arc4random_uniform/_buf would have been simpler but this route is pretty clever and I've see it pursued in swift. My only question is whether is an upperbound (only saw lower if there is one) to generated numbers or if there should be like there is in other implementations: https://opensource.apple.com/source/OpenSSH/OpenSSH-166/openssh/openbsd-compat/bsd-arc4random.c

src/pal/pal_apple.h
```diff
@@ -1,92 +1,192 @@
...
+    {
+      if (page_aligned || is_aligned_block<page_size>(p, size))
+      {
+        SNMALLOC_ASSERT(is_aligned_block<page_size>(p, size));
+
+        // mask has least-significant bits set
+        mach_vm_offset_t mask = page_size - 1;
+
+        int flags =
+          VM_FLAGS_FIXED | VM_FLAGS_OVERWRITE | VM_MAKE_TAG(PALAnonID);
```
devnexen, line 108: would putting back VM_MAKE_TAG into a default flag work for you ?

ryancinsight:
Also what's the benefit of this over CCrandomGeneratebytes? https://opensource.apple.com/source/CommonCrypto/CommonCrypto-60061/include/CommonRandom.h

amari:
> Also what's the benefit of this over CCrandomGeneratebytes? https://opensource.apple.com/source/CommonCrypto/CommonCrypto-60061/include/CommonRandom.h

## What does Apple say?

`SecRandomCopyBytes` is the only one of these functions that has documentation on Apple's developer site.

I followed the documentation rabbit-hole, and Apple's own CommonCrypto docs direct developers to use SecRandomCopyBytes.

[Cryptographic interfaces#Common Crypto library](https://developer.apple.com/security/) -> [Cryptographic Services Guide#GeneratingRandomNumbers](https://developer.apple.com/library/archive/documentation/Security/Conceptual/cryptoservices/RandomNumberGenerationAPIs/RandomNumberGenerationAPIs.html#//apple_ref/doc/uid/TP40011172-CH12-SW1) -> [Randomization Services Guide](https://developer.apple.com/documentation/security/randomization_services?language=objc) -> `SecRandomCopyBytes`

## Call Tracing

Did some digging on opensource.apple.com, I traced the calls:
`SecRandomCopyBytes(rnd, count, bytes)` -> `CCRandomCopyBytes(kCCRandomDefault, bytes, count)` -> `CCRandomGenerateBytes(bytes, count)`.
It looks like Apple ignores the RNG selection and uses the default.

## Wishful thinking
The most "correct" way to implement it would be to use corecrypto directly, but afaik it's an internal library:
```c

#include <corecrypto/ccrng.h> // ccrng, ccrng_state, ccrng_generate
 
static uint64_t get_entropy64() {
  uint64_t result;
  int status;

  struct ccrng_state *rng = ccrng(&status);

  if (status != 0 || ccrng_generate(rng, sizeof(result), reinterpret_cast<void*>(&result)) != CCERR_OK) {
  }
  
  return result;
}
```

## RNG specs

Here the specs of the RNG as of macOS 10.13:
https://github.com/apple/darwin-xnu/blob/main/EXTERNAL_HEADERS/corecrypto/ccrng.h#L28-L45

src/pal/pal_apple.h
amari: I'd rather leave it as an expression. If in the future Apple refactors VM_MAKE_TAG into a function or any other non constexpr it should still compile.

ryancinsight:
@amari thanks for explanation! these docs are definitely a rabbit hole! Based on trace results I personally still prefer to be as direct as possible especially if it's skipping rng:

#elif defined(__APPLE__)
#include <CommonCrypto/CommonRandom.h>

And somehow use CCRandomGenerateBytes(buf, buf_len) == kCCSuccess;
Which should be in the header above afaik

However, your example would indeed be better

amari:
> @amari thanks for explanation! these docs are definitely a rabbit hole! Based on trace results I personally still prefer to be as direct as possible especially if it's skipping rng:
> 
> #elif defined(**APPLE**)
> #include <CommonCrypto/CommonRandom.h>
> 
> And somehow use CCRandomGenerateBytes(buf, buf_len) == kCCSuccess;
> Which should be in the header above afaik
> 
> However, your example would indeed be better though very similar: https://opensource.apple.com/source/CommonCrypto/CommonCrypto-60061/include/CommonRandom.h.auto.html

I can't find any information about whether `CCRandomGenerateBytes` is allowed on the App Store. (As the saying goes, no news is good news.) If there's an issue in the future, `SNMALLOC_APPLE_APPSTORE_COMPAT` should be defined and use `SecCopyRandomBytes`.

ryancinsight:
They shouldn't give you a hard time based on this: https://www.niap-ccevs.org/Documents_and_Guidance/view_td.cfm?TD=0510

amari:
@devnexen "It works on my machine™️". Please test when you can.

devnexen:
> @devnexen "It works on my machine™️". 

The most feared sentence for QA ever :-)
Build fails because of `abort` otherwise.

devnexen:
build all good I notice tough a little performance decrease locally, total test take b/w 3/7 seconds more than master branch. do not know if this is more due to the crypto change.

mjp41:
> build all good I notice tough a little performance decrease locally, total test take b/w 3/7 seconds more than master branch. do not know if this is more due to the crypto change.

If the crypto is on the fast path, we have more performance issues.  If this is a lot of short tests, then I could imagine it would take longer.  I tend to use mimalloc-bench to give a rough idea of any regressions.

devnexen:
good point, seems appealing but also seems the microquill bench archive itself is not accessible.

amari:
@devnexen 

## The performance regression(s)

I've been benchmarking and digging through the XNU kernel on sourcegraph to find the performance regression(s). Here are my findings:

1. The Apple silicon/ARM bugs were due to `PALPOSIX::page_size` being defined as `Aal::smallest_page_size`, which is `0x1000` for ARM. It needs to be `16KiB` for Apple user-space on ARM.
2. `MADV_FREE` does not behave as expected. I'm not going to write the entire call tree, but the pages are never marked as "reusable" by the kernel and this can be observed through profiling. I tested `MADV_FREE` and `MADV_FREE_REUSABLE`. It is significant. At least ~75% to ~90% less dirty memory is used by `func-malloc-16` (observed on x86_64).
3. Unlike Linux, `madvise` and `mprotect` will never return `EAGAIN`.
4. `mach_vm_protect` and `mach_vm_behavior_set` are slower (100x) than `mprotect` and `madvise`. Apparently these functions pack their arguments into message buffers and use RPC to communicate with the kernel. (I had thought that Apple used registers for this.) This was slowing down `func-malloc-16` significantly.

## mimalloc-bench

I ran mimalloc-bench with leanN and redis. Performance is on-par with master.

> master
```
# --------------------------------------------------
# benchmark allocator elapsed rss user sys page-faults page-reclaims
      
leanN sn 34.81 466080 119.84 5.29 17 118441
      
redis sn 7.258 27696 3.47 0.17 0 7247
```
> pal-apple
```
# --------------------------------------------------
# benchmark allocator elapsed rss user sys page-faults page-reclaims
      
leanN sn 34.90 480648 119.34 5.18 20 122216
      
redis sn 7.109 27620 3.41 0.16 0 7226
```

## Where things stand

I refactored `PALApple` to inherit from `PALBSD`. However, the only shared code is `PALBSD::zero` and `PALPOSIX::error`. `notify_using`, `notify_not_using`, and `print_stack_trace` have Apple-specific definitions (x86_64 now calls `MADV_FREE_REUSABLE` due to reasons above). I also removed all mach_* calls except for the single `mach_vm_map` in `reserve_aligned`. I also removed the address randomization from `reserve_aligned`, which makes virtual memory profiles easier to compare.

ctest is a little slower on x86_64 (~5 to 10 seconds). This is due to replacing `MADV_FREE` with `MADV_FREE_REUSABLE`.

devnexen:
> @devnexen
> 
> ## The performance regression(s)
> I've been benchmarking and digging through the XNU kernel on sourcegraph to find the performance regression(s). Here are my findings:
> 

Nice rework
> 1. The Apple silicon/ARM bugs were due to `PALPOSIX::page_size` being defined as `Aal::smallest_page_size`, which is `0x1000` for ARM. It needs to be `16KiB` for Apple user-space on ARM.
> 2. `MADV_FREE` does not behave as expected. I'm not going to write the entire call tree, but the pages are never marked as "reusable" by the kernel and this can be observed through profiling. I tested `MADV_FREE` and `MADV_FREE_REUSABLE`. It is significant. At least ~75% to ~90% less dirty memory is used by `func-malloc-16` (observed on x86_64).
> 3. Unlike Linux, `madvise` and `mprotect` will never return `EAGAIN`.

Not sure about this statement, with madvise and MADV_FREE/MADV_FREE_REUSABLE it happened to me, rarely for sure tough.
> 4. `mach_vm_protect` and `mach_vm_behavior_set` are slower (100x) than `mprotect` and `madvise`. Apparently these functions pack their arguments into message buffers and use RPC to communicate with the kernel. (I had thought that Apple used registers for this.) This was slowing down `func-malloc-16` significantly.
> 

Really 100x ? Not fighting it really glad finally mmap calls are back :-)
> ## mimalloc-bench
> I ran mimalloc-bench with leanN and redis. Performance is on-par with master.
> 
> > master
> 
> ```
> # --------------------------------------------------
> # benchmark allocator elapsed rss user sys page-faults page-reclaims
>       
> leanN sn 34.81 466080 119.84 5.29 17 118441
>       
> redis sn 7.258 27696 3.47 0.17 0 7247
> ```
> 
> > pal-apple
> 
> ```
> # --------------------------------------------------
> # benchmark allocator elapsed rss user sys page-faults page-reclaims
>       
> leanN sn 34.90 480648 119.34 5.18 20 122216
>       
> redis sn 7.109 27620 3.41 0.16 0 7226
> ```
> 
> ## Where things stand
> I refactored `PALApple` to inherit from `PALBSD`. However, the only shared code is `PALBSD::zero` and `PALPOSIX::error`. `notify_using`, `notify_not_using`, and `print_stack_trace` have Apple-specific definitions (x86_64 now calls `MADV_FREE_REUSABLE` due to reasons above). I also removed all mach_* calls except for the single `mach_vm_map` in `reserve_aligned`. I also removed the address randomization from `reserve_aligned`, which makes virtual memory profiles easier to compare.
> 
> ctest is a little slower on x86_64 (~5 to 10 seconds). This is due to replacing `MADV_FREE` with `MADV_FREE_REUSABLE`.

Too bad microquill is down for now :-) I filled an issue on the project.

src/pal/pal_apple.h
```diff
@@ -18,75 +28,158 @@ namespace snmalloc
...
+      // must be initialized to 0 or addr is interepreted as a lower-bound.
+      mach_vm_address_t addr = 0;
+
+#  ifdef USE_POSIX_COMMIT_CHECKS
+      vm_prot_t prot = committed ? VM_PROT_READ | VM_PROT_WRITE : VM_PROT_NONE;
+#  else
+      vm_prot_t prot = VM_PROT_READ | VM_PROT_WRITE;
 #  endif
+
+      kern_return_t kr = mach_vm_map(
```
devnexen, line 144: does it still hold up with your recent changes ?

src/pal/pal_apple.h
```diff
@@ -18,75 +28,158 @@ namespace snmalloc
...
 
-    /**
-     * Note: The root's implementation works fine on Intel
-     * however mprotect/PROT_NONE fails on ARM
-     * especially since the 11.2 release (seems known issue
-     * spotted in various projects; might be a temporary fix).
-     */
-    template<bool page_aligned = false>
-    static void zero(void* p, size_t size) noexcept
+    static constexpr int default_mach_vm_map_flags = VM_MAKE_TAG(PALAnonID);
```
mjp41, line 41: This feels strange that an `fd` and `flag` are got from the same macro.

src/pal/pal_apple.h
amari: Yes. This is necessary b/c there's no equivalent BSD/POSIX syscall. I was aiming for correctness and future-proofing.

ctest and mimalloc-bench report similar performance as master.

mjp41 (changes requested):
Thanks for such an in depth analysis, and rounds of changes.  I am happy subject to a couple of minor things.

src/pal/pal_apple.h
devnexen: Ok fair enough, I was just thinking about perf again.

src/pal/pal_apple.h
devnexen: what s wrong with anonymous_memory_fd ?

src/pal/pal_apple.h
```diff
@@ -18,75 +28,158 @@ namespace snmalloc
...
+
+    static void print_stack_trace()
     {
-      if constexpr (Aal::aal_name != ARM)
-        PALBSD::zero(p, size);
-      else
-        ::bzero(p, size);
+      constexpr int SIZE = 1024;
+      void* buffer[SIZE];
+      auto nptrs = backtrace(buffer, SIZE);
```
devnexen, line 47: curiosity question, does it bring useful info ? backtrace on Linux is usually good however I have mostly lower experiences with other oses.

src/pal/pal_apple.h
amari: I used two values for correctness.  mmap is restricted to a subset of the flags that can be passed to mach_vm_map, therefore `anonymous_memory_fd` is a subset of `default_mach_vm_map_flags`. There have been alot of painful API breakages from Apple lately, and I want to make a possible future problem easier to debug/isolate.

src/pal/pal_apple.h
amari: I switched back the PALPOSIX's definition.

src/pal/pal_apple.h
```diff
@@ -18,75 +28,158 @@ namespace snmalloc
...
-    /**
-     * Anonymous page tag ID.
-     *
-     * Darwin platform allows to gives an ID to anonymous pages via
-     * the VM_MAKE_TAG's macro, from 240 up to 255 are guaranteed
-     * to be free of usage, however eventually a lower could be taken
-     * (e.g. LLVM sanitizers has 99) so we can monitor their states
-     * via vmmap for instance. This value is provided to `mmap` as the file
-     * descriptor for the mapping.
-     */
```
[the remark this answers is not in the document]
amari: I rewrote comment and elaborated about how VM_MAKE_TAG relates to mmap and mach_vm_map, and how to choose a value.

src/pal/pal_apple.h
devnexen: glad you did

devnexen:
I like the new changes minus would like to see back the EAGAIN loops.

amari:
@devnexen 
> I like the new changes minus would like to see back the EAGAIN loops.

I'm working on it. What OS version gave you EAGAIN? I going to look through sourcegraph to find any more undocumented errnos to be safe.

devnexen:
> @devnexen
> 
> > I like the new changes minus would like to see back the EAGAIN loops.
> 
> I'm working on it. What OS version gave you EAGAIN? I going to look through sourcegraph to find any more undocumented errnos to be safe.

Well macOs (never saw it anywhere else at least until now), it s pretty much contextual, depends of the current overall nature of the workload. since january I have my M1 maybe it occured 4/5 times.

amari:
> > @devnexen
> > > I like the new changes minus would like to see back the EAGAIN loops.
> > 
> > 
> > I'm working on it. What OS version gave you EAGAIN? I going to look through sourcegraph to find any more undocumented errnos to be safe.
> 
> Well macOs Big Sur (never saw it anywhere else at least until now), it s pretty much contextual, depends of the current overall nature of the workload. since january I have my M1 maybe it occured 4/5 times.

I tried to find a root cause. I couldn't find one. Oh well. The `EAGAIN` loops are back! ;)

devnexen:
The overall looks good I got nothing major else to say.

mjp41 (approved):
LGTM

mjp41:
@davidchisnall I am planning to merge this if you don't complain today ;-)

src/pal/pal_apple.h
```diff
@@ -1,92 +1,218 @@
...
-    static constexpr int anonymous_memory_fd = VM_MAKE_TAG(PALAnonID);
 
-    /**
-     * Note: The root's implementation works fine on Intel
-     * however mprotect/PROT_NONE fails on ARM
-     * especially since the 11.2 release (seems known issue
-     * spotted in various projects; might be a temporary fix).
-     */
-    template<bool page_aligned = false>
-    static void zero(void* p, size_t size) noexcept
```
davidchisnall, line 47: We seem to have lost this function.  @mjp41, is it no longer needed?  If so, we should update the porting document to remove it and remove it from all of the other PALs.

src/pal/pal_apple.h
```diff
@@ -1,92 +1,218 @@
...
+#  endif
     }
 
     /**
-     * same remark as above but we need to mark the page as REUSE
-     * first
+     * Notify platform that we will be using these pages.
      */
     template<ZeroMem zero_mem>
     static void notify_using(void* p, size_t size) noexcept
```
davidchisnall, line 114: This function now looks almost identical to the POSIX PAL version.  Why do we need a separate one?

src/pal/pal_apple.h
davidchisnall: Please can you add a comment here explaining that this call is used because `mmap` doesn't provide the alignment.  @staceyson, I take it that `MAP_ALIGNED` is still not there on new XNU?

src/pal/pal_apple.h
mjp41: It is coming from the PALBSD one, but I think we should keep this version.  Good catch

src/pal/pal_apple.h
```diff
@@ -1,92 +1,218 @@
...
+        error("Failed to allocate memory\n");
+      }
+
+      return reinterpret_cast<void*>(addr);
+    }
+
+    /**
+     * Source of Entropy
+     *
+     * Apple platforms do not have a public getentropy implementation, so use
```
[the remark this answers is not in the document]
amari: `arc4random_buf` can silently fail. `getentropy` and `CCGenerateRandomBytes` both return an error code.

https://opensource.apple.com/source/Libc/Libc-1439.40.11/gen/FreeBSD/arc4random.c.auto.html
```c
void
arc4random_buf(void *buf, size_t buf_size)
{
	arc4_init();
        // the error is ignored
	ccrng_generate(rng, buf_size, buf);
}
```

https://opensource.apple.com/source/CommonCrypto/CommonCrypto-60178.40.2/lib/CommonRandom.c.auto.html
```c
CCRNGStatus CCRandomGenerateBytes(void *bytes, size_t count)
{
    int err;
    struct ccrng_state *rng;

    if (0 == count) {
        return kCCSuccess;
    }

    if (NULL == bytes) {
        return kCCParamError;
    }

    rng = ccDRBGGetRngState();
    // the error is captured
    err = ccrng_generate(rng, count, bytes);
    if (err == CCERR_OK) {
        return kCCSuccess;
    }

    return kCCRNGFailure;
}
```

I found an app that links against `arc4random` on the App Store.

src/pal/pal_apple.h
```diff
@@ -1,92 +1,218 @@
...
+     * Notify platform that we will be using these pages.
      */
     template<ZeroMem zero_mem>
     static void notify_using(void* p, size_t size) noexcept
     {
       SNMALLOC_ASSERT(
-        is_aligned_block<PALBSD::page_size>(p, size) || (zero_mem == NoZero));
+        is_aligned_block<page_size>(p, size) || (zero_mem == NoZero));
+
+      if constexpr (zero_mem == YesZero)
```
davidchisnall, line 119: Please put this logic in `zero()`

src/pal/pal_apple.h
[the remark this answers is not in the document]
amari: I'll correct the comment so that it is more clear and concise.

Also, `arc4random` is not seeded with `genentropy`:

https://opensource.apple.com/source/Libc/Libc-1439.40.11/gen/FreeBSD/arc4random.c.auto.html
```c
static struct ccrng_state *rng;

static void
arc4_init(void)
{
	int err;

	if (rng != NULL) return;

	rng = ccrng(&err);
	if (rng == NULL) {
#if OS_CRASH_ENABLE_EXPERIMENTAL_LIBTRACE
		os_crash("arc4random: unable to get ccrng() handle (%d)", err);
#else
		os_crash("arc4random: unable to get ccrng() handle");
#endif
	}
}

uint32_t
arc4random(void)
{
	uint32_t rand;

	arc4random_buf(&rand, sizeof(rand));

	return rand;
}

void
arc4random_buf(void *buf, size_t buf_size)
{
	arc4_init();
	ccrng_generate(rng, buf_size, buf);
}
```

src/pal/pal_apple.h
```diff
@@ -1,92 +1,218 @@
...
+      // `mach_vm_behavior_set` is observably slower in benchmarks.
+      //
+      // macOS 11 Big Sur may behave in an undocumented manner.
       while (madvise(p, size, MADV_FREE_REUSE) == -1 && errno == EAGAIN)
         ;
 
       if constexpr (zero_mem == YesZero)
-        zero<true>(p, size);
+      {
+        ::bzero(p, size);
```
davidchisnall, line 153: We can only reach this if the `mmap` call above failed.  I don't think that `mmap` should be able to fail, so the error condition there should probably be `PAL::error` (which kills the process): if we can't `mmap` over some address space that we've previously allocated, then something is badly wrong and we're probably trying to `mmap` over some address space that we shouldn't be trying to `mmap` over so `bzero`ing it will make things even worse.

src/pal/pal_apple.h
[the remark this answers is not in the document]
amari: The dependency is CommonCrypto. CoreCrypto is private. Yes, `ccrng_generate` is a private function.

CommonCrypto (`/usr/lib/system/libcommonCrypto.dylib`) is implicitly linked by `dyld` (macOS linker.) We have no say in the matter as library developers. If there's any "library initialization" that is dependent on malloc in CommonCrypto, I'd argue that the behavior is undefined and is intrinsic to macOS in production.

I'd argue that Apple makes no guarantees about future behavior of standard system APIs, as we saw with the macOS 11.2 release (a minor release!) that broke mprotect (W^X) and many projects experienced issues.

The descriptions of ccrng and ccrng_generate:

https://github.com/apple/darwin-xnu/blob/main/EXTERNAL_HEADERS/corecrypto/ccrng.h
```c
/*!
 @function   ccrng
 @abstract   Initializes an AES-CTR mode cryptographic random number generator and returns the statically-allocated rng object.
             Getting a pointer to a ccrng has never been simpler!
             Call this function, get an rng object and then pass the object to ccrng_generate() to generate randoms.
             ccrng() may be called more than once. It returns pointer to the same object on all calls.
 @result  a cryptographically secure random number generator or NULL if fails
 @discussion
 - It is significantly faster than using the system /dev/random
 - FIPS Compliant: NIST SP800-90A + FIPS 140-2
 - Seeded from the system entropy.
 - Provides at least 128bit security if the system provide 2bit of entropy / byte.
 - Entropy accumulation
 - Backtracing resistance
 - Prediction break with frequent (asynchronous) reseed
 */
```
```c
/*!
 @function   ccrng_generate
 @abstract   Generate `outlen` bytes of output, stored in `out`, using ccrng_state `rng`.
 @param rng  `struct ccrng_state` representing the state of the RNG.
 @param outlen  Amount of random bytes to generate.
 @param out  Pointer to memory where random bytes are stored, of size at least `outlen`.
 @result 0 on success and nonzero on failure.
 */
```

ccrng "Initializes an AES-CTR mode cryptographic random number generator and returns the **statically-allocated rng object** [emphasis mine]."

Standards-compliant entities/organizations deploy to macOS and iOS using this API. See, https://www.niap-ccevs.org/Documents_and_Guidance/view_td.cfm?TD=0510

src/pal/pal_apple.h
amari: PALPOSIX behaves in the same fashion. PALWindows calls `error` if `VirtualAlloc2FromApp` fails.

I completely agree with your comment, but I wanted to keep the inheritance from PALBSD and PALPOSIX, therefore I duplicated the behavior.

Also, Apple is now using undocumented errnos (e.g. mprotect can return EAGAIN on some production builds of macOS). So, I can't say for certain that a failed `MAP_FIXED` can occur only in fatal failure cases. Nor can I say that in the future `MADV_FREE_REUSABLE` will never cause `MAP_FIXED` to fail.

src/pal/pal_apple.h
amari: @devnexen, who opened #278, tested this branch on both Intel and M1 machines and found no regression.

About a week ago, I crawled XNU on sourcegraph in an attempt to diagnose root cause(s) for the M1 bugs, as well as reconcile the difference between the x86_64 and ARM codepaths of PALApple:

1. The Apple silicon/ARM bugs were due to PALPOSIX::page_size being defined as Aal::smallest_page_size, which is 0x1000 for ARM. It needs to be 16KiB for Apple user-space on ARM. There's quite a bit of "creative" engineering going on in XNU as of macOS 11.3 . On ARM, the kernel uses 4KiB pages internally, but requires userspace to use 16KiB pages, unless the process is "exotic" (i.e. rosetta).
2. MADV_FREE does not behave as expected. I'm not going to write the entire call tree, but the pages are never marked as "reusable" by the kernel and this can be observed through profiling. I tested MADV_FREE and MADV_FREE_REUSABLE. It is significant. At least ~75% to ~90% less dirty memory is used by func-malloc-16 (observed on x86_64). Apple's own malloc implementation as well as many ports for Apple Operating Systems use MADV_FREE_REUS{E, ABLE} instead of MADV_FREE. See https://bugs.chromium.org/p/chromium/issues/detail?id=713892   https://opensource.apple.com/source/libmalloc/libmalloc-53.1.1/src/magazine_malloc.c.auto.html


---

The reason why `notify_using` and `zero` look like duplicate functionality is b/c of optimization. Also related to another one of @davidchisnall comments below, **when `PALPOSIX::zero` calls `mmap`, it can fail and call `bzero` on the same pages that mmap failed to overwrite**.

In pseudocode, for brevity.

master:
```
PALPOSIX::zero<page_aligned>:
        if page_aligned || is_aligned_block<OS::page_size>(p, size)
            mmap
            if r != MAP_FAILED
                return;
    bzero

PALPOSIX::notify_using:
    madvise(MADV_FREE)
    mprotect(PROT_READ|PROT_WRITE)
    zero<true>
```

pal-apple:
```
PALApple::notify_using:
        mmap
        if r != MAP_FAILED
            return;
    mprotect(PROT_READ|PROT_WRITE)
    madvise(MADV_FREE_REUSE)
    bzero
```

On macOS (and iOS) we are free to move the mmap before mprotect, thereby eliminating the other 2 syscalls. Otherwise it would be `mprotect`, `mmap`, which is wasteful.

src/pal/pal_apple.h
amari: `MADV_FREE` != `MADV_FREE_REUSABLE` on XNU.
Please see the above comment. https://github.com/microsoft/snmalloc/pull/336#discussion_r655723362

src/pal/pal_apple.h
amari: I didn't want to deviate from PALPOSIX::notify_using's current behavior. `PALPOSIX::notify_using` will call `PALPOSIX::zero`. Currently if `mmap` fails, `bzero` is called on the pages that mmap failed to overwrite.

I wanted to maintain identical failure conditions as PALPOSIX.

amari: I feel that this is more of an issue of `PALPOSIX::notify_using`'s behavior. I too would like PALPOSIX and PALApple to abort the current process, like PALWindows. But I feel that belongs in a different pull request; one that will also update `PORTING.md` to include the invariant "failed `mmap` calls will cause the process to abort" or a something similar.

src/pal/pal_apple.h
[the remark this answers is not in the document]
amari: `ccrng` is closed source. `ccrng_generate` calls the function pointer in `ccrng_state`.

src/pal/pal_apple.h
amari: Please see https://github.com/microsoft/snmalloc/pull/336#discussion_r655723362

src/pal/pal.h
```diff
@@ -120,7 +120,9 @@ namespace snmalloc
   // define `PAGE_SIZE` as a macro.  We don't use `PAGE_SIZE` as our variable
   // name, to avoid conflicts, but if we do see a macro definition then check
   // that our value matches the platform's expected value.
-#ifdef PAGE_SIZE
+  // On macOS 11, system headers (mach/i386/vm_param.h and mach/arm/vm_param.h)
+  // define `PAGE_SIZE` as an extern making it unsuitable for this assertion.
+#if !defined(__APPLE__) && defined(PAGE_SIZE)
   static_assert(
     PAGE_SIZE == OS_PAGE_SIZE,
```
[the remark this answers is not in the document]
amari: Isn't `__builtin_constant_p` always `0` on DEBUG builds due to building with `-o0`?

src/pal/pal_apple.h
mjp41: That seems reasonable to move that to a different PR.

src/pal/pal_apple.h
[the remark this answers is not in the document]
mjp41: > If there is an error, we are killing the process.

I would much sooner kill the process.  We build different versions of the library with and without the mitigations.  If it fails with mitigations, then the user can try the version without mitigations.  Silent failure seems like a problem for security auditing.

src/pal/pal_apple.h
davidchisnall: Thanks, that makes a lot of sense.  It would be great to have a comment explaining this.

src/pal/pal_apple.h
davidchisnall: Thanks.  We already have some template parameters for the generic PALs that allow tweaking things like the default `mmap` flags, I was wondering why this couldn't be done to make `MADV_FREE` something tunable, but it looks as if the order of operations is sufficiently different that this isn't possible (or, rather, would add a lot more complexity than having a custom implementation here).

davidchisnall (approved):

amari:
@mjp41 `LowMemoryNotification` will be a different PR.

The `perf-low_memory_*` tests pass, but will sometimes get `SIGKILL`'d due to memory usage even though I can see the reclamation happen on the dirty and swap graphs of the profiler.
macOS has 3 memory pressure states: Normal, Warning, and Critical. Not only is a process is picked to be killed based on memory usage, but also the number of times it has triggered the Warning and/or Critical states. There are sysctl knobs that govern this behavior, but they are undocumented so I still need to experiment.

mjp41 (approved):

mjp41:
> @mjp41 `LowMemoryNotification` will be a different PR.
> 
> The `perf-low_memory-*` tests pass, but will sometimes get `SIGKILL`'d due to memory usage even though I can see the reclamation happen on the dirty and swap graphs of the profiler.
> macOS has 3 memory pressure states: Normal, Warning, and Critical. Not only is a process is picked to be killed based on memory usage, but also the number of times it has triggered the Warning and/or Critical states. The current test in `low_memory.cc` calls `reach_pressure` then `reduce_pressure` 10 times in a loop. There are sysctl knobs that govern this behavior, but they are undocumented so I still need to experiment.

Perfect thanks.  We are happy to alter the test to trigger it fewer times on Apply platforms if that helps.

PR #341, opened 2021-06-30
Fix the build with GCC and libc cleanup.

davidchisnall:
GCC is a lot more picky about extern "C" definitions in namespaces than
clang and so the `friend` declaration wasn't picked up by the version of
the function declared in the `snmalloc` namespace.  Move it out to where
GCC is happy with it.


mjp41 (approved):
LGTM

PR #342, opened 2021-07-05
Allow disabling the Debug-mode check in the ABA code

ihaller:
The check relies on a thread_local variable which might not be available when using SlowAllocator on custom platforms.


mjp41 (approved):
LGTM

PR #343, opened 2021-07-07
Major refactor of snmalloc

mjp41:
This is a giant refactor.  This is targeting a separate dev branch `snmalloc2`, where we can stage and develop the remaining features.  

The key difference is the pagemap design. At the end of the PR description, there is a nomenclature that covers terms used in the PR description.

# Pagemap

The Pagemap now stores all the meta-data for the object allocation.  The meta-data in the pagemap is effectively a triple of the sizeclass, the remote allocator, and a pointer to a 64 byte block of meta-data for this chunk of memory.  By storing the pointer to a block, it allows the pagemap to handle multiple slab sizes without branching on the fast path.  There is one entry in the pagemap per 16KiB of address space, but by using the same entry in the pagemap for 4 adjacent entries, then we can treat a 64KiB range can be treated as a single slab of allocations.

This change also means there is almost no capability amplification required by the implementation on CHERI for finding meta-data.  The only amplification is required, when we change the way a chunk is used to a size of object allocation.

# Backend

There is a second major aspect of the refactor that there is now a narrow API that abstracts the Pagemap, PAL and address space management. This should better enable the compartmentalisation and makes it easier to produce alternative backends for various research directions.  This is a template parameter that can be used to specialised by the front-end in different ways.

# Thread local state

The thread local state has been refactored into two components, one (called 'localalloc') that is stored directly in the TLS and is constant initialised, and one that is allocated in the address space (called 'coreallloc') which is lazily created and pooled. 

# Difference

This removes Superslabs/Medium slabs as there meta-data is now part of the pagemap.

# Nomenclature
- Chunk: a 2^n sized and aligned block of memory
- Slab: a chunk that is used to supply multiple allocations of the same size
- object allocation: an allocation of any size
- small allocation: an allocation that is from a slab.
- large allocation: an allocation that is backed by a single chunk.


mjp41:
Things not currently done.  I'll create an issue to track these at some point, probably once I have had feedback on this.

- [ ] Decommit - notify_using/notify_notusing
- [ ] Distinction between PAL and Backend.
- [ ] Low memory notification
- [ ] CHERI support
- [ ] Windows perf - currently windows implementation will perform commit a lot, this needs improving for perf on Windows.
- [ ] Some performance regressions.

mjp41:
This branch now has a passing CI.  I think this is reasonably stable, and I plan to move to a PR based process of continuing work.  As so much as changed, I don't see a line by line review as making sense.  This PR is mainly for getting CI up and running, and a point at which I believe the code is stable enough for others to try and contribute fixes.

mjp41:
# Pagemap

The Pagemap now stores all the meta-data for the object allocation. The meta-data in the pagemap is effectively a triple of the sizeclass, the remote allocator, and a pointer to a 64 byte block of meta-data for this chunk of memory. By storing the pointer to a block, it allows the pagemap to handle multiple slab sizes without branching on the fast path. There is one entry in the pagemap per 16KiB of address space, but by using the same entry in the pagemap for 4 adjacent entries, then we can treat a 64KiB range can be treated as a single slab of allocations.

This change also means there is almost no capability amplification required by the implementation on CHERI for finding meta-data. The only amplification is required, when we change the way a chunk is used to a size of object allocation.

# Backend

There is a second major aspect of the refactor that there is now a narrow API that abstracts the Pagemap, PAL and address space management. This should better enable the compartmentalisation and makes it easier to produce alternative backends for various research directions. This is a template parameter that can be used to specialised by the front-end in different ways.

# Thread local state

The thread local state has been refactored into two components, one (called 'localalloc') that is stored directly in the TLS and is constant initialised, and one that is allocated in the address space (called 'coreallloc') which is lazily created and pooled.

# Difference

This removes Superslabs/Medium slabs as there meta-data is now part of the pagemap.

PR #345, opened 2021-07-13
pagemap: don't eagerly commit on LazyCommit PALs

nwf:
This dramatically improves the runtime of at least `perf-external_pointer` tests, which call `get` with `potentially_out_of_range = true` quite a lot, at least on my Linux box.


mjp41 (approved):
LGTM

PR #346, opened 2021-07-13
Improved OEPal integration with the new snmalloc architecture

ihaller:
PalNoAlloc forwards the Entropy setting of the underlying PAL as well as using its zeroing function.
Removed the unused parts of the PalOpenEnclave after the refactor.
`PalOpenEnclave::setup_initial_range` should be replace by:
`snmalloc::FixedGlobals fixed_handle;
fixed_handle.init(snmalloc::CapPtr<void, snmalloc::CBChunk>(heap_start), heap_size);`

### DIFF
--- a/src/pal/pal_noalloc.h
+++ b/src/pal/pal_noalloc.h
@@ -17,17 +17,18 @@ namespace snmalloc
    * address-space manager is initialised with all of the memory that it will
    * ever use.
    *
-   * It takes an error handler delegate as a template argument. This is
+   * It takes an underlying PAL delegate as a template argument. This is
    * expected to forward to the default PAL in most cases.
    */
-  template<typename ErrorHandler>
+  template<SNMALLOC_CONCEPT(ConceptPAL) UnderlyingPAL>
   struct PALNoAlloc
   {
     /**
      * Bitmap of PalFeatures flags indicating the optional features that this
      * PAL supports.
      */
-    static constexpr uint64_t pal_features = NoAllocation;
+    static constexpr uint64_t pal_features = NoAllocation |
+      ((UnderlyingPAL::pal_features & Entropy) != 0 ? Entropy : 0);
 
     static constexpr size_t page_size = Aal::smallest_page_size;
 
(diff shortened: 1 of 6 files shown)


src/pal/pal_noalloc.h
```diff
@@ -17,17 +17,18 @@ namespace snmalloc
    * address-space manager is initialised with all of the memory that it will
    * ever use.
    *
-   * It takes an error handler delegate as a template argument. This is
+   * It takes an underlying PAL delegate as a template argument. This is
    * expected to forward to the default PAL in most cases.
    */
-  template<typename ErrorHandler>
+  template<SNMALLOC_CONCEPT(ConceptPAL) UnderlyingPAL>
```
davidchisnall, line 23: Please don't make this require the PAL concept.  This does not need to implement all of the features of a PAL, it needs only to provide error handling code.  In my existing use, the error handler is significantly simpler than a full PAL.  In particular, it can't allocate or deallocate memory.

src/pal/pal_noalloc.h
```diff
@@ -17,17 +17,18 @@ namespace snmalloc
...
   struct PALNoAlloc
   {
     /**
      * Bitmap of PalFeatures flags indicating the optional features that this
      * PAL supports.
      */
-    static constexpr uint64_t pal_features = NoAllocation;
+    static constexpr uint64_t pal_features = NoAllocation |
```
davidchisnall, line 30: Requiring `pal_features` from the error handler is not ideal.  Can we add an additional optional template parameter for the entropy source, if we want to provide one?  Alternatively, a subclass of this could provide an entropy source if necessary.

mjp41 (approved):

src/pal/pal_noalloc.h
ihaller: I agree that that I should just go back to a typename, but for the OpenEnclave platform it is not just the error handling that needs delegation, but also the zeroing interface and the entropy generation. The latter is also not needed if the UnderlyingPAL does not specify Entropy as an available feature.

ihaller: I like the idea of having other PALs replace this one and not fit in as template arguments, but then we need to add a PAL template to FixedGlobals which defaults to Pal. Does that sound good @mjp41 ?

mjp41 (approved):

davidchisnall (approved):

PR #347, opened 2021-07-13
Cleaner implementation of signed pointers.

mjp41:
This encodes a back pointer in each node.  The back pointer is stored
in an encoded form so that it is hard to corrupt and trick the allocator
into following incorrect pointers.

This changes the encoding from previously being a Feistel network on
the next pointer that was using the prev as part of the key, to now
effectively using a doubly linked queue, where the back pointers are
scrambled, so it is hard to forge them.

This has the positive effects of
 - Not needing to store previous while building the list, as the append
   nows, curr and next at the point of writing into next, and does not
   need an additional previous.
 - The encoding is not affecting the actual next value, so more
   instructions can be executed in parallel by the CPU.

Future extensions, store a changing key in the FreeListBuilder so it
becomes harder to try to forge the previous token.

This approach can also be applied to the remote list, and will in a
subsequent PR.  This enables the idea to be tested.

### DIFF
--- a/src/mem/freelist.h
+++ b/src/mem/freelist.h
@@ -21,7 +21,7 @@
  *
  * If f just returns the first parameter, then this degenerates to a doubly
  * linked list.  (Note that doing the degenerate case can be useful for
- * debugging snmalloc bugs.) By making it a function of both objects, it
+ * debugging snmalloc bugs.) By making it a function of both pointers, it
  * makes it harder for an adversary to mutate prev_encoded to a valid value.
  *
  * This provides protection against the free-list being corrupted by memory
@@ -93,10 +93,13 @@ namespace snmalloc
     }
 
     /**
+     * Assign next_object and update its prev_encoded if CHECK_CLIENT.
+     *
      * Static so that it can be used on reference to a FreeObject.
      *
-     * Returns the next_object field of the next parameter, so can be used to
-     * iterate the construction.
+     * Returns a pointer to the next_object field of the next parameter as an
+     * optimization for repeated snoc operations (in which
+     * next->next_object is nullptr).
      */
     static CapPtr<FreeObject, CBAlloc>* store(
       CapPtr<FreeObject, CBAlloc>* curr,
@@ -299,7 +302,7 @@ namespace snmalloc
         if (end[1] != &head[1])
         {
           // The start token has been corrupted.
-          // TOUTOC issue, but small window here.
+          // TOCTTOU issue, but small window here.
           head[1]->check_prev(get_fake_signed_prev(1, entropy));
 
           terminate_list(1, entropy);
--- a/src/mem/freelist.h
+++ b/src/mem/freelist.h
@@ -1,7 +1,36 @@
 #pragma once
 /**
  * This file encapsulates the in disused object free lists
- * that are used per slab of small objects.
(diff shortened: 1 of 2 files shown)


src/mem/freelist.h
```diff
@@ -121,11 +93,42 @@ namespace snmalloc
...
 
     /**
-     * Read the next pointer handling any required decoding of the pointer
+     * Static so that it can be used on reference to a FreeObject.
+     *
+     * Returns the next_object field of the next parameter, so can be used to
+     * iterate the construction.
      */
-    CapPtr<FreeObject, CBAlloc> read_next(uint32_t key, LocalEntropy& entropy)
+    static CapPtr<FreeObject, CBAlloc>* store(
```
nwf, line 101:
```suggestion
    static CapPtr<FreeObject, CBAlloc>* store_next(
```

src/mem/freelist.h
```diff
@@ -121,11 +93,42 @@ namespace snmalloc
     }
 
     /**
-     * Read the next pointer handling any required decoding of the pointer
+     * Static so that it can be used on reference to a FreeObject.
+     *
+     * Returns the next_object field of the next parameter, so can be used to
+     * iterate the construction.
```
nwf, line 99:
```suggestion
     * Returns a pointer to the next_object field of the next parameter as an
     * optimization for repeated snoc operations (in which
     * next->next_object is nullptr).
```

src/mem/freelist.h
```diff
@@ -121,11 +93,42 @@ namespace snmalloc
     }
 
     /**
-     * Read the next pointer handling any required decoding of the pointer
+     * Static so that it can be used on reference to a FreeObject.
```
nwf, line 96:
```suggestion
     * Assign next_object and update its prev_encoded if CHECK_CLIENT.
     *
     * Static so that it can be used on reference to a FreeObject.
```

src/mem/freelist.h
```diff
@@ -283,63 +251,18 @@ namespace snmalloc
...
-          local_curr = address_cast(next) & 0xffff'ffff;
-          count++;
-          iter = &next->next_object;
-        }
-      }
-      return count;
+      end[index] = FreeObject::store(end[index], n, entropy);
     }
 
     /**
```
nwf, line 268: Annoyingly, I can't comment on lines not considered part of the context diff.  In any case, just below this text field, the comment says "Termination uses the bottom bit" which I think is no longer true?

src/mem/freelist.h
```diff
@@ -383,62 +298,29 @@ namespace snmalloc
         // If second list is non-empty, perform append.
         if (end[1] != &head[1])
         {
+          // The start token has been corrupted.
+          // TOUTOC issue, but small window here.
```
nwf, line 302:
```suggestion
          // TOCTTOU issue, but small window here.
```
is the more traditional spelling, I think?

nwf (commented):
Nitpicks in prose, but :+1: on code.

src/mem/freelist.h
mjp41: I agree with this change, but I have another PR to follow this one, which I'll apply this in.

src/mem/freelist.h
mjp41: Thanks, good catch

PR #349, opened 2021-07-16
Refactor pagemap and ASM relationship

mjp41:
The primary aim of this PR is to be able to enforce the invariant that all memory in the address space manager has committed pages for the corresponding entries in the pagemap.  This will improve windows support, and enable turning off pages inside the pagemap that are no currently required. 

To support this, I have made all Pals that perform allocation provide `reserve_at_least`.

### DIFF
--- a/src/backend/backend.h
+++ b/src/backend/backend.h
@@ -11,7 +11,7 @@ namespace snmalloc
    * This class implements the standard backend for handling allocations.
    * It abstracts page table management and address space management.
    */
-  template<typename PAL, bool fixed_range>
+  template<SNMALLOC_CONCEPT(ConceptPAL) PAL, bool fixed_range>
   class BackendAllocator
   {
   public:
--- a/src/backend/backend.h
+++ b/src/backend/backend.h
@@ -50,9 +50,13 @@ namespace snmalloc
       FlatPagemap<MIN_CHUNK_BITS, MetaEntry, PAL, fixed_range> pagemap;
 
     public:
-      void init()
+      template<bool fixed_range_ = fixed_range>
+      std::enable_if_t<!fixed_range_> init()
       {
-        pagemap.init(&address_space);
+        static_assert(
+          fixed_range_ == fixed_range, "Don't set SFINAE parameter!");
+
+        pagemap.init();
 
         if constexpr (fixed_range)
         {
@@ -60,10 +64,15 @@ namespace snmalloc
         }
       }
 
-      void init(CapPtr<void, CBChunk> base, size_t length)
+      template<bool fixed_range_ = fixed_range>
+      std::enable_if_t<fixed_range_>
+      init(CapPtr<void, CBChunk> base, size_t length)
       {
-        address_space.add_range(base, length);
-        pagemap.init(&address_space, address_cast(base), length);
+        static_assert(
+          fixed_range_ == fixed_range, "Don't set SFINAE parameter!");
+
+        auto [heap_base, heap_length] = pagemap.init(base, length);
+        address_space.add_range(heap_base, heap_length);
 
         if constexpr (!fixed_range)
         {
--- a/src/backend/backend.h
+++ b/src/backend/backend.h
@@ -65,14 +65,13 @@ namespace snmalloc
       }
 
       template<bool fixed_range_ = fixed_range>
-      std::enable_if_t<fixed_range_>
-      init(CapPtr<void, CBChunk> base, size_t length)
+      std::enable_if_t<fixed_range_> init(void* base, size_t length)
       {
         static_assert(
           fixed_range_ == fixed_range, "Don't set SFINAE parameter!");
 
         auto [heap_base, heap_length] = pagemap.init(base, length);
-        address_space.add_range(heap_base, heap_length);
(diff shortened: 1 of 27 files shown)


mjp41:
I have added some more faetures to this PR.  It now switched on the POSIX commit checks for the CHECK_CLIENT builds.  This enables the guard page like behaviour for things we are not currently using, and is a pre-req for protecting meta-data.

Additionally, in the CHECK_CLIENT mode it now over-allocates the pagemap, so that the location of the pagemap relative to other state is randomised.

src/backend/pagemap.h
```diff
@@ -117,15 +121,32 @@ namespace snmalloc
...
+      // Allocate a power of two extra to allow the placement of the
+      // pagemap be difficult to guess.
+      size_t additional_size = bits::next_pow2(REQUIRED_SIZE) * 2;
+      size_t request_size = REQUIRED_SIZE + additional_size;
+#else
+      size_t request_size = REQUIRED_SIZE;
+#endif
 
-      auto new_body = reinterpret_cast<T*>(new_body_untyped);
+      auto new_body_untyped = Pal::reserve(request_size);
```
ihaller, line 135: I am not sure if these should not be PAL instead of Pal. If the former then maybe it could be considered to differentiate the naming a bit more (maybe a refactor to remove the usage of Pal in other places in favour of DefaultPal?)

mjp41: Good catch. I will Remove `Pal` in a subsequent PR.

src/backend/backend.h
```diff
@@ -50,20 +50,28 @@ namespace snmalloc
       FlatPagemap<MIN_CHUNK_BITS, MetaEntry, PAL, fixed_range> pagemap;
 
     public:
-      void init()
+      template<bool fixed_range_ = fixed_range>
+      std::enable_if_t<!fixed_range_> init()
```
nwf, line 54:
```suggestion
      std::enable_if_t<!fixed_range_, void> init()
```

src/backend/backend.h
```diff
@@ -50,20 +50,28 @@ namespace snmalloc
...
 
         if constexpr (fixed_range)
         {
           abort();
         }
       }
 
-      void init(CapPtr<void, CBChunk> base, size_t length)
+      template<bool fixed_range_ = fixed_range>
+      std::enable_if_t<fixed_range_> init(void* base, size_t length)
```
nwf, line 68:
```suggestion
      std::enable_if_t<fixed_range_, void> init(void* base, size_t length)
```

src/backend/pagemap.h
```diff
@@ -19,84 +20,143 @@ namespace snmalloc
...
     address_t base{0};
     size_t size{0};
 
+  public:
     /**
-     * Commit entry
+     * Ensure this range of pagemap is accessible
      */
-    void commit_entry(void* p)
+    void register_range(address_t p, size_t length)
```
nwf, line 54: Should this be gated, either with `enable_if` or at all call sites, on `!pal_supports<LazyCommit>`?  Somewhat relatedly, do we ever `notify_not_using` any part of the pagemap?

src/backend/pagemap.h
```diff
@@ -19,84 +20,143 @@ namespace snmalloc
...
+        base == bits::align_up(base, bits::one_at_bit(GRANULARITY_BITS)));
+
+      return {heap_base, pointer_diff(heap_base, end)};
+    }
+
+    /**
+     * Initialise the pagemap without bounds.
+     */
+    template<bool has_bounds_ = has_bounds>
+    std::enable_if_t<!has_bounds_> init()
```
nwf, line 116:
```suggestion
    std::enable_if_t<!has_bounds_, void> init()
```

src/backend/pagemap.h
```diff
@@ -19,84 +20,143 @@ namespace snmalloc
...
+      static_assert(bits::is_pow2(sizeof(T)), "Next line assumes this.");
+      size_t offset = get_entropy64<PAL>() & (additional_size - sizeof(T));
+      auto new_body =
+        reinterpret_cast<T*>(pointer_offset(new_body_untyped, offset));
+#else
+      auto new_body = reinterpret_cast<T*>(new_body_untyped);
+#endif
+      // Ensure bottom page is committed
+      // ASSUME: new memory is zeroed.
+      PAL::template notify_using<NoZero>(
```
nwf, line 153: Again, `LazyCommit`?

nwf (commented):
Some nits, but otherwise LGTM

nwf:
Whoops, took too long, sorry.  Hopefully still useful comments for the future.

src/backend/pagemap.h
mjp41: So I am not sure about that.  On POSIX, with `SNMALLOC_CHECK_CLIENT`, we are using `PROT_NONE` a lot, so the `using` and `not_using` is required.  I could change `LazyCommit` to be off for POSIX with `SNMALLOC_CHECK_CLIENT`.  Then I could potentially gate these things.  At the moment, I am just hoping the empty bodied functions get inlined, and erased to not cost when we are exploiting `LazyCommit.`

PR #350, opened 2021-07-17
Don't depend on <entropy> if we don't need it.

davidchisnall:
Including <entropy> brings in <iostream>, which breaks the in-libc
build.


mjp41 (approved):
LGTM.  This will still pull in `<random>` elsewhere when unnecessary, but we can fix when someone needs a more minimal includes for those platforms.

PR #352, opened 2021-07-20
MetaEntry vs nullptr UB

nwf:
It is UB to offset from `nullptr` (except perhaps with a 0 offset).  Apparently clang is able to use this to reason, given `void* p`, that comparing `__builtin_align_down(p, x)` against `handle.fake_large_remote` (i.e., a `static inline constexpr` `nullptr`) must be the same as comparing `p` itself against `nullptr`.

In `MetaEntry`'s constructor, converting the provided `RemoteAllocator*` to `uintptr_t` before offsetting avoids the UB.  (While here, don't use `address_cast()`, as `address_t` will, on CHERI, be `ptraddr_t` and not `uintptr_t`.)

Doing the alignment in `get_remote` at `uintptr_t` before casting to `RemoteAllocator*`, rather than converting and then aligning, prevents the reasoning above from eliminating the alignment.

### DIFF
--- a/src/ds/address.h
+++ b/src/ds/address.h
@@ -17,11 +17,20 @@ namespace snmalloc
   using address_t = Aal::address_t;
 
   /**
+   * Perform arithmetic on a uintptr_t.
+   */
+  inline uintptr_t pointer_offset(uintptr_t base, size_t diff)
+  {
+    return base + diff;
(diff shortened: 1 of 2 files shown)


mjp41 (approved):
LGTM

PR #354, opened 2021-07-21
Add protection to meta-data

mjp41:
The meta-data in the `CHECK_CLIENT` mode is allocated from its own areas,
where most of the pages have been disabled, and the location within this
range is randomised.

This protects from trivial OOB read/write hitting the meta-data.
More complex controlled offsets are also mitigated due to the sparse
level of pages being turned on (i.e. not `PROT_NONE`).

### DIFF
--- a/src/backend/backend.h
+++ b/src/backend/backend.h
@@ -14,6 +14,19 @@ namespace snmalloc
   template<SNMALLOC_CONCEPT(ConceptPAL) PAL, bool fixed_range>
   class BackendAllocator
   {
+    // Size of local address space requests.  Currently aimed at 2MiB large
+    // pages but should make this configurable (i.e. for OE, so we don't need as
+    // much space).
+    constexpr static size_t LOCAL_CACHE_BLOCK = bits::one_at_bit(21);
+
+#ifdef SNMALLOC_CHECK_CLIENT
+    // When protecting the meta-data, we use a smaller block for the meta-data
+    // that is randomised inside a larger block.  This needs to be at least a
+    // page so that we can use guard pages.
+    constexpr static size_t LOCAL_CACHE_META_BLOCK =
+      bits::max(MIN_CHUNK_SIZE * 2, OS_PAGE_SIZE);
+#endif
+
   public:
     using Pal = PAL;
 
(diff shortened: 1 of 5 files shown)


src/backend/backend.h
```diff
@@ -89,41 +108,92 @@ namespace snmalloc
...
-            local_state->local_address_space.template commit_block<PAL>(
-              p, size);
+          return p;
         }
+
+        auto refill =
+          global.template reserve<false>(LOCAL_CACHE_BLOCK, h.pagemap);
+        if (refill == nullptr)
+          return nullptr;
+        auto refill_size = LOCAL_CACHE_BLOCK;
```
ihaller, line 140: Maybe consider moving refilll_size above the reservation and using it as the argument, but this is a small nitpick.

src/backend/backend.h
```diff
@@ -89,41 +108,92 @@ namespace snmalloc
...
+
+        // Find a suitable offset into the block
+        size_t offset_mask = size_request - rsize;
+        size_t offset;
+        do
+        {
+          offset = get_entropy64<PAL>() & offset_mask;
+        } while ((offset == 0) || (offset == offset_mask));
+        p = pointer_offset(p, offset);
+        PAL::template notify_using<NoZero>(p.unsafe_ptr(), rsize);
```
ihaller, line 186: Is the reason for using notify_using directly as opposed to global.commit_block because p and rsize are guaranteed to be page aligned?

src/backend/backend.h
```diff
@@ -89,41 +108,92 @@ namespace snmalloc
...
+      // During start up we need meta-data before we have a local allocator
+      // This code protects that meta-data with randomisation, and guard pages.
+      if (local_state == nullptr && is_meta)
       {
-        auto& a = h.address_space;
-        p = a.template reserve_with_left_over<true>(size, h.pagemap);
+        size_t rsize = bits::max(OS_PAGE_SIZE, bits::next_pow2(size));
+        size_t size_request = rsize * 64;
+
+        p = global.template reserve<false>(size_request, h.pagemap);
```
ihaller, line 174: Would it make sense to create a method for the pattern of using a larger allocation from an address space to create a randomized inner object? It is only used in two places, but would make it easier to make changes uniformly later on.

ihaller:
LGTM

src/backend/backend.h
mjp41: Done.

src/backend/backend.h
mjp41: Yeah, I have made it use the PAL directly in both places now, as it is always big enough.  (Until @nwf tells me I have broken his 64KiB page PowerPC machine again.)

PR #356, opened 2021-07-23
Make address space manager use pagemap for next pointers

mjp41:
The address space manager uses the pagemap entry to form linked lists of
unused address space above MIN_CHUNK_SIZE.  It continues to use
references in the block below that threshold.

In the CHECK_CLIENT mode this makes it hard to corrupt the ASM as only
meta-data uses allocations below MIN_CHUNK_SIZE from the ASM. This
allocations will be protected with guard pages by the backend.

### DIFF
--- a/src/backend/address_space_core.h
+++ b/src/backend/address_space_core.h
@@ -20,6 +20,11 @@ namespace snmalloc
    */
   class AddressSpaceManagerCore
   {
+    struct FreeChunk
+    {
+      CapPtr<FreeChunk, CBChunk> next;
+    };
+
     /**
      * Stores the blocks of address space
      *
@@ -35,12 +40,12 @@ namespace snmalloc
      * bits::BITS is used for simplicity, we do not use below the pointer size,
      * and large entries will be unlikely to be supported by the platform.
      */
-    std::array<CapPtr<void, CBChunk>, bits::BITS> ranges = {};
+    std::array<CapPtr<FreeChunk, CBChunk>, bits::BITS> ranges = {};
 
     /**
      * Checks a block satisfies its invariant.
      */
-    inline void check_block(CapPtr<void, CBChunk> base, size_t align_bits)
+    inline void check_block(CapPtr<FreeChunk, CBChunk> base, size_t align_bits)
     {
       SNMALLOC_ASSERT(
         base == pointer_align_up(base, bits::one_at_bit(align_bits)));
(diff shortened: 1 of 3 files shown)


src/backend/address_space_core.h
```diff
@@ -26,22 +23,19 @@ namespace snmalloc
...
-     * of size 2^n.
-     *
-     * The second entry ranges[n][1] is a pointer to a linked list of blocks
-     * of this size. The final block in the list is not committed, so we commit
-     * on pop for this corner case.
+     * The array indexes based on power of two size.
      *
-     * Invariants
-     *  ranges[n][1] != nullptr => ranges[n][0] != nullptr
+     * The entries for each size form a linked list.  For sizes below,
```
nwf, line 28:
```suggestion
     * The entries for each size form a linked list.  For sizes below
```

src/backend/address_space_core.h
```diff
@@ -57,45 +51,77 @@ namespace snmalloc
...
     {
-      check_block(base, align_bits);
-      SNMALLOC_ASSERT(align_bits < 64);
-      if (ranges[align_bits][0] == nullptr)
+      if (align_bits >= MIN_CHUNK_BITS)
       {
-        // Prefer first slot if available.
-        ranges[align_bits][0] = base;
+        // The pagemap stores MetaEntrys, abuse the metaslab field to be the
+        // next block in the stack of blocks.
```
nwf, line 72:
```suggestion
        // next block in the stack of blocks.
        //
        // The pagemap entries here have nullptr (i.e., fake_large_remote) as
        // their remote, and so other accesses to the pagemap (by external_pointer,
        // for example) will not attempt to follow this "Metaslab" pointer.
```

Is that right?

src/backend/address_space_core.h
```diff
@@ -57,45 +51,77 @@ namespace snmalloc
...
+        MetaEntry t(reinterpret_cast<Metaslab*>(next.unsafe_ptr()), nullptr, 0);
+        pagemap.set(address_cast(base), t);
         return;
       }
 
-      if (ranges[align_bits][1] != nullptr)
+      *(reinterpret_cast<CapPtr<void, CBChunk>*>(base.unsafe_ptr())) = next;
```
nwf, line 78: I fiddled a bit with adding a `struct FreeChunk { CapPtr<FreeChunk, CBChunk> next; };`, which renders this line `base->next = next;`, which is great, and similarly for line 100 below, which just becomes `return base->next;`, but there get to be a lot of casts elsewhere (if you're curious, I've [attached a diff](https://github.com/microsoft/snmalloc/files/6868415/snmalloc-address_space_core-FreeChunk.txt)).  As somewhat sketchy as this line feels, it's probably fine.

mjp41: That looks much nicer

src/backend/address_space_core.h
mjp41: Yes, but also this is memory that is completely controlled by snmalloc, so I don't think anyone should query it with `external_pointer`.  It is not going to be a stack, global, mmap that is done elsewhere in the program.  Though, not entirely sure what the correct full spec of `external_pointer` is for that argument.

PR #357, opened 2021-07-23
Correct REMOTE_MIN_ALIGN

nwf:
Take the maximum of...

* CACHELINE_SIZE (for performance)

* next_pow2(NUM_SIZECLASSES + 1) so that, when the pagemap points to a Remote,
  the (small) size class stuffed in the bottom bits can be removed by alignment

* next_pow2(NUM_LARGE_CLASSES + 1) so that, when the pagemap isn't pointing to
  a Remote, when the associated chunk is (part of) a large allocation, aligning
  the Remote* results in 0.

The last of these conditions will almost never be the deciding factor, as there
are generally many more small size classes than large ones, but it shouldn't
hurt to be safe.

### DIFF
--- a/src/mem/remoteallocator.h
+++ b/src/mem/remoteallocator.h
@@ -10,10 +10,19 @@
 
 namespace snmalloc
 {
-  // Remotes need to be aligned enough that all the
-  // small size classes can fit in the bottom bits.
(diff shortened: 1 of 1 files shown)


mjp41 (approved):
LGTM, good catch.

PR #358, opened 2021-07-26
XOR encoded next_object

mjp41:
This commit adds a simple XOR encoding to the next_object pointer in
FreeObjects.  This removes the trivial way of getting hold of a physical
address from the system by observing the free list pointers in
deallocated objects.

### DIFF
--- a/src/mem/freelist.h
+++ b/src/mem/freelist.h
@@ -41,19 +41,6 @@
 
 namespace snmalloc
 {
-  struct FreeListKey
-  {
-    address_t key;
-
-    FreeListKey(uint64_t key_)
-    {
-      if constexpr (bits::BITS == 64)
-        key = static_cast<address_t>(key_);
-      else
-        key = key_ & 0xffff'ffff;
-    }
-  };
-
   /**
    * This function is used to sign back pointers in the free list.
    *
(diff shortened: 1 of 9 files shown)


src/mem/freelist.h
```diff
@@ -103,6 +90,22 @@ namespace snmalloc
...
+    /**
+     * Encode next
+     */
+    inline static CapPtr<FreeObject, CBAlloc> encode_next(
+      address_t curr, CapPtr<FreeObject, CBAlloc> next, const FreeListKey& key)
+    {
+      UNUSED(curr); // Used for alternative encoding schemes.
+#ifdef SNMALLOC_CHECK_CLIENT
+      // TODO don't do this on CHERI.
+      return CapPtr<FreeObject, CBAlloc>(address_cast(next) ^ key.key_next);
```
nwf, line 102: I think that "don't do this on CHERI" can be pronounced as...

```suggestion
      if constexpr (aal_supports<StrictProvenance>)
      {
        UNUSED(key);
        return next;
      }
      return CapPtr<FreeObject, CBAlloc>(address_cast(next) ^ key.key_next);
```

mjp41: Thanks. I change that slightly, as it is typically best to use an `else` with `constexpr`.

nwf (approved):

PR #359, opened 2021-07-28
Snmalloc2 API cleanups for sandbox use.

davidchisnall:
This is the set of changes required for snmalloc2 to be usable by the
process sandboxing code and incorporates some API changes that reduce
the amount of code required to embed snmalloc.  Highlights:

 - The existing code expects to have an instance of the Globals object
   but this is only ever used as a handle for the type.  There is no
   mechanism for providing a different globals object to different
   consumers (except by providing a different type) and so all methods
   are purely stateless and refer either to static fields or to fields
   of a singleton.  This mechanism does not benefit from lazy static
   initialisation because there is a separate mechanism for lazy
   initialisation.  This version removes any instantiation of the
   globals class and makes all state per-type, not per instance.  All
   methods on the globals class are now static and we no longer thread
   an empty object through a load of methods internally.
 - Not every configuration needs to use the lazy initialisation hooks.
   They now need to be provided only if they are used.  If the
   configuration does not provide an `ensure_init` method, it is not
   called.  If it does not provide an `is_initialised` method then the
   global initialisation state is not checked.
 - There is now an `snmalloc::Options` class that default initialises
   itself to the default behaviour.  Every configuration must provide a
   `constexpr` instance of this class.  Each flag can be separately
   overridden and new flags can be added without breaking any existing
   API consumers.


src/mem/localalloc.h
```diff
@@ -90,40 +100,53 @@ namespace snmalloc
...
-      // After performing underlying operation, in the case of teardown already
-      // having begun, we must flush any state we just acquired.
-      if (post_teardown)
+      if constexpr (!SharedStateHandle::Options.LocalAllocSupportsLazyInit)
+      {
+        return action(core_alloc, args...);
+      }
+      else
       {
+        // Initialise the thread local allocator
```
ihaller, line 118: Is my understanding correct that this code-path gets executed on every allocation and deallocation when the CoreAllocOwnsLocalState is disabled? Since core_alloc will stay nullptr and cannot be set externally as far as I can tell? Am I missing something here?

ihaller (commented):
Best to wait until mjp41 also has a chance to look at it, but overall it looks good with a few comments.

src/snmalloc.h
```diff
@@ -7,10 +7,12 @@
 #include "mem/globalconfig.h"
 
 // The default configuration for snmalloc
+#ifndef SNMALLOC_PROVIDE_OWN_CONFIG
```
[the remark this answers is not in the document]
davidchisnall: This is what I'm doing.  Now that everything is templated nicely, there's no need to define things in the middle of the snmalloc includes: you can include everything then specify your config and then the type for it.  This is only necessary if you want to replace the *default* config, which is necessary only for using the override things.

src/mem/localalloc.h
```diff
@@ -90,40 +100,53 @@ namespace snmalloc
...
+        if constexpr (SharedStateHandle::Options.CoreAllocOwnsLocalState)
```
nwf, line 119: Can you expand on the rationale for this being conditional?  What sets `core_alloc` if this is false?

nwf (changes requested):

src/mem/corealloc.h
```diff
@@ -506,6 +591,24 @@ namespace snmalloc
       return small_alloc_slow<zero_mem>(sizeclass, fast_free_list, rsize);
     }
 
+    /**
+     * Accessor for the local state.  This hides whether the local state is
+     * stored inline or provided externally from the rest of the code.
+     */
+    SNMALLOC_FAST_PATH
+    LocalState& get_local_state()
```
mjp41, line 599: This name is not descriptive enough, as `local_state` makes no reference to this being the `backend`'s state.  Can you change this to `get_backend_local_state`, and ensure the comments mention `backend`.

src/mem/fixedglobalconfig.h
```diff
@@ -30,32 +30,17 @@ namespace snmalloc
       return backend_state;
     }
 
-    ChunkAllocatorState& get_slab_allocator_state()
+    static ChunkAllocatorState& get_slab_allocator_state(void*)
```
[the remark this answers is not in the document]
mjp41: Is this parameter required because you have remove the dynamic instance of `handle`?

src/mem/globalalloc.h
```diff
@@ -6,55 +6,56 @@
 namespace snmalloc
 {
   template<class SharedStateHandle>
-  inline static void aggregate_stats(SharedStateHandle handle, Stats& stats)
+  inline static void aggregate_stats(Stats& stats)
   {
-    auto* alloc = Pool<CoreAllocator<SharedStateHandle>>::iterate(handle);
+    auto* alloc = Pool<CoreAllocator<SharedStateHandle>>::template iterate<
```
mjp41, line 11: Should we have some `static_assert`s about `SharedStateHandle::Options.CoreAllocIsPoolAllocated` in this file, so that incorrect usage has a better error message?

src/mem/localalloc.h
```diff
@@ -90,40 +100,53 @@ namespace snmalloc
...
-
-      // Perform underlying operation
-      auto r = action(core_alloc, args...);
-
-      // After performing underlying operation, in the case of teardown already
-      // having begun, we must flush any state we just acquired.
-      if (post_teardown)
+      if constexpr (!SharedStateHandle::Options.LocalAllocSupportsLazyInit)
+      {
+        return action(core_alloc, args...);
```
mjp41, line 114: Is this reachable?

src/mem/localalloc.h
```diff
@@ -252,30 +274,85 @@ namespace snmalloc
...
+    {
+      T::ensure_init();
+    }
+
+    /**
+     * SFINAE helper.  Matched only if `T` does not implement `ensure_init`.
+     * Does nothing if called.
+     */
+    template<typename T>
+    SNMALLOC_FAST_PATH auto call_ensure_init(T*, long)
```
mjp41, line 324: Is the `T*` only required to make the template parameter used?  Does `UNUSED(T);` work if that is the case?

src/override/malloc-extensions.cc
```diff
@@ -6,10 +6,8 @@ using namespace snmalloc;
 
 void get_malloc_info_v1(malloc_info_v1* stats)
 {
-  auto unused_chunks =
-    Globals::get_handle().get_slab_allocator_state().unused_memory();
-  auto peak =
-    Globals::get_handle().get_slab_allocator_state().peak_memory_usage();
+  auto unused_chunks = Globals::get_slab_allocator_state().unused_memory();
```
mjp41, line 9: Do we need to guard against the case where this requires a parameter to `get_slab_allocator_state`?  I guess in those cases don't have a default?

src/snmalloc.h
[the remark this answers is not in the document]
mjp41: I think you need provide some information here.  My original intent was for any custom configuration to be
```
#include <snmalloc_core.h>

// Custom config that defines what Alloc means
...
namespace snmalloc
{
  using Alloc = ...
} // namespace snmalloc

#incude <snmalloc_front.h>
```

But with this I am not sure how to use this header.  Do I need to include something before 
```
#include <snmalloc.h>
```
If I include this header with `SNMALLOC_PROVIDE_OWN_CONFIG` set, then `snmalloc_front.h` won't compile as it doesn't have a definition of `snmalloc::Alloc`?

mjp41 (commented):
So I have two reservations with this PR
* removing the dynamic handle instance
* test coverage of alternative configurations

For the first, I agree that the existing code did not make use of this.   It was added as I thought, we might have multiple groups of allocators using different regions of memory for there backends, but I am not sure what I had done was sufficient for that.   Overall, I am fine for it to dissappear, but fear it will need something threading through like this, and the parameter to `get_slab_allocator_state` already feels like it is needed.   However, as there is no illustration of that usage, it is hard to say for sure.

Second, none of the alternatives are currently tested.  I am okay with having pretty minimal coverage, but could we perhaps change the `thread_alloc_external` test to cover the pre-initialised case (`LocalAllocSupportsLazyInit = false`)?

Ultimately, I would like to make `func-malloc-*` be configurable, so we can at least get basic coverage of all the configuration, but that is beyond this PR.

davidchisnall:
For the sandbox case, I currently have an empty `GlobalState` and put everything sandbox-specific in `LocalState`.  It would be easy for `LocalState` to have a reference to a per-larger-context if it's necessary (e.g. one `LocalState` per thread with a reference to a `SandboxState`).  Providing a way of requesting a `LocalState` per {some context} is probably less invasive than a similar thing for different `GlobalState`.

src/snmalloc.h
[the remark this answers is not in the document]
davidchisnall: In my `sandbox.h` I [include `snmalloc.h`](https://github.com/microsoft/verona/blob/bf705fa0dc443f2c0d8331c3ea5d17bf34d6dbed/experiments/process_sandbox/include/process_sandbox/sandbox.h#L22) and later define a config that provides the [sandbox definitions for the outside allocator](https://github.com/microsoft/verona/blob/bf705fa0dc443f2c0d8331c3ea5d17bf34d6dbed/experiments/process_sandbox/include/process_sandbox/sandbox.h#L323).  This is then [used directly as an explicit template instantiation](https://github.com/microsoft/verona/blob/bf705fa0dc443f2c0d8331c3ea5d17bf34d6dbed/experiments/process_sandbox/include/process_sandbox/sandbox.h#L387).  

When this file is included by thing in the host side, nothing I've written uses the definition of `Alloc`, but in the child I do and I depend on it being a custom thing and I also need to include `sandbox.h`, which includes `snmalloc.h`.

Without this change, if I include a header that includes `snmalloc.h` *after* I include the snmalloc headers that I need to define my custom malloc then I get a build failure because the `snmalloc.h` inclusion redefines `Alloc` with a different type.

With this change, it's a per-compilation-unit decision whether you want to override the default allocator definition and, if you do, then you must do so before you include `snmalloc.h` (which you may do indirectly).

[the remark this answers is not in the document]
mjp41: @davidchisnall if you included `snmalloc_core.h` does your example still work, and if not, what is it missing?

My view intent was roughly
* snmalloc_core.h  - things you need to build a custom snmalloc
* snmalloc_front.h - given a definition of snmalloc::Alloc give me the common things like thread local allocators, tests across the allocator pool (glocal_alloc.h), etc.
* snmalloc.h - default allocator setup for this platform.

Happy to change, but want to understand.

src/mem/fixedglobalconfig.h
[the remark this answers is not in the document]
mjp41: > is necessary if you want to be able to have different allocator state per non-type-based allocator context

Okay, that was my intent with `handle`. By doing it using the backend local state (`LocalState` is the `Backend::LocalState`?), this feels a bit like you are breaking the layering as the Config is using a backend component to decide which SlabAllocator state to use, which is the next level out.

src/snmalloc.h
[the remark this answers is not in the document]
davidchisnall: I see.  I'd expect most users of {foo} to include {foo}.h.  I presumably could just include `snmalloc_core.h` in my `sandbox.h` instead of `snmalloc.h`, but I'm still a bit nervous if I add a library dependency and it includes `snmalloc.h` instead of `snmalloc_core.h` and that breaks my ability to define a custom `snmalloc::Alloc`.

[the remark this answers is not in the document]
mjp41: I guess my thought was that most people including `snmalloc.h` would just want a good allocator for their platform directly linked into the application.  

My concern with your current `#ifdef` is that it still includes `snmalloc_front.h`, which includes `ScopedAllocator` and `ThreadAlloc` both of which assume `snmalloc::Alloc` exists.  So I am surprised that this doesn't fail to compile. 

I also think `globalconfig.h` should not be included if `SNMALLOC_PROVIDE_OWN_CONFIG` is set.  As you might not want the statics to be compiled into your binary, if you are building your own config.

Perhaps, the right way to go is for `snmalloc.h` to degenerate to `snmalloc_core.h` if `SNMALLOC_PROVIDE_OWN_CONFIG` is set:

```C++
#pragma once

// Core implementation of snmalloc independent of the configuration mode
#include "snmalloc_core.h"

#ifndef SNMALLOC_PROVIDE_OWN_CONFIG
// Default implementation of global state
#  include "mem/globalconfig.h"

// The default configuration for snmalloc
namespace snmalloc
{
  using Alloc = snmalloc::LocalAllocator<snmalloc::Globals>;
}

// User facing API surface, needs to know what `Alloc` is.
#  include "snmalloc_front.h"
#endif
```

Then in we could perhaps eliminate `snmalloc_core` as it is not required, as you can always use `snmalloc.h`.  Would this be acceptable?  I think it would fix your concern,

src/backend/pagemap.h
```diff
@@ -66,6 +66,28 @@ namespace snmalloc
...
+    {
+      constexpr size_t COVERED_BITS = bits::ADDRESS_BITS - GRANULARITY_BITS;
+      constexpr size_t ENTRIES = bits::one_at_bit(COVERED_BITS);
+      return ENTRIES * sizeof(T);
+    }
+
+    /**
+     * Initialise with pre-allocated memory.
+     */
+    void init(T* address)
```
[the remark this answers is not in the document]
ihaller: My thinking was that it is best to let the Pagemap report the exact size instead of trying to predict how big it wants it to be, but we can figure it out in a future PR if somebody wants to use it like that.

src/snmalloc.h
```diff
@@ -3,8 +3,12 @@
 // Core implementation of snmalloc independent of the configuration mode
 #include "snmalloc_core.h"
 
+// If you define SNMALLOC_PROVIDE_OWN_CONFIG then you must provide your own
```
mjp41, line 6: Not sure I agree with this comment.  I was expecting to do something like

```C++
#define SNMALLOC_PROVIDE_OWN_CONFIG
#include <snmalloc.h>
// define snmalloc::Alloc
...

#include <snmalloc_front.h>
```

src/mem/metaslab.h
```diff
@@ -177,6 +177,16 @@ namespace snmalloc
...
 
+    /**
+     * Constructor, provides the remote and sizeclass embedded in a single
+     * pointer-sized word.  This format is not guaranteed to be stable and so
+     * the second argument of this must always be the return value from
+     * `get_remote_and_sizeclass`.
+     */
+    MetaEntry(Metaslab* meta, uintptr_t remote_and_sizeclass)
+    : meta(meta), remote_and_sizeclass(remote_and_sizeclass)
+    {}
```
nwf, line 188: Is this used?  (is github just hiding a file from me in the diff view?  I don't think so?)

src/backend/backend.h
```diff
@@ -51,54 +52,41 @@ namespace snmalloc
...
+      static_assert(fixed_range_ == fixed_range, "Don't set SFINAE parameter!");
 
-        pagemap.init();
+      pagemap.init();
 
-        if constexpr (fixed_range)
-        {
-          abort();
-        }
+      if constexpr (fixed_range)
```
mjp41, line 70: I think the `enable_if_t` makes this redundant.  (My fault, sorry).

mjp41 (approved):
LGTM, I like the refactor. I will probably take a second refactoring pass, once I experiment with something like #272 on this branch.  I think I will need to adjust a little further, but I am not sure how until I try.

src/backend/address_space_core.h
```diff
@@ -1,6 +1,8 @@
 #pragma once
 #include "../ds/address.h"
 #include "../ds/flaglock.h"
+#include "../mem/allocconfig.h"
+#include "../mem/metaslab.h"
```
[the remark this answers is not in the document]
davidchisnall: I think it's fine in terms of the layering now, though we may want to split `mem`.  There are three categories of thing:

 - Templated bits of allocator that are completely generic.  These are in `mem` and are safe to use in a back end implementation.
 - Back ends, which can consume things from the first category and provide instantiation-specific policy.
 - Front ends, which consume things from both categories.  These are also in `mem`, which is probably the source of the confusion.

@mjp41, what do you think about moving all of the templated things that don't refer to `snmalloc::Alloc` into something like `src/core`?

src/mem/metaslab.h
davidchisnall: [Here](https://github.com/microsoft/verona/blob/4cb68fbae688c1e0ef8b701e61eed00d072a6374/experiments/process_sandbox/src/libsandbox.cc#L148).

src/backend/backend.h
davidchisnall: I'll remove it.

src/backend/address_space_core.h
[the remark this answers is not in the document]
mjp41: Happy to do something like that: https://github.com/microsoft/snmalloc/pull/366#issuecomment-893371081

davidchisnall:
Deleted some dead code, clarified a comment, rebased on the new version and tidied up commit history.  I'll merge once CI is happy.

PR #360, opened 2021-07-30
Attempt to get GitHub actions working with FreeBSD runners.

davidchisnall:


.github/workflows/main.yml
```diff
@@ -0,0 +1,185 @@
...
+    # Run the clang-format check and error if it generates a diff
+    - name: Run clang-format
+      working-directory: ${{github.workspace}}/build
+      run: |
+        set -eo pipefail
+        make clangformat
+        git diff --exit-code
+    - name: Run clang-tidy
+      run: |
+        #clang-tidy-9  src/override/malloc.cc  -header-filter="`pwd`/*" -warnings-as-errors='*' -export-fixes=tidy.fail -- -std=c++17 -mcx16 -DSNMALLOC_PLATFORM_HAS_GETENTROPY=0
```
mjp41, line 180: Is this commented out?

mjp41 (commented):
Are there any GCC (g++) builds? We use to test with g++-8 before I think.  Perfectly fine to use a different version, but I think we should have at least one Debug and Release for GCC.

Otherwise, looks good other than ARM 32-bit has been running for three hours, so far.

.github/workflows/main.yml
davidchisnall: clang-tidy isn't installed on the runner.  I'm not sure if it's actually necessary because I couldn't see how the old one was reporting errors (which is slightly embarrassing, because I'm pretty sure I wrote this bit of the old CI system).  Do we ever get errors from this in PRs?  If not, I'll disable it entirely.  If so, I'll try to make it work.

mjp41: I have had it regularly complain at me.  Normally for the better.

My understanding was it generated a fixes file:
```
 -export-fixes=tidy.fail 
```

Then  checked if that existed :
```
        if [ -f tidy.fail ] ; then
          cat tidy.fail
          exit 1
        fi
```

It worked pretty well on ADO.

davidchisnall:
> Are there any GCC (g++) builds? We use to test with g++-8 before I think. Perfectly fine to use a different version, but I think we should have at least one Debug and Release for GCC.

We are testing with the system compiler and standard library on all platforms unless explicitly overridden.  This is clang/libc++ on FreeBSD, Apple clang/libc++ on macOS, gcc/libstdc++ on Ubuntu.  There are two additional combinations (from line 46):

 - Clang 12/libc++ on FreeBSD (clang 11 is the system compiler)
 - Clang 10/libstdc++ on Ubuntu

The Clang 12/libc++ on FreeBSD is a build-only job, it doesn't run the tests.  The Ubuntu one does.  We have the following full list of build configurations:

 - Ubuntu 18.04, gcc 7.5, libstdc++, Release
 - Ubuntu 20.04, gcc 9, libstdc++, Release
 - FreeBSD 12.2, clang 10.0.1, libc++, Release
 - FreeBSD 13.0, clang 11.0.1, libc++, Release (build only)
 - macOS 10.15, Apple Clang 12.0, libc++, Release
 - macOS 11, Apple Clang 12.0.5, libc++, Release
 - Ubuntu 18.04, gcc 7.5, libstdc++, Debug
 - Ubuntu 20.04, gcc 9, libstdc++, Debug
 - FreeBSD 12.2, clang 10.0.1, libc++, Debug (build only)
 - FreeBSD 13.0, clang 11.0.1, libc++, Debug
 - FreeBSD 13.0, clang 12, libc++, (build only)
 - macOS 10.15, Apple Clang 12.0, libc++, Debug
 - macOS 11, Apple Clang 12.0.5, libc++, Debug
 - Ubuntu 20.04, clang 10, libstdc++, (build only)

On Windows, we test the matrix of {Windows Server 2019 + Visual Studio 2019, Windows Server 2016 + Visual Studio 2017} x { UTC, Clang } x { Release, Debug }, excluding Windows Server 2016 + Visual Studio 2017 x Clang x * (the Visual Studio 2017 doesn't come with a clang toolchain that CMake can use and I don't want to spend effort making something work with an old platform), plus Windows Server 2016 + Visual Studio 2017 x UTC x Release + the Win8 compatibility bits of the PAL.  

> Otherwise, looks good other than ARM 32-bit has been running for three hours, so far.

That's weird.  I'll try to see if I can reproduce it.  It seems to pass for me, so I don't know why it's failing in CI.

mjp41 (approved):
LGTM, minor change. There are a few unused files in here: https://github.com/microsoft/snmalloc/tree/master/ci

The two linux_x64 and linux_x86 Docker images are no longer used.  (Though, we are no longer testing 32bit Linux now.)

The two scripts are also no longer used.

mjp41:
Can you update the status badge too in README.md.  
```
[![snmalloc CI](https://github.com/microsoft/snmalloc/actions/workflows/main.yml/badge.svg)](https://github.com/microsoft/snmalloc/actions/workflows/main.yml)
```
[snmalloc CI](https://github.com/microsoft/snmalloc/actions/workflows/main.yml)

mjp41 (approved):

PR #362, opened 2021-08-04
Fix status badge.

davidchisnall:
The status badge should report the most recent status for the master
branch, not the most recent actions run.


mjp41 (approved):
LGTM

PR #363, opened 2021-08-04
Add cross CI for Ubuntu powerpc64el

nwf:
The naming is more than a little confusing.  AIUI, in Ubuntu's terminology, the
`powerpc` packages are for 32-bit machines, the `ppc64-powerpc` packages are for
Big Endian 64-bit machines, and the `ppc64el` packages are for Little Endian
64-bit machines.

Everyone seems to have agreed that the long-term answer for 64-bit PowerPC is
Little Endian (Debian maintains an unofficial BE port, but Canonical/Ubuntu,
RedHat, and SUSE all seem to have standardized on LE).

For maximal confusion, the appropriate triple's first component is, however,
`powerpc64le`.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -87,7 +87,7 @@ jobs:
     strategy:
       matrix:
         build-type: [ Release, Debug ]
-        arch: [ arm64, armhf ]
+        arch: [ arm64, armhf, ppc64el ]
         include:
           - arch: armhf
             system-processor: arm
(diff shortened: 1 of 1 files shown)


mjp41:
Does this have the 64KiB pages?  I am not going to be able to break either your or @davidchisnall's machines anymore?

nwf:
"Should have, yes" and "At least not in this way; growth mindset 'n all that", respectively, I think. :)

davidchisnall:
Looks good to me.  I don't know whether this actually enforces 64 KiB pages: does the user-mode qemu enforce this or does it just pass through the size to the host mmap / mprotect and allow 4 KiB pages (I'm now really curious to know how it works with x86-on-PowerPC if you try to mprotect a 4 KiB range)?

davidchisnall (approved):

PR #365, opened 2021-08-05
Clean up QEMU CI

davidchisnall:
Use a matrix containing objects rather than a matrix containing keys
that we try to use to define arrays of things.

Fix a typo in ARCH.  Too much Arm made it AARCH.


nwf (approved):
LGTM

PR #366, opened 2021-08-05
Allowing GlobalState to be extended easier for custom Backend implementations

ihaller:
This change is to enable code reuse when implementing a custom BackendAllocator, by enabling inheritance from it and its GlobalState.
Changes field protection from private to protected within GlobalState and templates the actual type of the Pagemap entries, with a default to MetaEntry.
Valid template arguments for the Pagemap entries should be able to seamlessly convert to and from MetaEntry and have a size that is a power of 2.

### DIFF
--- a/src/backend/backend.h
+++ b/src/backend/backend.h
@@ -11,7 +11,10 @@ namespace snmalloc
    * This class implements the standard backend for handling allocations.
    * It abstracts page table management and address space management.
    */
-  template<SNMALLOC_CONCEPT(ConceptPAL) PAL, bool fixed_range>
+  template<
+    SNMALLOC_CONCEPT(ConceptPAL) PAL,
+    bool fixed_range,
+    typename PageMapEntry = MetaEntry>
   class BackendAllocator
   {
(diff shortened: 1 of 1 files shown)


nwf:
How messy would it be to not default the template parameter to `MetaEntry` (which is defined in `src/mem`, not `src/backend`)?

ihaller:
It would only need to be specified in 2 files, both located in `src/mem` and once when people implement a new `CommonConfig`. I am happy to make this change, if people decide that it is the right thing to do.

nwf:
It's probably best left for another PR if it's not trivial.  This one LGTM.

mjp41 (approved):
LGTM, @davidchisnall?

mjp41:
@nwf the relationship between `src/mem` and `src/backend` is not the cleanest, I am hoping that more eyes will bring clarity.  We might need a `src/core` which contains components the `backend` can depend on.

davidchisnall (approved):

davidchisnall:
I'm not using `BackendAllocator` at all, but this doesn't conflict with anything that I am using or any of my other changes.

PR #369, opened 2021-08-06
PAL Haiku update.

devnexen:
with the beta3 release, and realizing MADV_DONTNEED is a "no-op"
MADV_FREE however is the only making sure the page range is discarded.


davidchisnall (approved):
Looks good to me.

PR #370, opened 2021-08-06
Added the ability to perform a reverse lookup in the Pagemap

ihaller:
Given a const reference to an entry in the Pagemap (as the one returned by get), return the base address of the range it corresponds to.

### DIFF
--- a/src/backend/pagemap.h
+++ b/src/backend/pagemap.h
@@ -188,7 +188,7 @@ namespace snmalloc
     /**
      * Get the number of entries.
      */
-    constexpr size_t num_entries() const
+    [[nodiscard]] constexpr size_t num_entries() const
     {
       if constexpr (has_bounds)
       {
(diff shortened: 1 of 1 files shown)


src/backend/pagemap.h
```diff
@@ -223,6 +238,21 @@ namespace snmalloc
...
 
+    /**
+     * Return the starting address corresponding to a given entry within the
+     * Pagemap. Also checks that the reference actually points to a valid entry.
+     */
+    [[nodiscard]] address_t get_address(const T& t) const
+    {
+      address_t entry_offset = address_cast(&t) - address_cast(body);
+      address_t entry_index = entry_offset / sizeof(T);
+      if (entry_offset % sizeof(T) != 0 || entry_index >= num_entries())
```
mjp41, line 249: Should this be an assert?  Is this code likely to be performance sensitive?

mjp41 (approved):
LGTM

src/backend/pagemap.h
ihaller: While I don't expect this code to be performance sensitive (not more than Pagemap::get), I think that an assert might be more appropriate, since it only does address computation and not pointer indexing.

PR #372, opened 2021-08-12
Miscellaneous accumulated trivia, 202108 edition

nwf:
None of these are expected to have functional implications.

### DIFF
--- a/src/test/func/memory/memory.cc
+++ b/src/test/func/memory/memory.cc
@@ -255,7 +255,7 @@ void test_external_pointer()
       if ((size_t)p4 != (size_t)p1 + size - 1)
       {
         std::cout << "size: " << size << " end(p4): " << p4 << " p1: " << p1
-                  << "  p1+size-1: " << (void*)((size_t)p1 + size - 1)
+                  << "  p1+size-1: " << pointer_offset(p1, size - 1)
                   << std::endl;
       }
       SNMALLOC_CHECK((size_t)p4 == (size_t)p1 + size - 1);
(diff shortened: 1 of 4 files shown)


davidchisnall (approved):
LGTM, I'll let @mjp41 merge if he's happy with them.

src/test/func/memory/memory.cc
```diff
@@ -255,8 +255,8 @@ void test_external_pointer()
       if ((size_t)p4 != (size_t)p1 + size - 1)
       {
         std::cout << "size: " << size << " end(p4): " << p4 << " p1: " << p1
-                  << "  p1+size-1: " << (void*)((size_t)p1 + size - 1)
-                  << std::endl;
+                  << "  p1+size-1: "
+                  << (void*)(uintptr_t)((size_t)p1 + size - 1) << std::endl;
```
mjp41, line 259: The `(void*)` here are due to me being lazy and wanting to print the address in hex, and that is how pointers are printed.  We could fix this with `std::hex << `, or in this case actually using `pointer_offset`.
```suggestion
                  << pointer_offset(p1, size - 1) << std::endl;
```

src/test/func/pagemap/pagemap.cc
```diff
@@ -40,9 +40,9 @@ void check_get(
 
   if (value.v != expected.v)
   {
-    std::cout << "Location: " << (void*)address << " Read: " << value.v
-              << " Expected: " << expected.v << " on " << file << ":" << lineno
-              << std::endl;
+    std::cout << "Location: " << (void*)(uintptr_t)address
```
mjp41, line 43:
```suggestion
    std::cout << "Location: " << std::hex << address
```

mjp41 (approved):
LGTM couple of minor bits.

mjp41 (approved):

PR #373, opened 2021-08-12
Use atomics to serialize large deallocs

nwf:
I'm not sure this is the right approach, and so am quite open to suggestions.  But, AFAICT, there's nothing otherwise preventing two LocalAlloc-s from attempting to thread the same large allocation (which are not considered "owned") onto their free lists in parallel, which would lead to all kinds of future excitement.

### DIFF
--- a/src/backend/backend.h
+++ b/src/backend/backend.h
@@ -271,6 +271,11 @@ namespace snmalloc
       return pagemap.template get<potentially_out_of_range>(p);
     }
 
+    static MetaEntry& get_meta_data_mut(address_t p)
+    {
+      return pagemap.get_mut(p);
+    }
+
     /**
      * Set the metadata associated with a chunk.
      */
--- a/src/backend/pagemap.h
+++ b/src/backend/pagemap.h
@@ -239,6 +239,23 @@ namespace snmalloc
     }
 
     /**
+     * Like get, but gives mutable access to a real pagemap entry.
+     */
+    T& get_mut(address_t p)
+    {
+      if constexpr (has_bounds)
+      {
+        if (p - base > size)
+        {
+          PAL::error("Internal error: Pagemap read access out of range.");
+        }
+        p = p - base;
+      }
+
+      return body[p >> SHIFT];
+    }
+
+    /**
      * Return the starting address corresponding to a given entry within the
      * Pagemap. Also checks that the reference actually points to a valid entry.
      */
--- a/src/mem/localalloc.h
+++ b/src/mem/localalloc.h
@@ -495,10 +495,12 @@ namespace snmalloc
       // Large deallocation or null.
       if (likely(p != nullptr))
       {
-        // Check this is managed by this pagemap.
-        check_client(entry.get_sizeclass() != 0, "Not allocated by snmalloc.");
+        sizeclass_t sizeclass = entry.get_sizeclass();
 
-        size_t size = bits::one_at_bit(entry.get_sizeclass());
+        // Check this is managed by this pagemap.
+        check_client(sizeclass != 0, "Not allocated by snmalloc.");
+
+        size_t size = bits::one_at_bit(sizeclass);
 
         // Check for start of allocation.
         check_client(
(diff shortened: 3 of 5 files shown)


nwf:
This happened to work locally since `MetaEntry` still had its copy constructor and assignment operators.  A deeper set of changes is necessary for platforms for which that isn't true.  Hm.

src/backend/pagemap.h
```diff
@@ -238,6 +238,23 @@ namespace snmalloc
       return body[p >> SHIFT];
     }
 
+    /**
+     * Like get, but gives mutable access to a real pagemap entry.
```
davidchisnall, line 242: `get` also returns a `T&`, so should allow mutation.  I'm not sure what's different here.

src/backend/backend.h
```diff
@@ -271,6 +271,11 @@ namespace snmalloc
       return pagemap.template get<potentially_out_of_range>(p);
     }
 
+    static MetaEntry& get_meta_data_mut(address_t p)
```
mjp41, line 274: I am not keen to expose this.  It is hard to implement in the sandbox RPC case. I would be much keener if we could use the metaslab for this.

src/mem/metaslab.h
```diff
@@ -208,18 +210,34 @@ namespace snmalloc
...
     [[nodiscard]] sizeclass_t get_sizeclass() const
     {
-      return remote_and_sizeclass & (alignof(RemoteAllocator) - 1);
+      return static_cast<sizeclass_t>(remote_and_sizeclass.load()) &
+        (alignof(RemoteAllocator) - 1);
+    }
+
+    /**
+     * Attempts to atomically take ownership of a chunk that is being
+     * deallocted.  May fail if multiple threads attempt at once, but spurious
```
mjp41, line 231:
```suggestion
     * deallocated.  May fail if multiple threads attempt at once, but spurious
```

src/mem/metaslab.h
```diff
@@ -208,18 +210,34 @@ namespace snmalloc
...
+     * failures are not permitted; one thread must win in the event of a race.
+     */
+    [[nodiscard]] ChunkRecord* claim_for_large_dealloc(sizeclass_t sizeclass)
```
mjp41, line 234: I would prefer this to be part of `backend.h`.  This is something that a new backend has to support, and if the pagemap is mapped read-only then that could be complex to achieve.

src/mem/localalloc.h
```diff
@@ -509,9 +511,20 @@ namespace snmalloc
...
+         * TODO: There's nothing to say that this dealloc() is paired with the
+         * most recent allocation of this object, but we can at least check that
+         * it hasn't become freed, resized, or used as slab in the interim.
+         */
+        MetaEntry& entry_mut =
+          SharedStateHandle::get_meta_data_mut(address_cast(p));
+        ChunkRecord* slab_record = entry_mut.claim_for_large_dealloc(sizeclass);
+        if (unlikely(slab_record == nullptr))
+        {
+          return;
```
mjp41, line 525: Should this be an error that we report.  At least in the `SNMALLOC_CHECK_CLIENT` mode.  Also, can we switch off checking?  I would really like any enforcement of correct usage to be under `SNMALLOC_CHECK_CLIENT`, then we can get a true base-line for manual memory management checking costs.

mjp41 (commented):
So we definitely need to address this to detect double free of large allocs.  Personally, I would like to see this as actually part of the backend API.  I am wondering if we should change the backend to have a `dealloc_chunk`, and then the backend is responsible for detecting double free, and mis-sized frees.
```C++
dealloc_chunk(void* chunk_base, RemoteAllocator* remote, sizeclass_t sizeclass)
{
   ...
}
```
This then puts the CAS/exch in here for detecting double free.  Passing `remote` and `sizeclass` could help with a racing free. 

I had already been wondering if we should change the API like this, so that I can build a lazy decommit like feature into the backend.  It would also involve moving the `SlabAllocator` into the backend. 

I would be fine with exchange rather than CAS, and fail if what we get back is wrong.

src/backend/backend.h
davidchisnall: I agree - I missed the layering when I read the code.  In the sandbox case, the pagemap is read-only in the child and so anything that updates it must be part of the backend API, where they can become RPCs.  

I think, after looking at @ihaller's code, it would be good if we could avoid exposing `MetaEntry` outside of the back end entirely.  The back end is then free to use whatever storage it wants, including using a larger structure in the pagemap.

src/mem/metaslab.h
ihaller: +1

nwf:
Abandoning this attempt; will revisit in light of #441.

PR #374, opened 2021-08-12
RISCV AAL & CI

nwf:
Thanks to @davidchisnall, adding cross-arch Linux CI is much easier now.


nwf:
`lld` apparently does not get along well with the `gcc`-provided `crt1.so`, so this is going to require that we go find a cross-binutils `ld.bfd` for RISC-V, and that's going to take a great deal more understanding of Ubuntu and some associated. shenanigans with the CI tooling.  Retracting for now.

PR #375, opened 2021-08-13
Make threshold for waking always a percentage

mjp41:
The threshold for waking is used to ensure that we only use a slab when
it has sufficient space that we won't hit the slow path to soon after
using this slab. In the checked version, this is also used to give some
entropy in layout.  Changing this to always be a pertcentage in the
checked case increases the effectiveness of the free list to detect OOB
write.


nwf (commented):
LGTM.  To check my understanding: this should have no change for slabs with fewer than 128 (= 4 * 32) objects, which looks like small size classes above 6 (<= 127 bytes)?

PR #377, opened 2021-08-23
Changes to integrate snmalloc2 into Verona

ihaller:
The primary change is to allow a generic usage of the Pool type, which does not just get its pool state from the SharedStateHandle (this being necessary for the pool of allocators). The SharedStateHandle template parameter is promoted to the class as all methods need it. Also the Pool class now takes a function as parameter which is the accessor to the pool state. There is a helper class to create a singleton global pool state for arbitrary types.
For ease of use I have also added an alias AllocPool<typename SharedStateHandle> for the allocator pool.
If people agree with the overall design, then I will add a test for this functionality.

An additional small change is to delete the copy constructor and copy assignment operator for LocalAllocator. This is to avoid the misuse of the ThreadAlloc::get API using the pattern `auto alloc = ThreadAlloc::get`.

Something else that I discovered, was that the new `snmalloc::debug_check_empty` API can result in corruption, if a thread is being destroyed in parallel (CoreAllocator::attached_cache is manipulated both by the thread being destroyed and the called of the global method).
When used with Verona scheduler threads, their state should be stable when invoking this API so I did not make an attempt to fix the race condition, but this is something to consider. One test in Verona does use external threads and it triggers this issue, so I had to extend the lifetime of the external thread to fix it.

### DIFF
--- a/src/mem/pool.h
+++ b/src/mem/pool.h
@@ -21,7 +21,10 @@ namespace snmalloc
   template<class T>
   class PoolState
   {
-    template<typename TT, typename SharedStateHandle>
+    template<
+      typename TT,
+      typename SharedStateHandle,
+      PoolState<TT>& get_state()>
     friend class Pool;
 
   private:
@@ -34,35 +37,28 @@ namespace snmalloc
   };
 
   /**
-   * Class used to instantiate non-allocator pools using a Singleton PoolState.
+   * Class used to instantiate a global non-allocator PoolState.
    */
-  template<typename T, typename SharedStateHandle>
-  class PoolStateHandle
+  template<typename T>
+  class SingletonPoolState
   {
-    static PoolState<T>& pool_state =
-      Singleton<PoolState<T>, PoolStateHandle::make_state>::get();
-
-    static PoolState<T>* make_state()
-    {
-      return ChunkAllocator::alloc_meta_data<PoolState<T>, SharedStateHandle>(
-        nullptr);
-    }
+    static inline PoolState<T> state;
 
   public:
     static PoolState<T>& pool()
     {
-      return pool_state;
+      return state;
     }
   };
 
-  template<typename T, typename SharedStateHandle>
+  template<typename T, typename SharedStateHandle, PoolState<T>& get_state()>
   class Pool
   {
   public:
     template<typename... Args>
     static T* acquire(Args&&... args)
     {
-      PoolState<T>& pool = SharedStateHandle::pool();
+      PoolState<T>& pool = get_state();
       T* p = pool.stack.pop();
 
       if (p != nullptr)
@@ -99,14 +95,14 @@ namespace snmalloc
       // is returned without the constructor being run, so the object is reused
       // without re-initialisation.
       p->reset_in_use();
-      SharedStateHandle::pool().stack.push(p);
+      get_state().stack.push(p);
     }
 
     static T* extract(T* p = nullptr)
     {
       // Returns a linked list of all objects in the stack, emptying the stack.
       if (p == nullptr)
-        return SharedStateHandle::pool().stack.pop_all();
+        return get_state().stack.pop_all();
 
       return p->next;
     }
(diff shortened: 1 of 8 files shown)


src/mem/pool.h
```diff
@@ -32,11 +33,33 @@ namespace snmalloc
...
+    }
+
+  public:
+    static PoolState<T>& pool()
+    {
+      return pool_state;
+    }
+  };
+
+  template<typename T, typename SharedStateHandle>
```
nwf, line 58: If you think all the uses of this are going to be of the shape `Pool<T<U>, U>`, maybe codify that?  `template<template<typename U> TC, U> class Pool { using T = TC<U>; /* ... */ }` ?

nwf (commented):
One question but LGTM either way.

src/mem/pool.h
ihaller: I don't think that all uses will be of the form Pool<T<U>< U>. Only the allocator pool has this shape for now, the Verona pools (LocalEpoch and LocalLog) don't. But thanks for the comment, I will try to keep this pattern in mind for the future.

ihaller:
Sorry about the massive change after creating the PR. Just as I created it, an alternative structure came to my mind and I tried to quickly code it up as it avoids code duplication.

davidchisnall (commented):
Aside from missing comments, this looks good to me.

src/mem/pool.h
```diff
@@ -32,14 +36,88 @@ namespace snmalloc
...
+  class SingletonPoolState
+  {
+    static inline PoolState<T> state;
+    /**
+     * Thread-local initialization marker for a contention-free way to skip
+     * initialization. SharedStateHandle::ensure_init allows for multiple calls,
+     * but it uses a reentrency-safe check which is more expensive.
+     */
+    static thread_local inline bool initialized;
```
mjp41, line 54: Could you add an ecplicit `{false}` here?
```suggestion
    static thread_local inline bool initialized{false};
```

mjp41: I like the contention free check here.  We should perhaps use this trick elsewhere.

src/mem/pool.h
```diff
@@ -32,14 +36,88 @@ namespace snmalloc
...
 
-  template<typename T>
+  /**
+   * Helper class used to instantiate a global PoolState.
+   *
+   * SingletonPoolState::pool is the default provider for the PoolState within
+   * the Pool class.
+   */
+  template<typename T, typename SharedStateHandle>
+  class SingletonPoolState
```
mjp41, line 46: Is it possible to use
https://github.com/microsoft/snmalloc/blob/6e638742e3c66549174d4c264bd05c9435938ac1/src/ds/helpers.h#L15-L46

mjp41 (approved):
LGTM, couple of minor things.

src/mem/pool.h
ihaller: Initially I tried to use Singleton, but it assumes that the Object is copy assignable (obj = init()), which PoolState cannot be (ABA is not copy contructible and it is good that it is not). As I looked through the code, changing Singleton to use obejct constructors instead of init did not look viable as there are many use case with non-class types.

src/mem/pool.h
ihaller: An alternative suggested to me was to use SharedStateHandle::is_initialized, which looks at the global flag, but does not acquire a lock to do so. Would you prefer that one instead?

src/mem/pool.h
mjp41: Cool.  Thanks just wanted to check it had been considered.

src/mem/pool.h
mjp41: I am happy with what you have.

src/mem/pool.h
davidchisnall: This seems like something that it would be good to fix in Singleton, since the ability to copy a singleton is not likely to be desirable.  I believe the copy here isn't actually a copy, it's expected to invoke guaranteed copy elision.  Can this be worked around by providing a private copy constructor with an `assert(0)` in it and making Singleton a friend?  If the copy constructor is actually executed, then we have a bug, but if guaranteed copy elision actually works then the object is created in place and the copy constructor is never called.

mjp41 (approved):
LGTM

PR #382, opened 2021-08-26
Move bits/address_bits to AAL/PAL

nwf:
FreeBSD RISC-V is the primary motivating example of a platform that really constrains the AS: it exposes a 38 bit AS for user programs (that is, it uses the sv39 page table format and cleaves the AS in half for kernel/user).  If we go with the traditionally hard-coded 48 bit AS estimate from `ds/bits`, we end up trying to allocate the whole user AS (48 bits in AS - 14 bits for a 16K chunk size + 4 bits for a `MetaEntry` = 38 bits of `Pagemap`).  As you might imagine, this doesn't go well for us (and it goes somehow even worse for us on CHERI when we ask for a 39-bit AS allocation...).  FreeBSD MIPS is similar and exposes a 39 bit AS and so also suffers (even if we only ask for half the AS for the Pagemap, it still doesn't really work out well for us), but I don't think that matters as much going forward.

### DIFF
--- a/src/aal/aal.h
+++ b/src/aal/aal.h
@@ -67,6 +67,66 @@ namespace snmalloc
       default_address_t,
       Arch>::address_t;
 
+  private:
+    /**
+     * SFINAE template and default case.  T will be Arch and the second template
+     * argument defaults at the call site (below).
+     */
+    template<typename T, typename = int>
+    struct default_bits_t
+    {
+      static constexpr size_t value = sizeof(size_t) * 8;
+    };
+
+    /**
+     * SFINAE override case.  T will be Arch, and the computation in the second
+     * position yields the type int iff T::bits exists and is a substituion
+     * failure otherwise.  That is, if T::bits exists, this specialization
+     * shadows the default; otherwise, this specialization has no effect.
+     */
+    template<typename T>
+    struct default_bits_t<T, decltype(((void)T::bits, 0))>
+    {
+      static constexpr size_t value = T::bits;
+    };
+
+  public:
+    /**
+     * Architectural word width as overridden by the underlying Arch-itecture or
+     * defaulted as per above.
+     */
+    static constexpr size_t bits = default_bits_t<Arch>::value;
+
+  private:
+    /**
+     * Architectures have a default opinion of their address space size, but
+     * this is mediated by the platform (e.g., the kernel may cleave the address
+     * space in twain or my use only some of the radix points available to
+     * hardware paging mechanisms).
+     *
+     * This is more SFINAE-based type-level trickery; see default_bits_t, above,
+     * for more details.
+     */
+    template<typename T, typename = int>
+    struct default_address_bits_t
+    {
+      static constexpr size_t value = (bits == 64) ? 48 : 32;
+    };
+
+    /**
+     * Yet more SFINAE; see default_bits_t for more discussion.  Here, the
+     * computation in the second parameter yields the type int iff
+     * T::address_bits exists.
+     */
+    template<typename T>
+    struct default_address_bits_t<T, decltype(((void)T::address_bits, 0))>
+    {
+      static constexpr size_t value = T::address_bits;
+    };
+
+  public:
+    static constexpr size_t address_bits = default_address_bits_t<Arch>::value;
+
     /**
      * Prefetch a specific address.
      *
--- a/src/mem/sizeclasstable.h
+++ b/src/mem/sizeclasstable.h
@@ -41,7 +41,7 @@ namespace snmalloc
 
   // Large classes range from [SUPERSLAB, ADDRESS_SPACE).// TODO
   static constexpr size_t NUM_LARGE_CLASSES =
-    bits::ADDRESS_BITS - MAX_SIZECLASS_BITS;
+    Pal::address_bits - MAX_SIZECLASS_BITS;
 
   inline SNMALLOC_FAST_PATH static size_t
   aligned_size(size_t alignment, size_t size)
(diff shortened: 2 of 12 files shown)


src/aal/aal.h
```diff
@@ -67,6 +67,41 @@ namespace snmalloc
       default_address_t,
       Arch>::address_t;
 
+    /* Architecture word width */
+    template<typename T, typename = size_t>
```
davidchisnall, line 71: Is T here a SFINAE thing?  Can you add a comment explaining what it's for?

src/aal/aal.h
```diff
@@ -67,6 +67,41 @@ namespace snmalloc
...
+
+    static constexpr size_t bits = default_bits_t<Arch>::value;
+
+    /*
+     * Architectures have a default opinion of their address space size, but
+     * this is mediated by the platform (e.g., the kernel may cleave the address
+     * space in twain or my use only some of the radix points available to
+     * hardware paging mechanisms).
+     */
+    template<typename T, typename = size_t>
```
davidchisnall, line 91: What are the template parameters here?

src/mem/sizeclasstable.h
```diff
@@ -179,7 +179,7 @@ namespace snmalloc
 
     auto rsize = sizeclass_to_size(sc);
 
-    if constexpr (bits::is64())
+    if constexpr (bits::BITS >= 64)
```
davidchisnall, line 182: I think we have two concepts here that feel slightly conflated:

 - What is the size of a C/C++ integer type that can express any address?  Or possibly what is the type that can represent any address.
 -  How many of those bits are actually ever used?

This wants to be gated on something like `std::is_same_v<uint64_t, Pal::address_t>`.

src/pal/pal_noalloc.h
```diff
@@ -4,23 +4,29 @@
...
    *
    * This is a minimal PAL for pre-reserved memory regions, where the
    * address-space manager is initialised with all of the memory that it will
    * ever use.
-   *
-   * It takes an error handler delegate as a template argument. This is
-   * expected to forward to the default PAL in most cases.
    */
-  template<typename ErrorHandler>
+  template<SNMALLOC_CONCEPT(PALNoAllocBase) BasePAL>
```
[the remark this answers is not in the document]
davidchisnall: Yes, but I think the `PALNoAllocBase` is a restricted subset and so is fine.  The important thing is that the base PAL-like thing shouldn't be required to be able to allocate memory because accidentally passing through `notify_not_using` on a shared memory segment may not do the right thing.

src/pal/pal_posix.h
```diff
@@ -132,6 +132,26 @@ namespace snmalloc
 
     static constexpr size_t page_size = Aal::smallest_page_size;
 
+    /*
+     * Address bit width is not purely a property of the architecture; it is
+     * potentially mediated by the platform.  By default, though, just pass
+     * through the aal's notion.
+     */
+
+    template<typename T, typename = size_t>
```
davidchisnall, line 141: These look duplicated from the things in the AAL, can they be merged somehow?

src/pal/pal_windows.h
```diff
@@ -62,6 +62,8 @@ namespace snmalloc
 
     static constexpr size_t page_size = 0x1000;
 
+    static constexpr size_t address_bits = Aal::address_bits;
```
davidchisnall, line 65: Please add a doc comment.

src/pal/pal_posix.h
nwf: I played with that for a while and I don't have an easy answer to hand.  The challenge is that SFINAE really does rely on the the S aspect, and C++'s template language doesn't let us quantify over the right kinds to composite things in a context where SFINAE holds.  Less abstractly, C++ templates don't have a first-class notion of a field name, so we can't build an AST node `T::field` in a SFINAE context, so the template name itself must encode the field name being tested.

That is, we'd really like to be able to write something like...
```
template <typename Value, Value def, typename Override, field f, typename _sfinae = int>
class Default
{
  Value v = def;
}

template<typename Value, Value def, typename Override, field f, decltype((void)Override::f, (int)0)>
class Default
{
  Value v = Override::f;
}
```
so that `Override::f` not existing continues to trigger a SF (INAE) in the last template argument, so that the compiler doesn't even see, and complain about, the same bit of syntax at the expression level three lines down.  But we can't, because `field` isn't a quantifiable thing.  And we can't write something like
```
template <typename Value, Value def, Value override, typename _sfinae = int>
class Default
{
  Value v = def;
}

template<typename Value, Value def, Value override, decltype((void)override, (int)0)>
class Default
{
  Value v = override;
}
```
because that requires that we construct our override (i.e., `Override::f`) at the call-site, but if it doesn't exist, that's not a SF, that's just AE.  And since SFINAE applies only at the particular template being considered, we can't try hiding the composition inside another template without basically duplicating the logic.

Less generally, I don't think we want to composite this with the AAL to have a three-layer subsumption architecture in the PAL definition because not all AALs are `AAL_Generic<>`, so we should treat `AAL::address_bits` as opaquely defined over there.

src/mem/sizeclasstable.h
nwf: Hm.  I think this might want to be gated rather on `sizeof(offset) >= 8`, as that will more closely determine whether the expression overflows?  @mjp41?

mjp41: It is really that the code is designed to work with the existing potential offsets with 64bit multiplication.  If we had a larger multiply, then that would be fine as long as we didn't have larger offsets.

This code still needs some tidying from the snmalloc2 refactor. So I will harden the choices when I do that.   So this is fine for now.

nwf:
Ah, software.  Let's see how this fares.  I think we don't need to be doing the SFINAE dance in PALPOSIX since those use inheritance (though, admittedly, curiously with `PALLinux : PALPOSIX<PALLinux>` rather than just `PALLinux : PALPOSIX`) and that suffices for `::address_bits`.

PR #383, opened 2021-08-31
Added constness to argv in Opt

ihaller:
Enables more options for writing tests and Opt should not be modifying argv content anyway.


mjp41 (approved):
LGTM

PR #384, opened 2021-09-01
CMake cleanup.

davidchisnall:
Modernise and tidy the CMake a bit:

 - Use generator expressions for a lot of conditionals so that things
   are more reliable with multi-config generators (and less verbose).
 - Remove C as a needed language.  None of the code was C but we were
   using C to test if headers worked.  This was fragile because a build
   with `CMAKE_CXX_COMPILER` set might have checked things compiled with
   the system C compiler and then failed when the specified C++ compiler
   used different headers.
 - Rename the `BACKTRACE_HEADER` macro to `SNMALLOC_BACKTRACE_HEADER`.
   This is exposed into code that consumes snmalloc and so should be
   'namespaced' (to the degree that's possible with C macros).
 - Use functions instead of macros for better variable scoping.
 - Factor out some duplicated bits into functions.
 - Update to the latest way of telling CMake to use C++17 or C++20.
 - Migrate everything that's setting global properties to setting only
   per-target properties.
 - Link with -nostdlib++ if it's available.  If it isn't, fall back to
   enabling the C language and linking with the C compiler.
 - Make the per-test log messages verbose outputs.  These kept scrolling
   important messages off the top of the screen for me.
 - Make building as a header-only library a public option.
 - Add install targets that install all of the headers and provide a
   config option.  This works with the header-only configuration for
   integration with things like vcpkg.
 - Fix a missing `#endif` in the `malloc_useable_size` check.  This was
   failing co compile on all platforms because of the missing `#endif`.
 - Bump the minimum version to 3.14 so that we have access to
   target_link_options.  This is necessary to use generator expressions
   for linker flags.

With these changes, it's now possible to install snmalloc and then, in
another project, do something like this:

```cmake
find_package(snmalloc CONFIG REQUIRED)
target_link_libraries(t1 snmalloc::snmalloc)
target_link_libraries(t2 snmalloc::snmallocshim-static)
```

In this example, `t1` gets all of the compile flags necessary to include
snmalloc headers for its build configuration.  `t2` is additionally
linked to the snmalloc static shim library.


CMakeLists.txt
```diff
@@ -1,13 +1,13 @@
-cmake_minimum_required(VERSION 3.8)
-project(snmalloc C CXX)
+cmake_minimum_required(VERSION 3.14)
```
nwf, line 1: Just because I've been dealing with similar things over at the CL, I'll note that this is too new for Ubuntu `bionic`/18.04, which has only 3.10 (but is fine for `focal`/20.04, which has 3.16; `hirsuit` has 3.18).

nwf: ... and yet it works on our 18.04 CI?  Is https://packages.ubuntu.com/search?keywords=cmake full of lies?

CMakeLists.txt
```diff
@@ -1,31 +1,75 @@
...
-option(SNMALLOC_USE_PTHREAD_DESTRUCTORS "Build using pthread destructors. Reduces the system dependencies, but may interact badly with C++ on some platforms." OFF)
+# Options that apply only if we're not building the header-only library
+cmake_dependent_option(SNMALLOC_RUST_SUPPORT "Build static library for rust" OFF "NOT SNMALLOC_HEADER_ONLY_LIBRARY" OFF)
+cmake_dependent_option(SNMALLOC_STATIC_LIBRARY "Build static libraries" ON "NOT SNMALLOC_STATIC_LIBRARY" OFF)
+cmake_dependent_option(SNMALLOC_OPTIMISE_FOR_CURRENT_MACHINE "Compile for current machine architecture" Off "NOT SNMALLOC_STATIC_LIBRARY" OFF)
+if (NOT SNMALLOC_HEADER_ONLY_LIBRARY)
+  # Pick a sensible default for the thread cleanup mechanism
+  if (${CMAKE_SYSTEM_NAME} STREQUAL FreeBSD)
+    set(SNMALLOC_CLEANUP_DEFAULT THREAD_CLEANUP)
+  elseif (UNIX)
```
mjp41, line 27: This doesn't work on Mac as the interaction with pthread and C++ destructors is bad.  It appeared the C++ TLS state had been zeroed before the pthread thread destructor was called. and then bad things happened.

CMakeLists.txt
```diff
@@ -95,162 +142,154 @@ macro(clangformat_targets)
...
+      endif()
 
-        # Remove all the duplicate new/malloc and free/delete definitions
-        target_link_libraries(${name} -Wl,--icf=all -fuse-ld=lld)
+      # Ensure that we do not link against C++ stdlib when compiling shims.
+      # If the compiler supports excluding the C++ stdlib implementation, use
+      # it.  Otherwise, fall back to linking the library as if it were C, which
+      # has roughly the same effect.
+      if (NOT ${SNMALLOC_CLEANUP} STREQUAL CXX11_DESTRUCTORS)
+        check_linker_flag(CXX "-nostdlib++" SNMALLOC_LINKER_SUPPORT_NOSTDLIBXX)
```
mjp41, line 276: I guess this is better than `--as-needed` as we will get link failures if we accidentally include something that uses C++.  But this might make Debugging harder, as adding anything like `std::cout` will then fail to compile.   Could we still include the `--as-needed` in the else branch.

CMakeLists.txt
```diff
@@ -338,3 +384,23 @@ if(NOT DEFINED SNMALLOC_ONLY_HEADER_LIBRARY)
 
   clangformat_targets()
 endif()
+
+install(TARGETS snmalloc EXPORT snmallocConfig)
```
mjp41, line 388: Should we be installing the `.so`s?

CMakeLists.txt
```diff
@@ -95,162 +142,154 @@ macro(clangformat_targets)
...
-      if (LLD_WORKS)
-        message("Using LLD.")
+    if(MSVC)
+      target_compile_options(${name} PRIVATE /Zi /W4 /WX /wd4127 /wd4324 /wd4201)
+      set(CMAKE_EXE_LINKER_FLAGS_RELEASE "${CMAKE_EXE_LINKER_FLAGS_RELEASE} /DEBUG")
+    else()
+      target_compile_definitions(${name} PRIVATE "SNMALLOC_EXPORT=__attribute__((visibility(\"default\")))")
+      target_compile_options(${name} PRIVATE
+        -fno-exceptions -fno-rtti -fomit-frame-pointer -ffunction-sections
+        -Wall -Wextra -Werror -Wundef
```
mjp41, line 253: Shouldn't the tests also be built with warnings?

CMakeLists.txt
davidchisnall: We are, on line 290 - they're installed as a side effect of installing the library target.  We only need to install the headers separately because snmalloc has an unusual directory structure that isn't supported by cmake.

CMakeLists.txt
davidchisnall: Yes.  Fixed.

CMakeLists.txt
davidchisnall: Note that this isn't applied to the tests: they still link the C++ standard library and so can use whatever they want.  This just prevents the shim libraries from depending on lib[std]c++.  I've hit problems in the past where we accidentally used something like `std::cout` on a code path that ended up being called before malloc was initialised and so we got deadlock rather than useful crashes so I'd much prefer we got a linker failure from this rather than a difficult-to-debug run-time condition.

mjp41: I agree entirely, that having a hard error for including the stdlib++ most of the time is the right default.  Prior to this change I was thinking about something like 
```sh
objdump -p libsnmallocshim.so | grep NEEDED
```

Perhaps, we could have command line option to enable stdlib++ (i.e. disable this code). I am just thinking about times where I have been trying to debug `LD_PRELOAD`ed things, and I carefully added some bits of logging.

I'm also concerned if we add stats printing on teardown back as an option then stdlib++ might be needed.

davidchisnall: I've added the linker flag that checks for undefined symbols and errors at link time with something more useful.  If you're doing some local tests with these things then you can patch this out.  We can also add the runtime dependency back in explicitly if and only if we enable stats so that we'll still get CI failures if we use anything from the C stdlib in non-stats builds.

CMakeLists.txt
davidchisnall: Discussed offline, but just in case anyone else finds this and is confused:

The GitHub-Hosted Actions Runners have CMake 3.21 installed from the Ubuntu snap package.

I would suggest moving to 3.18, since 3.21 is available for Ubuntu, macOS, Windows, and *BSD, but it looks as if Haiku currently has CMake packages for only 3.16.

mjp41 (approved):

PR #385, opened 2021-09-03
Metaslab: SlabLink field rather than parent class

nwf:
Introduce Metaslab::from_link(SlabLink*) to encapsulate the "container of"
dance.  Note that Metaslab was not a standard layout type prior to this change
(since both SlabLink and Metaslab defined non-static data members), and so the
reinterpret_cast<>s replaced here with ::from_link() were UB, but everyone lays
out classes as one expects so it was fine in practice.

Despite that Metaslab is now standard layout, we still don't fall back to the
inter-convertibility of a standard layout class and its first[*] data member
since we're going to want to put a common initial sequence across Metaslab and
ChunkRecord and the SlabLink isn't likely to be in it.

ETA: Yes, this is another case of the same thing as #270.

### DIFF
--- a/src/mem/corealloc.h
+++ b/src/mem/corealloc.h
@@ -304,7 +304,7 @@ namespace snmalloc
       while (curr != nullptr)
       {
         auto nxt = curr->get_next();
-        auto meta = reinterpret_cast<Metaslab*>(curr);
+        auto meta = Metaslab::from_link(curr);
         if (meta->needed() == 0)
         {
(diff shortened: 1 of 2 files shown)


mjp41 (approved):
LGTM

nwf:
Whoop, sorry, CI debug builds correctly pointed out that I was accidentally doing pointer arithmetic on `nullptr`.  That should be better now.

PR #388, opened 2021-09-06
Add missing `inline` from header.

davidchisnall:


mjp41 (approved):

.github/workflows/main.yml
```diff
@@ -28,9 +28,9 @@ jobs:
           - os: "ubuntu-latest"
             dependencies: "sudo apt install ninja-build"
           - os: "macos-11"
-            dependencies: "brew update && brew install ninja"
+            dependencies: "rm -f /usr/local/bin/2to3 ; brew update && brew install ninja"
```
mjp41, line 31: Can you add a comment to this hack

src/mem/metaslab.h
```diff
@@ -169,7 +169,7 @@ namespace snmalloc
     }
   };
 
-  Metaslab* Metaslab::from_link(SlabLink* lptr)
+  inline Metaslab* Metaslab::from_link(SlabLink* lptr)
```
nwf, line 172: `SNMALLOC_FAST_PATH`?

.github/workflows/main.yml
davidchisnall: Done.

src/mem/metaslab.h
davidchisnall: That may be desirable as well, the goal for this patch is to fix linker errors if two compilation units include this header file.  Any declaration in a header needs to be marked inline so that it becomes a COMDAT.  This was missed in review of #385 and broke Verona.

nwf:
LGTM

PR #389, opened 2021-09-09
WIP: Wild pointers and domestication

nwf:
This patch series introduces a notion of "wild" vs. "tame" pointers to the `CapPtr` taxonomy.  The intended use is that pointers from clients are considered wild and the `Backend` offers a domestication function.  In the traditional `globalconfig` case, this is a no-op; in the case of a sandboxed configuration, domestication gives us the opportunity to check that the pointer really is within the sandbox address space.

Not included here, but with steps towards, is the use of wild pointers when walking a message queue, so that a trusted sandbox-memory-manipulating `LocalAlloc` will additionally have to domesticate when traversing its incoming (in sandbox memory) message queue.

This also includes the promised replay of #313 on `snmalloc2`, converting the `CapPtr` annotation from a single `enum` to a cartesian product of several `enum`s.

Thoughts?


src/ds/ptrwrap.h
```diff
@@ -127,23 +196,21 @@ namespace snmalloc
 #  pragma warning(push)
 #  pragma warning(disable : 4702)
 #endif
-    constexpr explicit CapPtr(uintptr_t p) : unsafe_capptr(p) {}
+    explicit CapPtr(T* p) : unsafe_capptr(p) {}
```
ihaller, line 199: Can this not be constexpr without the type casting?

src/ds/ptrwrap.h
```diff
@@ -213,7 +269,7 @@ namespace snmalloc
    * several chunks) to be the allocation.
    */
   template<typename T>
-  inline SNMALLOC_FAST_PATH CapPtr<T, CBAllocE>
+  SNMALLOC_FAST_PATH CapPtr<T, CBAllocE>
```
ihaller, line 272: Did you mean to move the inline after SNMALLOC_FAST_PATH?

src/ds/ptrwrap.h
nwf: Good call; probably just forgot to put the keyword back during some churn.

src/ds/ptrwrap.h
nwf: Oh it's a mess and/or I don't know what I'm doing.  MSVC objects to `__forceinline inline`, which rules out `SNMALLOC_FAST_PATH inline` given the current definitions, but GCC sometimes balks at `__attribute__((always_inline))` without `inline`.  See 8fc0dcc and ideally save me from myself and C++.

mjp41: I don't like this.  `inline` is really about ODR linkage, and the macro is about adding compiler attributes for assuming this is a very common path.

nwf: Yea, I don't like it either.  Open to suggestions given the constraints. :(

nwf: Hm, OK, apparently this is all just me being overzealous with `SNMALLOC_FAST_PATH`.  Lemme rip that out and see if things get better. :)

nwf: Yea, that looks better!

src/ds/ptrwrap.h
```diff
@@ -225,7 +281,15 @@ namespace snmalloc
    */
   inline SNMALLOC_FAST_PATH void* capptr_reveal(CapPtr<void, CBAllocE> p)
   {
-    return p.unsafe_ptr();
+    return p.unsafe_capptr;
+  }
+
+  /**
+   * Dually, given a void* from the client, it's fine to call it CBAllocEW.
```
davidchisnall, line 288: Dually with respect to what?

src/mem/freelist.h
```diff
@@ -188,6 +173,63 @@ namespace snmalloc
...
+    void atomic_store_null(const FreeListKey& key)
+    {
+      atomic_next_object.store(
+        encode_next(address_cast(&next_object), nullptr, key),
+        std::memory_order_relaxed);
+    }
+
+    template<SNMALLOC_CONCEPT(capptr_bounds::c) B>
+    static CapPtr<FreeObject, B>
+    atomic_read_next(CapPtr<FreeObject, B> self, const FreeListKey& key)
```
davidchisnall, line 228: Why has this become a static method (and why is it wrapping a non-static method?)

src/mem/freelist.h
```diff
@@ -71,11 +71,14 @@ namespace snmalloc
    */
   class FreeObject
   {
+    template<bool, bool>
+    friend class FreeListBuilder;
+
     union
     {
-      CapPtr<FreeObject, CBAlloc> next_object;
+      FreeObject* next_object;
```
davidchisnall, line 79: Why can these be bare pointers?  Don't they have to be wild (except on systems with CHERI + MTE)?

src/mem/freelist.h
```diff
@@ -121,11 +119,12 @@ namespace snmalloc
...
+     *
+     * This is the internal, private version that manipulates raw pointers; see
+     * below for the wrapper that enforces CapPtr bounds.
      */
-    static CapPtr<FreeObject, CBAlloc>* store_next(
-      CapPtr<FreeObject, CBAlloc>* curr,
-      CapPtr<FreeObject, CBAlloc> next,
-      const FreeListKey& key)
+    static FreeObject**
+    store_next(FreeObject** curr, FreeObject* next, const FreeListKey& key)
```
davidchisnall, line 127: This makes me a bit nervous, shouldn't this take domesticated pointers?

src/mem/freelist.h
```diff
@@ -188,6 +173,63 @@ namespace snmalloc
...
+     * optimization for repeated snoc operations (in which
+     * next->next_object is nullptr).
+     */
+    template<SNMALLOC_CONCEPT(capptr_bounds::c) B>
+    static CapPtr<FreeObject, B>* store_next(
+      CapPtr<FreeObject, B>* curr,
+      CapPtr<FreeObject, B> next,
+      const FreeListKey& key)
+    {
+      return reinterpret_cast<CapPtr<FreeObject, B>*>(store_next(
```
davidchisnall, line 197: These reinterpret_casts make me nervous, please can you add a comment explaining why they're needed?

src/mem/freelist.h
```diff
@@ -216,16 +261,16 @@ namespace snmalloc
    *
    * Checks signing of pointers
    */
+  template<SNMALLOC_CONCEPT(capptr_bounds::c) B>
```
davidchisnall, line 264: I don't think I understand this code or what the template parameter does.  I expect this to be providing an iterator that returns domesticated pointers, over a list that may store wild pointers (and where the head of the list is domestic?)?

davidchisnall (commented):
Some comments inline.  It might come with familiarity, but I find the one-character things quite difficult to follow and kept having to look up what an E or W suffix meant.  Expanding these into short words in the names would not add much effort for typing with autocomplete and would make the code a lot more readable for someone encountering it for the first time.

src/ds/ptrwrap.h
```diff
@@ -22,96 +23,164 @@ namespace snmalloc
...
+      ConceptSame<decltype(T::platform), const platform> &&
+      ConceptSame<decltype(T::wild), const wild>;
+#endif
+    // clang-format on
+  } // namespace capptr_bounds
+
+  /*
+   * Defining these above and then `using` them to hide the namespace label like
+   * this is still shorter than defining these completely out here, since there
+   * isn't a notion of local namespace imports.
```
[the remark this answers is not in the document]
nwf: It probably makes more than might be desired visible in scope, but yes.

nwf:
This push fixes the SFINAE mechanism around `capptr_domesticate`, which was mistakenly always finding the default implementation (one of those delightful cases where mis-merging code both compiled and generally worked while being quite wrong).

A significant change here is that `RemoteAllocator::dequeue` now returns a `HeadPtr`.  We were doing all the work to domesticate `front` and then returning the wild version, which seems silly.  This patch is not yet pushed back to the offending initial commit for rebasing.

This patch series also ends with a minimal test of domestication that attempts to ensure that we are calling the callback appropriately (which doesn't compile on some of the CI bots, sigh; will fix).

That test highlights something sad about the current layering that we probably want to fix: every call to `RemoteAllocator::dequeue` domesticates one pointer ahead, to check the `signed_prev`, and then stores that back to the `front` node (re-wilding it in the process).  For `QueueHeadsAreTame` builds, this is fine, since the domesticator used for `front` will be a no-op.  For not-`QueueHeadsAreTame` builds, it might be nicer if `RemoteAllocator` offered an iterator that kept the `front` in private memory, only updating the in-shared-memory `front` value when the iterator reached its end (or was stopped early).

PR #391, opened 2021-09-14
Add memcpy with bounds checks.

davidchisnall:
The memcpy implementation is not completely stupid but is almost
certainly not as good as a carefully tuned and optimised one.

Building snmalloc with FreeBSD's libc memcpy + jemalloc and with this,
each 10 times, does not show a statistically significant performance
difference at 95% confidence.  The snmalloc version has very slightly
lower median and worst-case times.  This is in no way a sensible
benchmark, but it serves as a smoke test for significant performance
regressions.

The CI self-host job now uses the checked memcpy.

This also fixes an off-by-one error in the external bounds.  This is
triggered by ninja, so we will see breakage in CI if it is reintroduced.

In debug builds, we provide a verbose error containing the address of
the allocation, the base and bounds of the allocation, and a backtrace.

The backtrace was broken by the CI cleanup moving the BACKTRACE_HEADER
macro into the SNMALLOC_ namespace.  This is also fixed.

### DIFF
--- a/src/backend/pagemap.h
+++ b/src/backend/pagemap.h
@@ -244,7 +244,14 @@ namespace snmalloc
       }
 
       //  This means external pointer on Windows will be slow.
-      if constexpr (potentially_out_of_range && !pal_supports<LazyCommit, PAL>)
+      // TODO: With SNMALLOC_CHECK_CLIENT we should map that region read-only,
+      // not no-access, but this requires a PAL change.
+      if constexpr (
+        potentially_out_of_range
+#ifndef SNMALLOC_CHECK_CLIENT
+        && !pal_supports<LazyCommit, PAL>
+#endif
+      )
       {
         register_range(p, 1);
       }
--- a/src/override/memcpy.cc
+++ b/src/override/memcpy.cc
@@ -1,0 +1,232 @@
+#include "override.h"
+
+#include <errno.h>
+#include <stdio.h>
+#include <string.h>
+#if __has_include(<xlocale.h>)
+#  include <xlocale.h>
+#endif
+
+using namespace snmalloc;
+
+// glibc lacks snprintf_l
+#ifdef __linux__
+#  define snprintf_l(buf, size, loc, msg, ...) \
+    snprintf(buf, size, msg, __VA_ARGS__)
+// Windows has it with an underscore prefix
+#elif defined(_MSC_VER)
+#  define snprintf_l(buf, size, loc, msg, ...) \
+    _snprintf_l(buf, size, msg, loc, __VA_ARGS__)
+#endif
+
+namespace
+{
+  /**
+   * Should we check loads?  This defaults to on in debug builds, off in
+   * release (store-only checks)
+   */
+  static constexpr bool CheckReads =
+#ifdef SNMALLOC_CHECK_LOADS
+    SNMALLOC_CHECK_LOADS
+#else
+#  ifdef NDEBUG
+    false
+#  else
+    true
+#  endif
+#endif
+    ;
+
+  /**
+   * Should we fail fast when we encounter an error?  With this set to true, we
+   * just issue a trap instruction and crash the process once we detect an
+   * error. With it set to false we print a helpful error message and then crash
+   * the process.  The process may be in an undefined state by the time the
+   * check fails, so there are potentially security implications to turning this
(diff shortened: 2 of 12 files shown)


mjp41:
Sorry, could you add some test coverage.  We really need decent coverage and testing of a memcpy implementation.

It would be good to hook up some negative test coverage too.  Though, as we don't have anything like that at the moment, I would be happy to leave that to a separate PR, where we add negative test coverage for many of the security features.

davidchisnall:
We had an integration test (the self-host build now uses the checked memcpy) but I've also added a unit test.  This checks that we catch the bounds errors and that memcpy does the right thing in fun corner cases.

src/override/memcpy.cc
```diff
@@ -0,0 +1,233 @@
...
+  template<bool IsRead = false>
+  SNMALLOC_FAST_PATH inline void
+  check_bounds(const void* ptr, size_t len, const char* msg = "")
+  {
+    if constexpr (!IsRead || CheckReads)
+    {
+      auto& alloc = ThreadAlloc::get();
+      void* p = const_cast<void*>(ptr);
+
+      // FIXME: Overflow checking.
```
mjp41, line 139: Could you use `TODO` rather than `FIXME`?

src/backend/pagemap.h
```diff
@@ -244,7 +244,14 @@ namespace snmalloc
       }
 
       //  This means external pointer on Windows will be slow.
-      if constexpr (potentially_out_of_range && !pal_supports<LazyCommit, PAL>)
+      // FIXME: With SNMALLOC_CHECK_CLIENT we should map that region read-only,
```
mjp41, line 247: Could you use `TODO` rather `FIXME`.

src/override/memcpy.cc
```diff
@@ -0,0 +1,233 @@
...
+      if (unlikely(
+            pointer_offset(ptr, len) > alloc.external_pointer<OnePastEnd>(p)))
```
mjp41, line 141: Isn't the check you actually want:
```
pointer_diff(ptr, alloc.external_pointer<OnePastEnd>(p)) > len
```
then it works for overflow.

mjp41 (approved):
LGTM. Couple of minor bits

src/override/memcpy.cc
davidchisnall: I'd prefer to keep FIXME for things that are really important.  The lack of overflow checking is actually a security vulnerability here.  To me, TODO implies 'we should probably do this', FIXME means 'this is really a bug and we should fix it'.

src/override/memcpy.cc
davidchisnall: Yes, if you replace `>` with `<`, that's better.  Thanks!

src/override/memcpy.cc
```diff
@@ -0,0 +1,230 @@
...
+    {
+      char data[Size];
+    };
+    auto* d = static_cast<Block*>(dst);
+    auto* s = static_cast<const Block*>(src);
+    *d = *s;
+#endif
+  }
+
+  SNMALLOC_SLOW_PATH SNMALLOC_UNUSED_FUNCTION void crashWithMessage
```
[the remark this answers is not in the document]
davidchisnall: Could be if we have other consumers of it.  If we want to add more things with bounds checks, it definitely should be.

src/override/memcpy.cc
mjp41: I was to busy focussing on getting the `pointer_diff` argument order correct to get that right ;-)

src/override/memcpy.cc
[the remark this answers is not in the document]
mjp41: Well, there are a bunch of things that do checks

* freeing the start of object
* corruption of free list

So, giving those nicer error message would be good.  But clearly a different PR.  Let's leave that refactoring to then.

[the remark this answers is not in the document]
davidchisnall: Sounds good.  This function is quite specialised because it's trying to keep everything related to preparing the message in the slow-path function, so that we can just bail out and do all of the complicated formatting things in a separate function.  It would probably need to be refactored to be a generic that took a lambda that did the formatting if we wanted something more general, ideally with the fast-path and slow-path bits both being generic.  I'll have a think about what that should look like.

PR #392, opened 2021-09-16
Add compiler abstractions over fast fail.

davidchisnall:


src/override/memcpy.cc
```diff
@@ -1,3 +1,6 @@
+// Don't complain that _snprintf_l is unsafe.  It is only unsafe in certain
+// uses and this is not one of them.
+#define _CRT_SECURE_NO_WARNINGS
```
mjp41, line 3: Can you change the code causing the warning to use `snprintf_s`? I'm concerned that switching this off might hide other issues.

src/override/memcpy.cc
```diff
@@ -6,6 +9,11 @@
 #if __has_include(<xlocale.h>)
 #  include <xlocale.h>
 #endif
+#ifdef _MSC_VER
```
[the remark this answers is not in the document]
davidchisnall: I've changed this to an `SNMALLOC_FAST_PATH_INLINE` macro in the header and added a doc comment on each of its definitions explaining why it's defined that way.

src/override/memcpy.cc
davidchisnall: Done.

mjp41 (approved):

PR #393, opened 2021-09-27
Tidy TODOs from Free List

mjp41:
* Add extra key to freelist.  This follows the encoding Cedric suggested
  for a signature of two things. Free list key now has a pair of keys
  for encoding previous pointer. This makes it harder to extract the
  underlying keys out of the multiplication.

* Apply SFINAE to the extract_segment.


nwf (approved):
LGTM

PR #394, opened 2021-09-27
Expose notify_using_readonly

mjp41:
This exposes a readonly notify using, so that the underlying platform
can map the range of pages readonly into the application.  This improves
performance of external pointer on platforms that support lazy commit
of pages as it can access anything in the range.

### DIFF
--- a/src/backend/pagemap.h
+++ b/src/backend/pagemap.h
@@ -159,7 +159,7 @@ namespace snmalloc
 #ifdef SNMALLOC_CHECK_CLIENT
       // Allocate a power of two extra to allow the placement of the
       // pagemap be difficult to guess.
-      size_t additional_size = bits::next_pow2(REQUIRED_SIZE) * 2;
+      size_t additional_size = bits::next_pow2(REQUIRED_SIZE) * 4;
       size_t request_size = REQUIRED_SIZE + additional_size;
 #else
       size_t request_size = REQUIRED_SIZE;
@@ -178,6 +178,18 @@ namespace snmalloc
       size_t offset = get_entropy64<PAL>() & (additional_size - sizeof(T));
       auto new_body =
         reinterpret_cast<T*>(pointer_offset(new_body_untyped, offset));
+
+      if constexpr (pal_supports<LazyCommit, PAL>)
+      {
+        void* start_page = pointer_align_down<OS_PAGE_SIZE>(new_body);
+        void* end_page = pointer_align_up<OS_PAGE_SIZE>(
+          pointer_offset(new_body, REQUIRED_SIZE));
+        // Only commit readonly memory for this range, if the platform supports
+        // lazy commit.  Otherwise, this would be a lot of memory to have
+        // mapped.
+        PAL::notify_using_readonly(
+          start_page, pointer_diff(start_page, end_page));
+      }
 #else
       auto new_body = reinterpret_cast<T*>(new_body_untyped);
 #endif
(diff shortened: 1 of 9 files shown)


src/backend/pagemap.h
```diff
@@ -159,7 +159,7 @@ namespace snmalloc
 #ifdef SNMALLOC_CHECK_CLIENT
       // Allocate a power of two extra to allow the placement of the
       // pagemap be difficult to guess.
-      size_t additional_size = bits::next_pow2(REQUIRED_SIZE) * 2;
+      size_t additional_size = bits::next_pow2(REQUIRED_SIZE) * 4;
```
nwf, line 162: Unrelated change, but probably fine?

src/pal/pal_posix.h
```diff
@@ -215,6 +215,24 @@ namespace snmalloc
...
+     * Notify platform that we will be using these pages for reading.
+     *
+     * On POSIX platforms, lazy commit means that this is a no-op, unless
+     * we have initially mapped the pages as PROT_NONE.
+     */
+    static void notify_using_readonly(void* p, size_t size) noexcept
+    {
+      SNMALLOC_ASSERT(is_aligned_block<OS::page_size>(p, size));
+
+#ifdef SNMALLOC_CHECK_CLIENT
```
nwf, line 228: Conditioning on `SNMALLOC_CHECK_CLIENT` inside `notify_using_readonly` seems a little dodgy; at the moment the only call is already under that `ifdef` guard already, but I could imagine cases where we'd be wanting this not under that guard.

src/backend/pagemap.h
mjp41: Not unrelated, if the Pagemap is read-only, then you can try to slide along memory looking for metadata in the pagemap.  This changes it from being a 1/3 chance of hitting a live meta-data entry without a segfault, to 1/5 chance.  If the pages are PROT_NONE, you get one shot and will probably fail, i.e. 1/2^38 or something like that.  So just trying to reduce this as an attack vector.

src/pal/pal_posix.h
mjp41: Good point, at the moment the Pals behave differently for CHECK_CLIENT and not, with initially mapping memory readonly.  This isn't ideal. I think a larger refactor to make Pals not depend on CHECK_CLIENT makes sense, but then we will need to expose more features from the Pal interface.

mjp41: Will refactor based on Teams conversation, and add a second commit.

davidchisnall:
I think this is fine.  I'd also considered adding a permissions template parameter to `notify_using` and `reserve` (in reserve, this could replace `committed`, since no-access is equivalent to not-committed) but I'm happy with this if it's simpler / cleaner.

src/pal/pal_posix.h
mjp41: Performed a bit of refactoring to make the semantics, hopefully, clearer.

src/pal/pal_apple.h
```diff
@@ -126,12 +126,11 @@ namespace snmalloc
       while (madvise(p, size, MADV_FREE_REUSABLE) == -1 && errno == EAGAIN)
         ;
 
-#  ifdef SNMALLOC_CHECK_CLIENT
-      // This must occur after `MADV_FREE_REUSABLE`.
-      //
-      // `mach_vm_protect` is observably slower in benchmarks.
-      mprotect(p, size, PROT_NONE);
-#  endif
+      if constexpr (PalEnforceAccess)
```
davidchisnall, line 129: Braces around body of if statement?

src/pal/pal_apple.h
```diff
@@ -205,7 +205,7 @@ namespace snmalloc
     // Apple's `mmap` doesn't support user-specified alignment and only
     // guarantees mappings are aligned to the system page size, so we use
     // `mach_vm_map` instead.
-    template<bool committed>
+    template<bool state_using>
```
davidchisnall, line 208: Should this be a tri-state `enum`?  Are there cases where we want to provide read-only access to everything (e.g. the entire pagemap)?

src/pal/pal_bsd.h
```diff
@@ -35,13 +35,13 @@ namespace snmalloc
...
-#if defined(SNMALLOC_CHECK_CLIENT) && !defined(NDEBUG)
+#if !defined(NDEBUG)
       memset(p, 0x5a, size);
 #endif
       madvise(p, size, MADV_FREE);
-#ifdef SNMALLOC_CHECK_CLIENT
-      mprotect(p, size, PROT_NONE);
-#endif
+
+      if constexpr (PalEnforceAccess)
```
davidchisnall, line 43: Braces?

src/pal/pal_apple.h
mjp41: There aren't currently cases.  I propose adding if it is required.

PR #395, opened 2021-09-29
Added a sequential queue

mjp41:
This changes the slab lists to use a sequential queue.
They were previously stored in a stack.

This commit also tidies up some incomplete refactoring from the
initial snmalloc2 work.


src/ds/seqqueue.h
```diff
@@ -0,0 +1,105 @@
...
+   * Linked using the T::next field.
+   */
+  template<typename T>
+  class SeqQueue
+  {
+    static_assert(
+      std::is_same<decltype(T::next), T*>::value, "T->next must be an T*");
+
+    T* head{nullptr};
+    T** end{&head};
```
nwf, line 24: If this could be templatised on the pointer types involved, then `FreeListBuilder` could be a collection of these even with the `FreeObject` shenanigans of #389 (if we also renamed `FreeObject::T::next_object` to `FreeObject::T::next`.)

```suggestion
  template<typename T,
    template<typename> typename HPtr = Pointer,
    template<typename> typename QPtr = Pointer>
  class SeqQueue
  {
    static_assert(
      std::is_same<decltype(T::next), QPtr<T>>::value, "T->next must be a queue pointer to T");

    HPtr<T> head{nullptr};
    QPtr<T>* end{reinterpret_cast<QPtr<T>*>(&head)};
```

nwf (approved):
LGTM as is, if you don't like the additional templating suggestion.

src/ds/seqqueue.h
mjp41: To summarise, I'm happy in principle, but we need to make the `reinterpret_cast` such that the client know about it, and has to say it is okay.  This could be either by explicitly restricting the types on `HPtr` and `QPtr`, or by requiring a lambda that does the cast.

So going to leave this to your future PR.

PR #396, opened 2021-10-06
Make pointer_diff take const pointers.

davidchisnall:
It doesn't modify the objects and this avoids a const_cast in callers
that have a const pointer.


mjp41 (approved):
LGTM

PR #399, opened 2021-10-13
add LoongArch aal

SchrodingerZhu:
Just to show the portability of snmalloc to `loongarch`. 
https://loongson.github.io/LoongArch-Documentation/LoongArch-Vol1-EN 
```
$/home/schrodinger/Documents/qemu/build/qemu-loongarch64 -L $LA_PATH/../target/usr/ perf-singlethread-1
Count:  32768, Size:     16, ZeroMem: 0, Write: 0:     25995159 ns
Count:  32768, Size:     16, ZeroMem: 0, Write: 1:     23428680 ns
Count:  32768, Size:     16, ZeroMem: 1, Write: 0:     25952950 ns
Count:  32768, Size:     16, ZeroMem: 1, Write: 1:     25649431 ns
Count:  32768, Size:     32, ZeroMem: 0, Write: 0:     23379167 ns
Count:  32768, Size:     32, ZeroMem: 0, Write: 1:     23449238 ns
Count:  32768, Size:     32, ZeroMem: 1, Write: 0:     26265615 ns
Count:  32768, Size:     32, ZeroMem: 1, Write: 1:     26149518 ns
Count:  32768, Size:     64, ZeroMem: 0, Write: 0:     24059021 ns
Count:  32768, Size:     64, ZeroMem: 0, Write: 1:     23810705 ns
Count:  32768, Size:     64, ZeroMem: 1, Write: 0:     26568643 ns
Count:  32768, Size:     64, ZeroMem: 1, Write: 1:     26486249 ns
Count:  32768, Size:    128, ZeroMem: 0, Write: 0:     25350150 ns
Count:  32768, Size:    128, ZeroMem: 0, Write: 1:     25165214 ns
Count:  32768, Size:    128, ZeroMem: 1, Write: 0:     28383734 ns
Count:  32768, Size:    128, ZeroMem: 1, Write: 1:     28094443 ns
Count:   1024, Size:   4096, ZeroMem: 0, Write: 0:      2374569 ns
Count:   1024, Size:   4096, ZeroMem: 0, Write: 1:      2036175 ns
Count:   1024, Size:   4096, ZeroMem: 1, Write: 0:      2870639 ns
Count:   1024, Size:   4096, ZeroMem: 1, Write: 1:      2554837 ns
Count:   1024, Size:   8192, ZeroMem: 0, Write: 0:      5612747 ns
Count:   1024, Size:   8192, ZeroMem: 0, Write: 1:      4376410 ns
Count:   1024, Size:   8192, ZeroMem: 1, Write: 0:      5351678 ns
Count:   1024, Size:   8192, ZeroMem: 1, Write: 1:      5368780 ns
Count:   1024, Size:  16384, ZeroMem: 0, Write: 0:     11227398 ns
Count:   1024, Size:  16384, ZeroMem: 0, Write: 1:      9032074 ns
Count:   1024, Size:  16384, ZeroMem: 1, Write: 0:     11434806 ns
Count:   1024, Size:  16384, ZeroMem: 1, Write: 1:     11662212 ns
Count:   1024, Size:  32768, ZeroMem: 0, Write: 0:     21082663 ns
Count:   1024, Size:  32768, ZeroMem: 0, Write: 1:     17188197 ns
Count:   1024, Size:  32768, ZeroMem: 1, Write: 0:     24380463 ns
Count:   1024, Size:  32768, ZeroMem: 1, Write: 1:     24069601 ns
Count:   1024, Size:  65536, ZeroMem: 0, Write: 0:     40695674 ns
Count:   1024, Size:  65536, ZeroMem: 0, Write: 1:     32872525 ns
Count:   1024, Size:  65536, ZeroMem: 1, Write: 0:     48841939 ns
Count:   1024, Size:  65536, ZeroMem: 1, Write: 1:     48756758 ns
Count:   1024, Size: 131072, ZeroMem: 0, Write: 0:     84693715 ns
Count:   1024, Size: 131072, ZeroMem: 0, Write: 1:     66228266 ns
Count:   1024, Size: 131072, ZeroMem: 1, Write: 0:    105022838 ns
Count:   1024, Size: 131072, ZeroMem: 1, Write: 1:    104210255 ns
```
All multi-thread tests currently failed with segfault. But it seems that the process quits on thread creation. I may investigate it later.

### DIFF
--- a/src/snmalloc/aal/aal.h
+++ b/src/snmalloc/aal/aal.h
@@ -41,7 +41,7 @@
 #  define PLATFORM_IS_RISCV
 #endif
 
-#if defined (__loongarch__)
+#if defined(__loongarch__)
 #  define PLATFORM_IS_LOONGARCH
 #endif
 


mjp41:
Is it possible to get a stack trace?  My guess is that something about how thread local state is being initialised is interacting badly with our thread local state.

SchrodingerZhu:
@nwf loongarch seems to be a new arch released this year. (Though it is closely related to previous MIPS ISA variant maintained by Loongson.)

SchrodingerZhu:
while we may need to wait a bit longer for a qemu release with loongarch support, I would like to call for a deeper investigation on the asm I have used here.

dbar is the memory fence hint according to the manual that prevents eager reordering of instruction. I used this because I saw loongson submitted their linux kernel patches using dbar to relax the spin loops.

preld seems to fetch a single cache size of data while loongarch also provides preldx where user can specify the stride size, gap, and count of data prefetching. I don't see snmalloc previous suggests any optimal size used in its internal prefetching so I just used the default one sized by cache line. 

What you think of the above settings?

src/aal/aal_loongarch.h
```diff
@@ -0,0 +1,57 @@
+#pragma once
+#include <cstddef>
+#if _LOONGARCH_SZPTR == 64
+#  define SNMALLOC_VA_BITS_64
+#elif _LOONGARCH_SZPTR == 32
+#  define SNMALLOC_VA_BITS_32
```
xen0n, line 6: These are old-style (MIPS-style) built-in macros; better use `__SIZEOF_POINTER__` which is actually architecture-independent. (See https://github.com/loongson/LoongArch-Documentation/pull/28 for further details.)

src/aal/aal_loongarch.h
```diff
@@ -0,0 +1,57 @@
...
+#else 
+#  error "Unsupported pointer size"
+#endif
+namespace snmalloc
+{
+  /**
+   * ARM-specific architecture abstraction layer.
```
xen0n, line 13: ARM?

src/aal/aal_loongarch.h
```diff
@@ -0,0 +1,57 @@
...
+   */
+  class AAL_loongarch
```
xen0n, line 15: Better use consistent casing as with the enum variant:

```suggestion
  class AAL_LoongArch
```

src/aal/aal_loongarch.h
```diff
@@ -0,0 +1,57 @@
...
+  public:
+    /**
+     * Bitmap of AalFeature flags
+     */
+    static constexpr uint64_t aal_features =
+      IntegerPointers | NoCpuCycleCounters;
+
+    static constexpr enum AalName aal_name = LoongArch;
+
+    static constexpr size_t smallest_page_size = 0x10000;
```
xen0n, line 26: This means 64K, which is not true on LoongArch; according to kernel sources this should be 0x1000 (4K, theoretical minimum page size) or 0x4000 (16K, vast majority of systems) instead. (I'm not familiar with the codebase, at least yet, so I don't currently know which is appropriate.)

src/aal/aal_loongarch.h
```diff
@@ -0,0 +1,57 @@
...
+
+    /**
+     * On pipelined processors, notify the core that we are in a spin loop and
+     * that speculative execution past this point may not be a performance gain.
+     */
+    static inline void pause()
+    {
+      __asm__ __volatile__("dbar 0" : : : "memory");
```
xen0n, line 34: LoongArch baseline (v1.00) currently doesn't have a "pause" instruction similar to x86 `pause`, but `dbar` seems okay here.

xen0n (commented):
Just a drive-by review but as always it's great to see new faces doing LoongArch work 😄

src/aal/aal_loongarch.h
SchrodingerZhu: ouch!

src/aal/aal_loongarch.h
SchrodingerZhu: The results were based on my tests on a very early stage qemu, where `madvise` complains about page alignment for `0x1000`. But that simulator was buggy and I believe I should retest all these configurations on a recent simulator.

src/aal/aal_loongarch.h
SchrodingerZhu: This is to discourage CPU to issue mem ops. It is used mainly in some busy loops (e.g. spin locks).

SchrodingerZhu:
Just tested with snmalloc main branch and the latest `cross-tools` at https://github.com/loongson/build-tools.
Still a lot of mess. 
- The `ar` tool gives FPE.
- many operations still fail with `min_page_size= 0x1000` due to `madvise` returning EINVAL.

@xen0n is there any recent progress on improving the situation? 
```c++
#pragma once

#if __SIZEOF_POINTER__ == 8
#  define SNMALLOC_VA_BITS_64
#else
#  define SNMALLOC_VA_BITS_32
#endif

#include <cstddef>
namespace snmalloc
{
  /**
   * Loongarch-specific architecture abstraction layer.
   */
  class AAL_LoongArch
  {
  public:
    /**
     * Bitmap of AalFeature flags
     */
    static constexpr uint64_t aal_features =
      IntegerPointers | NoCpuCycleCounters;

    static constexpr enum AalName aal_name = LoongArch;

    static constexpr size_t smallest_page_size = 0x1000;

    /**
     * On pipelined processors, notify the core that we are in a spin loop and
     * that speculative execution past this point may not be a performance gain.
     */
    static inline void pause()
    {
      __asm__ __volatile__("dbar 0" : : : "memory");
    }

    /**
     * PRELD reads a cache-line of data from memory in advance into the Cache.
     * The access address is the 12bit immediate number of the value in the
     * general register rj plus the symbol extension.
     *
     * The processor learns from the hint in the PRELD instruction what type
     * will be acquired and which level of Cache the data to be taken back fill
     * in, hint has 32 optional values (0 to 31), 0 represents load to level 1
     * Cache If the Cache attribute of the access address of the PRELD
     * instruction is not cached, then the instruction cannot generate a memory
     * access action and is treated as a NOP instruction. The PRELD instruction
     * will not trigger any exceptions related to MMU or address.
     */
    static inline void prefetch(void* ptr)
    {
      __asm__ volatile("preld 0, %0, 0" : "=r"(ptr));
    }
  };

  using AAL_Arch = AAL_LoongArch;
} // namespace snmalloc

```

SchrodingerZhu:
close due to #553

PR #402, opened 2021-10-18
CHERI, take 2021.10

nwf:
Getting smaller every time; down to <500 changed lines now!  Passes most of the test harness on my desk, even (I am unaware of it failing any but I'm impatient and some of the tests seem to trigger some performance pathologies in CHERI `qemu`).  Let's see what CI thinks.

For the humans looking at this PR, this is something of a rampage and I'm happy to split the stack into multiple PRs if desired (already #401 is included here).  In rough overview, from bottom to top of the patch stack:

- NFC changes just tidying parts of the tree I was working in
- Things found while porting that change behavior but aren't really CHERI specific
- NFC changes "towards" CHERI
- refactoring `Metaslab`/`ChunkRecord` to include a `MetaCommon` that holds a high authority pointer to the chunk for CHERI
- Fixing some things the CHERI compiler noted
- Adding RISC-V and FreeBSD/RISC-V support
- The CHERI AAL and FreeBSD PAL adaptation
- Workarounds for known bugs in CheriBSD

Perhaps best reviewed by commit rather than by cumulative delta.

### DIFF
--- a/src/pal/pal_consts.h
+++ b/src/pal/pal_consts.h
@@ -14,8 +14,13 @@ namespace snmalloc
    *  - using_readonly
    *  - not_using
    * model.
+   *
+   * TODO: There is a known bug in CheriBSD that means round-tripping through
+   * PROT_NONE sheds capability load and store permissions (while restoring data
+   * read/write, for added excitement).  For the moment, just force this down on
+   * CHERI.
    */
-#ifdef SNMALLOC_CHECK_CLIENT
+#if defined(SNMALLOC_CHECK_CLIENT) && !defined(__CHERI_PURE_CAPABILITY__)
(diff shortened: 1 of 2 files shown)


nwf:
Hm.  While this appears to pass the test suite, please do not merge quite yet.  I appear to be handling `ChunkUser` chunks incorrectly -- there is some subversive amplification to `Chunk`s happening.  Will investigate.

nwf:
Cumulative differences since last review: https://github.com/microsoft/snmalloc/compare/d4e35e1..8dfbb44 ; the most semantically meaningful one is 83dc540 and the rest is just some noise of tidying (esp 1e2f7ff).  I think this take on CI addresses Alex's comment, above.

PR #412, opened 2021-11-03
ARM AAL support for Morello

nwf:
Use the "C" inline assembler constraint for capability registers on
Morello.

This gets us building and running on Morello.  Sorry to have overlooked it in #402.

### DIFF
--- a/src/aal/aal_arm.h
+++ b/src/aal/aal_arm.h
@@ -48,12 +48,12 @@ namespace snmalloc
     {
 #ifdef _MSC_VER
       __prefetch(ptr);
-#else
-#  ifdef SNMALLOC_VA_BITS_64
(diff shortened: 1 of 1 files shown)


src/aal/aal_arm.h
```diff
@@ -49,7 +49,9 @@ namespace snmalloc
 #ifdef _MSC_VER
       __prefetch(ptr);
 #else
-#  ifdef SNMALLOC_VA_BITS_64
+#  if defined(__CHERI_PURE_CAPABILITY__)
```
jrtc27, line 52: Is there a reason not to just use __builtin_prefetch for AArch64 and Morello? GCC 5.4 and Clang 9.0.0 both map it to prfm pldl1keep, which are the oldest godbolt.org has. 32-bit Arm is sadder, Clang still doesn't map it to anything and GCC looks like it might be 9.3+.

mjp41 (approved):
LGTM

PR #413, opened 2021-11-03
Minor changes

mjp41:
A few bits I found well experimenting with other stuff.


nwf:
LGTM other than `clang-format` being unhappy.

PR #415, opened 2021-11-09
Refactor use of sizeclasses

mjp41:
The primary aim for this refactor is to use a representation for
sizeclasses that uniformly covers both large and small.  This allows
certain operations such as alloc_size and external_pointer to be
uniformly implemented.

The additional types make clear which kind of sizeclass is in use.

This also tidies up the code for sizeclass based divisible by and
modulus.

It fixes a bug in rust_realloc that didn't correctly determine a realloc
was required for large classes.

### DIFF
--- a/src/mem/allocstats.h
+++ b/src/mem/allocstats.h
@@ -179,7 +179,7 @@ namespace snmalloc
 #endif
     }
 
-    void sizeclass_alloc(sizeclass_t sc)
+    void sizeclass_alloc(smallsizeclass_t sc)
     {
       UNUSED(sc);
 
@@ -189,7 +189,7 @@ namespace snmalloc
 #endif
     }
 
-    void sizeclass_dealloc(sizeclass_t sc)
+    void sizeclass_dealloc(smallsizeclass_t sc)
     {
       UNUSED(sc);
 
@@ -209,7 +209,7 @@ namespace snmalloc
 #endif
     }
 
-    void sizeclass_alloc_slab(sizeclass_t sc)
+    void sizeclass_alloc_slab(smallsizeclass_t sc)
     {
       UNUSED(sc);
 
@@ -219,7 +219,7 @@ namespace snmalloc
 #endif
     }
 
-    void sizeclass_dealloc_slab(sizeclass_t sc)
+    void sizeclass_dealloc_slab(smallsizeclass_t sc)
     {
       UNUSED(sc);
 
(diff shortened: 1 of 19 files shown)


src/mem/sizeclasstable.h
```diff
@@ -5,43 +5,127 @@
...
 
-  // Large classes range from [SUPERSLAB, ADDRESS_SPACE).// TODO
-  static constexpr size_t NUM_LARGE_CLASSES =
-    Pal::address_bits - MAX_SIZECLASS_BITS;
+    constexpr sizeclass_t(size_t value) : value(value) {}
+
+  public:
+    constexpr sizeclass_t() = default;
+
+    constexpr static sizeclass_t from_small(size_t size)
```
nwf, line 72:
```suggestion
    constexpr static sizeclass_t from_small_class(size_t small_class)
```

src/mem/sizeclasstable.h
```diff
@@ -5,43 +5,127 @@
...
+     * Takes the number of leading zero bits from the actual large size-1.
+     * See size_to_sizeclass_full
+     */
+    constexpr static sizeclass_t from_large(size_t size)
+    {
+      SNMALLOC_ASSERT(size < TAG);
+      return {size};
+    }
+
+    constexpr static sizeclass_t from_raw(size_t size)
```
nwf, line 90:
```suggestion
    constexpr static sizeclass_t from_raw(size_t raw)
```

nwf (commented):
Just naming; that must mean the rest of it is fine. :)

src/mem/allocstats.h
```diff
@@ -387,15 +387,15 @@ namespace snmalloc
...
+      //     continue;
 
-        csv << "LargeBucketedStats" << dumpid << allocatorid << (i + N)
-            << large_sizeclass_to_size(i) << large_push_count[i]
-            << large_pop_count[i] << csv.endl;
-      }
+      //   csv << "LargeBucketedStats" << dumpid << allocatorid << (i + N)
+      //       << large_sizeclass_to_size(i) << large_push_count[i]
+      //       << large_pop_count[i] << csv.endl;
+      // }
```
nwf, line 398: Is this being left in to be added back later?

nwf (commented):
Thanks for the `from_.*_class` names; I think I will trip less often with those.

Just a question about some dead? code.

src/mem/allocstats.h
mjp41: Yeah, the whole alloc_stats bit needs rewriting. I was trying not to delete code, but didn't want to fail compilation for this.

nwf (commented):
LGTM, then

nwf (approved):

PR #416, opened 2021-11-11
2021/11 CHERI followup

nwf:
* Updates the documentation on `StrictProvenance`
* fixes a test suite crash (unsure what changed; fix might not be exactly the right shape)
* moves to using the base of the provided capability rather than the cursor in `dealloc` (should we do the same for `external_pointer`)?
* adds a missed bounds setting in the address space core.  We might wish to refine the spatial `CapPtr` dimension to differentiate between chunks within the address space manager and chunks as handed out, but I haven't tried actually doing that to see if it works.

Opinions more than welcome.

### DIFF
--- a/docs/StrictProvenance.md
+++ b/docs/StrictProvenance.md
@@ -1,70 +1,77 @@
 # StrictProvenance Architectures
 
-To aid support of novel architectures, such as CHERI, which explicitly track pointer *provenance* and *bounds*, `snmalloc` makes heavy use of a `CapPtr<T, B>` wrapper type around `T*` values.
-You can view the annotation `B` on a `CapPtr<T, B>` as characterising the set of operations that are supported on this pointer, such as
+To aid code auditing and support of novel architectures, such as CHERI, which explicitly track pointer *provenance* and *bounds*, `snmalloc` makes heavy use of a `CapPtr<T, B>` wrapper type around `T*` values.
+You can think of the annotation `B` on a `CapPtr<T, B>` as capturing something about the role of the pointer, e.g.:
 
-* address arithmetic within a certain range (e.g, a `Superslab` chunk)
+* A pointer to a whole chunk or slab, derived from an internal `void*`.
+* A pointer to a particular allocation, destined for the user program
+* A putative pointer returned from the user program
+
+You can also view the annotation `B` as characterising the set of operations that are supported on this pointer, such as
+
+* nothing (because we haven't checked that it's actually a valid pointer)
+* memory access within a certain range (e.g, a chunk or an allocation)
 * requesting manipulation of the virtual memory mappings
 
-Most architectures and platforms cannot enforce these restrictions, but CHERI enables software to constrain its future use of particular pointers and `snmalloc` imposes strong constraints on its *client(s)* use of memory it manages.
+Most architectures and platforms cannot enforce these restrictions outside of static constraints, but CHERI enables software to constrain its future use of particular pointers and `snmalloc` imposes strong constraints on its *client(s)* use of memory it manages.
 The remainder of this document...
 
 * gives a "quick start" guide,
 * provides a summary of the constraints imposed on clients,
-* motivates and introduces the internal `ArenaMap` structure and the `capptr_amplify` function, and
-* describes the `StrictProvenance` `capptr_*` functions provided by the Architecture Abstraction Layer (AAL) and Platform Abstraction Layer (PAL).
+* describes the `StrictProvenance` `capptr_*` functions provided by `ds/ptrwrap.h`, the Architecture Abstraction Layer (AAL), and the Platform Abstraction Layer (PAL).
 
-## Preface
+## Limitations
 
 The `CapPtr<T, B>` and `capptr_*` primitives and derived functions are intended to guide developers in useful directions; they are not security mechanisms in and of themselves.
-For non-CHERI architectures, the whole edifice crumbles in the face of an overzealous `reinterpret_cast<>` or `unsafe_capptr` member access.
-On CHERI, these are likely to elicit capability violations, but may not if all subsequent access happen to be within bounds.
+For non-CHERI architectures, the whole edifice crumbles in the face of an overzealous `reinterpret_cast<>` or `unsafe_*ptr` call.
+On CHERI, these are likely to elicit capability violations, but may not if all subsequent access happen to be within the architecturally-enforced bounds.
 
 ## Quick Start Guide
 
+### What will I see?
+
+In practice, `CapPtr<T, B>` and the details of `B` overtly show themselves only in primitive operations or when polymorphism across `B` is required.
+(Or, sadly, when looking at compilation errors or demangled names in a debugger.)
+All the concrete forms we have found useful have layers of aliasing to keep the verbosity down: `capptr::B<T>` is a `CapPtr<T, capptr::bounds::B>` with `capptr::bounds::B` itself an alias for a `capptr::bound<...>` type-level object.
+This trend of aliasing continues into higher-level abstractions, such as the freelist, wherein one finds, for example, `freelist::HeadPtr`, which expands to a type involving several `CapPtr`s and associated annotations.
+
 ### How do I safely get an ordinary pointer to reveal to the client?
 
-If you are...
+Neglecting platform-specific details of getting authority to address space and associating memory in the first place, almost all memory manipulated by `snmalloc` comes from the `AddressSpaceManager`.
+Its `reserve(size)` method returns a `capptr::Chunk<void>`; this pointer conveys full authority to the region of `size` at which it points.
+To derive a pointer that is suitable for client use, we must
 
-* Adding an interface like `external_pointer`, and so you have a `CapPtr<T, CBAllocE>`, `e`, whose bounds you want to *inherit* when revealing some other `CapPtr` `p`, use `capptr_rebound(e, p)` to obtain another `CapPtr<T, CBAllocE>` with address from `p`, then go to the last step here.
+* further spatially refine the pointer: adjust its offset with `pointer_offset` and use `capptr_bound<T, capptr::bounds::AllocFull>` and
+* shed address space control: use `PAL::capptr_to_user_address_control()` to convert `AllocFull` to `Alloc`.
 
-* Otherwise, if your object is...
+If no additional spatial refinement is required, because the entire chunk is intended for client use,
 
-  * an entire `SUPERSLAB_SIZE` chunk or bigger, you should have in hand a `CapPtr<T, CBChunk>` from the large allocator.  Use `capptr_export` to make a `CapPtr<T, CBChunkE>`, then use `capptr_chunk_is_alloc` to convert that to a `CapPtr<T, CBAllocE>`, and then proceed.  (If, instead, you find yourself holding a `CapPtr<T, CBChunkD>`, use `capptr_chunk_from_chunkd` first.)
+* shed address space control: use `PAL::capptr_to_user_address_control()` to obtain a `ChunkUser`-bounded pointer, then
+* use `capptr_chunk_is_alloc` to capture intent, converting `ChunkUser` to `Alloc` without architectural consequence.
 
-  * of size `sz` and smaller than such a chunk,
-
-    * and have a `CapPtr<T, CBChunkE> p` in hand, use `Aal::capptr_bound<T, CBAllocE>(p, sz)` to get a `CapPtr<T, CBAllocE>`, and then proceed.
-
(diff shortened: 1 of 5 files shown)


docs/StrictProvenance.md
```diff
@@ -1,71 +1,75 @@
...
-* Begin by wrapping it with `CapPtr<T, CBAllocE>` and avoid using the raw `T*` thereafter.
+Nevertheless, if adding a new kind of deallocation, we suggest following the existing flows when given a `void* p_raw` from the client:
 
-* An `CapPtr<T, CBArena>` can be obtained using `large_allocator.capptr_amplify()`.
-Note that this pointer and its progeny are *unsafe* beyond merely having elevated authority: it is possible to construct and dereference pointers with types that do not match memory, resulting in **undefined behavior**.
+* Begin by calling `p_wild = capptr_from_client(p_raw)` to annotate it as `AllocWild` and avoid using the raw form thereafter.
 
-* Derive the `ChunkMapSuperslabKind` associated with the putative pointer from the client, by reading the `ChunkMap`.
-In some flows, the client will have made a *claim* as to the size (class) of the object which may be tentatively used, but should be validated (unless the client is trusted).
+* Check the `Wild` pointer for domestication with `p_tame = capptr_domesticate<SharedState>(state_ptr, p_wild)`; `p_tame` will be a `capptr::Alloc` and will alias `p_wild` or will be `nullptr`.
```
mjp41, line 63: I was thinking we could just access the page map without checking `wild`ness, because it covers the whole of the address space.  I guess that is not the case, as it may only be backing a 48bit address space, but can take a 64bit pointer. So we could potentially access out of bounds for the pagemap.  We could mask all accesses by the address space, which would just be a single additional instruction, but it would be on the fast path of deallocation.

mjp41 (approved):
LGTM

src/mem/localalloc.h
```diff
@@ -459,6 +464,14 @@ namespace snmalloc
       //  before init, that maps null to a remote_deallocator that will never
       //  be in thread local state.
 
+#  ifdef __CHERI_PURE_CAPABILITY__
+      // On CHERI platforms, snap the provided pointer to its base, ignoring
+      // any client-provided offset, which may have taken the pointer out of
+      // bounds.  The base, on the other hand, is guaranteed by monotonicity
+      // to be within the bounds originally returned by alloc().
```
jrtc27, line 471: It can be one-past-the-end though if the capability is zero length; will that be handled correctly?

nwf: Hrm.  Inasmuch as it's undefined behaviour, yes, but no, that's not what I meant to have happen.  (Such one-past-the-end, zero-length capabilities are also revoked with the wrong object, but since they confer no authority, we've let them slide.  Here, though, this won't work, since we, `snmalloc`, are about to amplify based on the provided address.)

The correct invariant, I think, is that, if the wild cap is indeed a descendant of one of ours, its `base + length - 1` always points within the original allocation.  Sadly, we don't have anything shorter than `CGetLength` `addi -1` `CSetOffset` to get there, I think.  This might have tangential bearing on the discussion around https://github.com/CTSRD-CHERI/cheri-architecture/pull/37 or the old set-offset-relative-to-the-other-end instruction that was proposed.  Until then, here's the three instruction sequence.

That, in turn, brings up the whole "we should actually be reconstructing caps so as to restore any shed permissions" thing.  I made a start at it but I'll push that through more tomorrow.  In the interim, we could keep this hunk as an incomplete improvement or just wait for that.

jrtc27: That just shifts the problem to the other end (i.e. `free(cheri_bounds_set(malloc(len), 0))` will give you a zero length cap with base pointing to the first byte and length 0, so that calculation takes you one byte _before_ the allocation).

jrtc27: Which makes me wonder if Cornucopia got this right (or if it's even possible to get right...)

nwf: Cornucopia does actually get this right, inasmuch as those zero-length caps convey no authority and so don't "need" to be revoked per se in order to preserve the invariant that all authority to the address space is removed before the end of the revocation.  It's perhaps a little surprising that some caps might survive that exercise, but go ahead and try to use one, right?

I suppose I'd be happy to reject length-zero caps early in `free` since that can't possibly be what you meant.  And then `base` is always in bounds and we're good, right?

nwf: (Though I suppose we should write down somewhere that length-zero caps are not to be used as tokens for anything.)

jrtc27: > Cornucopia does actually get this right, inasmuch as those zero-length caps convey no authority and so don't "need" to be revoked per se in order to preserve the invariant that all authority to the address space is removed before the end of the revocation. It's perhaps a little surprising that some caps might survive that exercise, but go ahead and try to use one, right?

Just don't use them as integer type tokens. Which is a good idea anyway, since someone else can come and derive a cap to the same end if their allocation happens to be adjacent to yours. Anything in the middle should be totally safe.

> I suppose I'd be happy to reject length-zero caps early in `free` since that can't possibly be what you meant. And then `base` is always in bounds and we're good, right?

You'd need to be careful about `malloc(0)`; if it returns NULL then everything's fine, but if it returns a unique non-NULL value you need to not leak those underlying allocations.

mjp41: `malloc(0)` currently returns a fresh allocation.  I found better compatibility with that.  An old version of Clang checked for `nullptr` return and said "Out of Memory", but was requested a size of 0.

It is fairly easy to change to returning `nullptr`, we could make that a configuration option?

nwf: I think we're fine as it stands re: `malloc(0)`.  To ensure that we never hit unrepresentable bounds due to capability compression, we always bound to the underlying sizeclass size rather than the requested size.

jrtc27: That will fail cheribsdtest. Bounds must be as tight as they can be.

mjp41: If you bound tightly, you have to rebound on free.  For the case where you don't have any UAF Cornucopia like thing.  I will add things to a queue using the internal space of the deallocated object.  So
```C++
  free(malloc(7))
```
will currently fail for CHERI, as I will write to the first 16 bytes of the allocation. To create the `next` and the `signed_prev` pointers.  We are currently make a strong assumption of the first two pointers being accessible.  @nwf you can probably correct me here.  Also, is the min allocation on CHERI 32 bytes?

davidchisnall: > That will fail cheribsdtest. Bounds must be as tight as they can be.

I don't think that this is required by the CHERI C abstract machine and it is actively harmful as a thing to test because it constrains the implementation space in a way that is surprising to C programmers.

jrtc27: Globals and stack allocations are always tightly bounded, it'd be more surprising to CHERI C programmers if malloc were not consistent with that. Yes, it's constraining, no, it's not strictly required for spatial safety, and no, we don't sufficiently specify what the bounds you actually get from malloc are, but when your hardware can represent bounds of 1 it's rather sad for malloc(1) not to have such bounds.

jrtc27: Yes, the realloc issue is awkward, but that's a problem with people assuming things that the C standard doesn't quite say is true.

malloc_usable_size, well, people are abusing it. The Linux man-pages project very clearly has:

> The value returned by malloc_usable_size() may be greater than the requested size of the allocation because of alignment and minimum size constraints. Although the excess bytes can be overwritten by the application without ill effects, this is not good programming practice: the number of excess bytes in an allocation depends on the underlying implementation.
> 
> The main use of this function is for debugging and introspection.

I also don't see many uses of it according to codesearch.debin.net (which needs a lot of filtering to avoid picking up all the embedded dlmalloc's and jemalloc's).

Yes, people may be more prone to forgetting bounds aren't always precise, but I don't think "being stricter about X means people forget to be careful about Y therefore we shouldn't be strict about X" is right. Especially when it's already true for globals and stack allocations. Also, if you make it dependent on whatever internal buckets the allocator has then you can end up with more surprise, because code that works on one machine may hit a bounds fault on another just because it happened to have an allocator with a bucket size closer to what you asked for.

The test in question was added in May 2016, so it's nothing new https://github.com/CTSRD-CHERI/cheribsd/commit/beef09dc28b18676a5f40987210a8397877c1853

mjp41: So my concern from chatting with Wes to understand Jessica's comments about `malloc(0)` is that we have to implement `malloc(0)` with its own sizeclass.

Here are my assumptions
* We don't want to store a per-allocation size, it is too costly.
   -  Hence, we can't check that you return the same bounds you requested
* To meet the tight bounds of malloc,  `malloc(0)` has 0 bounds
* malloc(0) is backed by an underlying allocation of 32bytes.

I think from @jrtc27's comments, we could do the following
```
auto p = malloc(32);
auto e = pointer_offset(p, 32);
auto z = restrict_bounds(e, 0);
```
So here `z` looks like a valid allocation of the next object in the slab with size 0.  This could now be used to `free` an unrelated allocation.

This means if we are to implement tight bounds, then we will need to alter the design of snmalloc to have a different slab for 0 sized allocations.  This is doable, and isoalloc does this.  But it is a constraint on the design of snmalloc.

Personally, I prefer giving the larger bounds, and you have to return the full original bounds to `free`.  I read that the argument of `free`:

> must have been returned by a previous call to malloc(), calloc() or realloc().

Personally, I view changing the bounds in the same was as offsetting from the base, not something `free` has to deal with.

That said, I am happy to take a patch that gives 0 its own sizeclass, but it will be of dubious benefit outside of conformance for CHERI.

mjp41: > The test in question was added in May 2016, so it's nothing new CTSRD-CHERI/cheribsd@beef09d

@jrtc27 I took a look at that test but it is a `<` not a `!=`.  So it looks like a larger bounds conforms to me?

jrtc27: Interesting, in all these years I'd never noticed that

mjp41: No worries.  This has been a super useful discussion for me.  Understanding the corner cases around amplifying a zero permission are super interesting.

nwf:
@mjp41, mind having another look?  Beyond merely rebasing, this push replaces the "avoid traps on nullptr" commit with https://github.com/microsoft/snmalloc/pull/416/commits/7dea60885f293944a1e90104a3886d33747f2ec0 which should be a little more general and elide unnecessary branches on Morello, and adds many more tests about CHERI properties, along with documentation about alternatives and the CHERI+MTE work, in https://github.com/microsoft/snmalloc/pull/416/commits/b70f3dbe0a2436cbfea68aabab2e3ff858720b21 .

mjp41 (approved):
LGTM

docs/StrictProvenance.md
nwf: For CHERI, at least, the bounds check of the Pagemap will be implicit.  Whether you think it's worth adding the additional instruction for non-CHERI, I leave in your/Saar's hands. :)

jrtc27:
Should whether the architecture traps or tag-clears be defined in the AAL rather than as scattered ifdefs?

nwf:
> Should whether the architecture traps or tag-clears be defined in the AAL rather than as scattered ifdefs?

Oh, that's not a bad plan.  Makes it more obvious what's going on.

src/mem/localalloc.h
```diff
@@ -449,6 +449,133 @@ namespace snmalloc
...
+       * above, at the cost of not catching as many misbehaving clients.
+       *
+       * In either case, having bounded by CRRL bounds, we would need to be
+       * *reconstructing* the capabilities headed to our free lists to be given
+       * out to clients again; there are many more CRRL classes than snmalloc
+       * sizeclasses (this is the same reason that we can always get away with
+       * CSetBoundsExact in capptr_bound).  Switching to CRRL bounds, if that's
+       * ever a thing we want to do, will be easier after we've done the
+       * plumbing for CHERI+MTE.
+       */
```
brooksdavis, line 559: In the CheriBSD jemalloc adaptation we do return zero-sized caps for `malloc(0)`. We also use CRRL bounds. This does complicate potential checks of capabilities passed to free as one can only check if the length is plausible unless a set of valid lengths is stored somewhere. This should probably be revisited.

src/backend/address_space_core.h
```diff
@@ -310,8 +310,11 @@ namespace snmalloc
       {
         if (rsize > size)
         {
-          add_range<PAL, Pagemap>(
-            local_state, pointer_offset(res, size), rsize - size);
+          size_t residual_size = rsize - size;
```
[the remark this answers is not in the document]
nwf: Oh, blech, yes, that's a good point; thank you for catching it.  Probably that means that we shouldn't be bounding inside the pool of free objects and should just do so when we definitely reserve space, since we'll do so with a much smaller mantissa than CHERI supports...

mjp41:
Apologies, I broke CI with the branch name change, can you rebase onto the latest `main` to get the checks to run.

mjp41:
LGTM

PR #421, opened 2021-11-18
clean up unused usages

SchrodingerZhu:

This PR makes `UNUSED` as a function marked by `[[maybe_unused]]`. This enables us to put `UNUSED` to namespace without pollute global macros.

### DIFF
--- a/src/ds/defines.h
+++ b/src/ds/defines.h
@@ -90,8 +90,6 @@
 #  define __has_builtin(x) 0
 #endif
 
-#define UNUSED(x) ((void)(x))
-
 namespace snmalloc
 {
   // Forwards reference so that the platform can define how to handle errors.
@@ -168,4 +166,6 @@ namespace snmalloc
 #else
   static constexpr bool CHECK_CLIENT = false;
 #endif
+  template <typename... Args>
+  void UNUSED(Args&& ... unused [[maybe_unused]]) {}
 } // namespace snmalloc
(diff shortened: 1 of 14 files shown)


davidchisnall:
I haven't seen this idiom before, but I *really* like it.  Unfortunately, the PR missed adding the namespaced version in a few places so it's failing CI.  Looks good once those are fixed though.

src/ds/defines.h
```diff
@@ -168,4 +166,6 @@ namespace snmalloc
 #else
   static constexpr bool CHECK_CLIENT = false;
 #endif
+  template <typename... Args>
+  void UNUSED(Args&& ... unused [[maybe_unused]]) {}
```
mjp41, line 170: Out of interest, does the following also work?
```C++
template <typename... Args>
inline void UNUSED(Args&& ...) {}
```
It is definitely unused, so I think you can just omit giving the argument pack a name.

But I love the change, awesome trick.

mjp41 (approved):
LGTM once CI is happy you have got all the namespace stuff resolved.

SchrodingerZhu:
There is one issue with msvc, however:

With always inline, GCC can elim the symbol even at debugging:
https://godbolt.org/z/5z8qE9Kqf
However, MSVC may generate some junk code:
https://godbolt.org/z/so1e4EsPj

mjp41:
I think that is fine for there to be a few artifacts in Debug builds.  I alway struggle to read the assembly unless it has been optimised anyway ;-)

PR #422, opened 2021-11-18
build fix for some 3rd party oses.

devnexen:


davidchisnall:
Huh, I thought everything except Linux added the `xlocale` APIs.  I added them to FreeBSD around 2011 and they were in Darwin before then.  Thanks for catching this!

mjp41 (approved):
LGTM, @devnexen thanks for keeping us running on so many platforms.

davidchisnall (approved):

PR #425, opened 2021-11-18
solaris systems build fix

devnexen:


src/test/func/memory/memory.cc
```diff
@@ -4,7 +4,7 @@
 #include <test/setup.h>
 #include <test/xoroshiro.h>
 #include <unordered_set>
-#if defined(__linux__) && !defined(SNMALLOC_QEMU_WORKAROUND)
+#if (defined(__linux__) || defined(__sun)) && !defined(SNMALLOC_QEMU_WORKAROUND)
```
nwf, line 7: Please update the comment below, as well?

davidchisnall: I thought that this was restricted to Linux originally because we just needed one POSIXy system to test.  If we change this to `defined(__unix__)` does it cause failures anywhere?

devnexen: ah nope this type does not exist on BSD, I think on solaris it s due to the long history of 32 bits passage to 64 bits.

davidchisnall: Sorry, my question is more why we're using the non-standard `setrlimit64` instead of the POSIX `setrlimit`.

mjp41: Let's investigate changing to `setrlimit` if it is more portable, but that is not part of this PR.  @schrodingerzhu originally wrote the test, which I reviewed.  But I didn't consider the portability of the test.  @davidchisnall  if you think it is important, do you want to raise an issue.

mjp41:
@devnexen thanks for keeping us running on a such a plethora of platforms.

PR #426, opened 2021-11-21
support allocator api

SchrodingerZhu:
export FFIs to enable snmalloc being used as a separate allocator
https://doc.rust-lang.org/stable/std/alloc/trait.Allocator.html

### DIFF
--- a/src/override/rust.cc
+++ b/src/override/rust.cc
@@ -171,10 +171,9 @@ extern "C"
   }
 
   // This function is to calculate the excessive size.
-  SNMALLOC_EXPORT size_t SNMALLOC_NAME_MANGLE(rust_round_size)(
-    size_t size, size_t alignment
-  ) {
+  SNMALLOC_EXPORT size_t
+  SNMALLOC_NAME_MANGLE(rust_round_size)(size_t size, size_t alignment)
+  {
     return round_size(aligned_size(alignment, size));
   }
-
 }
--- a/src/override/rust.cc
+++ b/src/override/rust.cc
@@ -9,47 +9,164 @@
 
 using namespace snmalloc;
 
-extern "C" SNMALLOC_EXPORT void*
-  SNMALLOC_NAME_MANGLE(rust_alloc)(size_t alignment, size_t size)
+namespace
 {
-  return ThreadAlloc::get().alloc(aligned_size(alignment, size));
-}
-
-extern "C" SNMALLOC_EXPORT void*
-  SNMALLOC_NAME_MANGLE(rust_alloc_zeroed)(size_t alignment, size_t size)
-{
-  return ThreadAlloc::get().alloc<YesZero>(aligned_size(alignment, size));
-}
-
-extern "C" SNMALLOC_EXPORT void
-  SNMALLOC_NAME_MANGLE(rust_dealloc)(void* ptr, size_t alignment, size_t size)
-{
-  ThreadAlloc::get().dealloc(ptr, aligned_size(alignment, size));
-}
-
-extern "C" SNMALLOC_EXPORT void* SNMALLOC_NAME_MANGLE(rust_realloc)(
-  void* ptr, size_t alignment, size_t old_size, size_t new_size)
-{
-  size_t aligned_old_size = aligned_size(alignment, old_size),
-         aligned_new_size = aligned_size(alignment, new_size);
-  if (
-    size_to_sizeclass_full(aligned_old_size).raw() ==
-    size_to_sizeclass_full(aligned_new_size).raw())
-    return ptr;
-  void* p = ThreadAlloc::get().alloc(aligned_new_size);
-  if (p)
+  enum class SizeChange
   {
-    std::memcpy(p, ptr, old_size < new_size ? old_size : new_size);
-    ThreadAlloc::get().dealloc(ptr, aligned_old_size);
+    Unknown,
+    Grow,
+    Shrink
+  };
+  template<ZeroMem zero_mem, SizeChange size_change>
+  SNMALLOC_FAST_PATH_INLINE void* realloc_helper(
+    Alloc& alloc,
+    void* ptr,
+    size_t old_alignment,
+    size_t new_alignment,
+    size_t old_size,
+    size_t new_size)
+  {
+    size_t aligned_old_size = aligned_size(old_alignment, old_size),
+           aligned_new_size = aligned_size(new_alignment, new_size);
+    if (
+      size_to_sizeclass_full(aligned_old_size).raw() ==
+      size_to_sizeclass_full(aligned_new_size).raw())
+      return ptr;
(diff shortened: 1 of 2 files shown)


src/override/rust.cc
```diff
@@ -9,47 +9,183 @@
...
+    // this may memset more than it needs, but doing so increases
+    // the possibility for large area to be efficiently zeroed by special
+    // methods
+    void* p = alloc.alloc<zero_mem>(aligned_new_size);
+    if (p)
+    {
+      if constexpr (size_change == SizeChange::Unknown)
+        std::memcpy(p, ptr, old_size < new_size ? old_size : new_size);
+      else if constexpr (size_change == SizeChange::Grow)
+        std::memcpy(p, ptr, old_size);
```
mjp41, line 44: Should we have an assert here, so that in a Debug build we can detect incorrect usage?

src/override/rust.cc
```diff
@@ -9,47 +9,183 @@
...
+
+  /*
+   * Custom heap allocator (for `allocator_api`)
+   * TODO: consider fallible allocation
+   */
+
+  SNMALLOC_EXPORT Alloc* SNMALLOC_NAME_MANGLE(rust_allocator_new)(void)
+  {
+    auto scope_allocator = ScopedAllocator{};
+    auto alloc = new (scope_allocator->alloc<sizeof(Alloc)>()) Alloc;
```
mjp41, line 100: If you tighten the scope here, then you are more likely to get the same underlying allocator in the `init`:
```suggestion
    void* block;
    { 
       auto scope_allocator = ScopedAllocator{};
       block = scope_allocator->alloc<sizeof(Alloc)>();
    }
    auto alloc = new (block) Alloc;
```

src/override/rust.cc
```diff
@@ -9,47 +9,183 @@
...
+    alloc->init();
+    return alloc;
+  }
+
+  SNMALLOC_EXPORT void SNMALLOC_NAME_MANGLE(rust_allocator_drop)(Alloc* alloc)
+  {
+    auto scope_allocator = ScopedAllocator{};
+    alloc->teardown();
```
mjp41, line 108: This is likely to use the same allocator for the deallocation. As it will be available, when the scoped allocator is initialised.
```suggestion
    alloc->teardown();
    auto scope_allocator = ScopedAllocator{};
```

mjp41 (commented):
So I am wondering if we should add a test to give basic coverage of this API in the snmalloc repo, so we detect breaks before it reaches snmalloc-rs?

src/override/rust.cc
```diff
@@ -41,14 +41,20 @@ namespace
...
+      {
+        SNMALLOC_ASSERT(new_size <= old_size);
         std::memcpy(p, ptr, new_size);
+      }
       alloc.dealloc(ptr, aligned_old_size);
     }
     return p;
   }
-}
+} // namespace
```
mjp41, line 57: This comment looks suspect

SchrodingerZhu: oh, that actually comes from `clang-tidy`, who was ready to add namespace indicator at the ending of the scope.

mjp41:
So on 

> https://github.com/microsoft/snmalloc/runs/4311202740?check_suite_focus=true

I have a Re-runall checks:

image**

I'll hit it for you.  Do you need some help with the Windows tests?

SchrodingerZhu:
Ha, I guess sth interesting is happening on Win. I will have a look later when I can clear some space for MSVC on my PC ;).

mjp41:
Running your test gives:
```
        _STL_ASSERT(_Left == _Right, "containers incompatible for swap");
```
failing on 
```
  std::swap(cloned, moved);
```

src/test/func/rust/test.cc
```diff
@@ -0,0 +1,127 @@
...
+
+template<class T1, class T2>
+bool operator==(const RAllocator<T1>&, const RAllocator<T2>&) noexcept
+{
+  return true;
+}
+
+void test_allocator_vector()
+{
+  using Inner = std::vector<size_t, RAllocator<size_t>>;
```
mjp41, line 85: Could you add some comments to this test explaining what it is testing?

src/override/rust.cc
```diff
@@ -9,47 +9,199 @@
 
 using namespace snmalloc;
 
-extern "C" SNMALLOC_EXPORT void*
-  SNMALLOC_NAME_MANGLE(rust_alloc)(size_t alignment, size_t size)
+namespace
 {
-  return ThreadAlloc::get().alloc(aligned_size(alignment, size));
-}
+  enum class SizeChange
```
davidchisnall, line 14: Please can you add a comment explaining (and, ideally, linking to) the Rust APIs that this is intended to expose?

src/override/rust.cc
```diff
@@ -9,47 +9,199 @@
...
+    size_t old_size,
+    size_t new_alignment,
+    size_t new_size)
+  {
+    size_t aligned_old_size = aligned_size(old_alignment, old_size),
+           aligned_new_size = aligned_size(new_alignment, new_size);
+    if (
+      size_to_sizeclass_full(aligned_old_size).raw() ==
+      size_to_sizeclass_full(aligned_new_size).raw())
+      return ptr;
```
davidchisnall, line 34:
```suggestion
    {
      return ptr;
    }
```

As a general style point we should have braces around all `if` statement bodies because it reduces bugs introduced when someone edits the code later, but this case is a particularly good example of needing them for readability because the wrapped line in the condition is at the same indent level as the body of the conditiona.

src/override/rust.cc
```diff
@@ -9,47 +9,199 @@
...
+    // this may memset more than it needs, but doing so increases
+    // the possibility for large area to be efficiently zeroed by special
```
davidchisnall, line 36: Those methods are exposed in the PAL, they could be called here.

SchrodingerZhu:
Sure! I am still sorting out the design.

SchrodingerZhu:
@mjp41 @davidchisnall I begin to wonder whether we should only expose sth like `sn_rust_grow`, `sn_rust_shrink`. 

I wanted to support sth like:
```rust
 fn allocator_supports_vector() {
        let allocator = SnAllocator::new();
        let mut vec = std::vec::Vec::new_in(&allocator);
        let mut sum: usize = 0;
        for i in 1..512usize {
            vec.push(i);
            sum += i * i;
        }

        let res = vec.into_iter().flat_map(|x| {
            let mut v = std::vec::Vec::new_in(&allocator);
            for _ in 0..x {
                v.push(x);
            }
            v
        }).sum();

        assert_eq!(sum, res);
    }
```
But I begin to wonder whether it is a bad decision to let user use a `Alloc*` instead of make `Alloc` thread_local.

mjp41:
@schrodingerzhu it this still something you are interested, or should I close the PR?

PR #429, opened 2021-11-24
Modify heuristic for adding new slabs.

mjp41:
If there is only one slab remaining, then we probabalisticly allocator a
new one. If a slab is barely in use, then this could cause us to
effectively double the number of slabs in use.

This commit checks if the remaining slab has enough remaining elements
to provide randomisation.

### DIFF
--- a/src/mem/freelist.h
+++ b/src/mem/freelist.h
@@ -740,6 +740,25 @@ namespace snmalloc
         UNUSED(domesticate);
 #endif
       }
+
+      /**
+       * Returns length of free list if you call close.
+       *
+       * The Builder is not always emptied when you call close,
+       * this returns the length of the list if you had just called
+       * close.
+       */
+      [[nodiscard]] size_t length_after_close() const
+      {
+        if constexpr (RANDOM)
+        {
+          return length[0] < length[1] ? length[0] : length[1];
+        }
+        else
+        {
+          return 0;
+        }
+      }
     };
   } // namespace freelist
 } // namespace snmalloc
--- a/src/mem/freelist.h
+++ b/src/mem/freelist.h
@@ -742,22 +742,17 @@ namespace snmalloc
       }
 
       /**
-       * Returns length of free list if you call close.
+       * Returns length of the shorter free list.
        *
-       * The Builder is not always emptied when you call close,
-       * this returns the length of the list if you had just called
-       * close.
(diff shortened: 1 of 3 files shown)


src/mem/freelist.h
```diff
@@ -740,6 +740,25 @@ namespace snmalloc
...
       }
+
+      /**
+       * Returns length of free list if you call close.
+       *
+       * The Builder is not always emptied when you call close,
+       * this returns the length of the list if you had just called
+       * close.
+       */
+      [[nodiscard]] size_t length_after_close() const
```
nwf, line 751:
```suggestion
      [[nodiscard]] size_t length_available() const
```

src/mem/freelist.h
```diff
@@ -740,6 +740,25 @@ namespace snmalloc
         UNUSED(domesticate);
 #endif
       }
+
+      /**
+       * Returns length of free list if you call close.
+       *
+       * The Builder is not always emptied when you call close,
+       * this returns the length of the list if you had just called
+       * close.
```
nwf, line 749:
```suggestion
       * this returns the length of the list that would be obtained
       * if `close` were called now.
```

nwf (commented):
LGTM, just a nit about naming (I don't think I like `length_after_close` but I could be convinced) and the exact wording of a comment.

src/mem/freelist.h
mjp41: I wasn't happy with the name.  I was trying to make something that made sense for both `RANDOM` and `!RANDOM`.  I have changed to using SFINAE, so it only exists for `RANDOM` and I think it is clearer.

nwf: And RANDOM is synonymous with defined(SNMALLOC_CHECK_CLIENT)?  Yes, if somewhat circuitously: the `Metaslab` `free_queue` field is a `freelist::builder` whose `RANDOM` template parameter is derived from `SNMALLOC_CHECK_CLIENT`.  Sounds good.

nwf (approved):

src/mem/freelist.h
mjp41: RANDOM is one feature that is enabled with SNMALLOC_CHECK_CLIENT.  At some point I would like to pull them apart, but now yet ;-)

PR #431, opened 2021-11-26
add SYS_getrandom fallback for posix PAL

SchrodingerZhu:


src/pal/pal_posix.h
```diff
@@ -342,6 +349,30 @@ namespace snmalloc
...
+        int cs;
+        auto current = std::begin(buffer);
+        auto target = std::end(buffer);
+        pthread_setcancelstate(PTHREAD_CANCEL_DISABLE, &cs);
+        while (current != target)
+        {
+          auto ret = syscall(SYS_getrandom, current, target - current, 0);
+          if (ret < 0)
+          {
+            if (errno == EINTR)
```
mjp41, line 367: Could this cause an infinite loop?  e.g. if we get an `ENOSYS`.  Should we hard error on that as it is not the expected behaviour?

src/pal/pal_posix.h
```diff
@@ -342,6 +349,30 @@ namespace snmalloc
...
+#elif defined(SYS_getrandom)
+        union
+        {
+          uint64_t result;
+          char buffer[sizeof(uint64_t)];
+        };
+        int cs;
+        auto current = std::begin(buffer);
+        auto target = std::end(buffer);
+        pthread_setcancelstate(PTHREAD_CANCEL_DISABLE, &cs);
```
mjp41, line 361: Why is this required?  I am not familiar with calling syscall directly.

src/pal/pal_posix.h
```diff
@@ -128,6 +133,8 @@ namespace snmalloc
     static constexpr uint64_t pal_features = LazyCommit | Time
 #if defined(SNMALLOC_PLATFORM_HAS_GETENTROPY)
       | Entropy
+#elif defined(SNMALLOC_PLATFORM_HAS_SYS_GETRANDOM)
```
mjp41, line 136: Might be nicer to use `||`

davidchisnall:
Doing a raw system call in portable code makes me *really* nervous - just because the syscall number `#define` exists doesn't mean that the arguments are what we expect or even stable (see: Apple breaking every Go program a few years ago by changing the arguments of their `gettimeofday` system call).

What platform(s) is this for?  If it's a single one, I'd much rather we added it to that platform's PAL and avoided trying to detect that it exists.  If it's more than one then this could be a separate class that PALs could `#include` and inherit from, but then we know that the target platform actually is expected to support the system call, with the argument layout.

SchrodingerZhu:
there are some old settings on CentOS 7 where the kernel is 4＋ but the libc is 2.17 ( To be clear, I somewhat hate the situation to support these ancient stuffs but CentOS 7 might still be one of the most prevalent OS in production environment). The problem is interfaces like getrandom/getentropy is not provided in glibc.

I ported the code pieces mimicing musl and mimalloc. In mimalloc, daanx makes the function to check whether syscall is functional and if it fails it will never use it again.

I notice a problem in my code here:  If the problem comes from linux side, maybe it is better to restrict this part to Linux and the stability can be improved a bit.

davidchisnall:
If the `getrandom` system call is available on all Linux kernels currently supported, then I think it makes sense to put this unconditionally in the Linux PAL (I was a bit confused because I thought Linux syscall numbers were usually in the `__NR_*` space not the `SYS_*` space).

SchrodingerZhu:
> If the getrandom system call is available on all Linux kernels currently supported, then I think it makes sense to put this unconditionally in the Linux PAL (I was a bit confused because I thought Linux syscall numbers were usually in the __NR_* space not the SYS_* space).

Ha, that is actually quite interesting:
```c++
#ifndef _SYSCALL_H
# error "Never use <bits/syscall.h> directly; include <sys/syscall.h> instead."
#endif

#define __GLIBC_LINUX_VERSION_CODE 330240

#ifdef __NR_FAST_atomic_update
# define SYS_FAST_atomic_update __NR_FAST_atomic_update
#endif

...
```

src/pal/pal_posix.h
SchrodingerZhu: this implementation comes from musl (I guess it is used to make sure each getentropy call should not be broken in the middle) but yes, I think we can skip this part.

src/pal/pal_linux.h
```diff
@@ -7,6 +7,11 @@
 #  include <string.h>
 #  include <sys/mman.h>
 
+#  ifndef SNMALLOC_PLATFORM_HAS_GETENTROPY
```
davidchisnall, line 10: Do we actually need the CMake check?  Linux exposes the syscall numbers as macros, so we should be able to just include the header and check if it defines the syscall that we expect.  We could probably do this unconditionally for Linux, since I think any Linux version that is currently supported, including LTS kernels, will support this system call.

CMakeLists.txt
```diff
@@ -196,6 +195,8 @@ add_as_define(USE_SNMALLOC_STATS)
 add_as_define(SNMALLOC_QEMU_WORKAROUND)
 add_as_define(SNMALLOC_CI_BUILD)
 add_as_define(SNMALLOC_PLATFORM_HAS_GETENTROPY)
+add_as_define(SNMALLOC_PLATFORM_HAS_SYS_GETRANDOM)
```
davidchisnall, line 198: This doesn't seem to be set anywhere anymore?

src/pal/pal_linux.h
```diff
@@ -85,6 +94,42 @@ namespace snmalloc
         madvise(p, size, MADV_FREE);
       }
     }
+
+    static uint64_t get_entropy64()
+    {
+#  ifdef SNMALLOC_PLATFORM_HAS_GETENTROPY
+      uint64_t result;
+      if (getentropy(&result, sizeof(result)) != 0)
+        error("Failed to get system randomness");
```
davidchisnall, line 103: Not a new bug, but this can cause deadlocks during early init.  If the system call fails then the POSIX PAL calls libc functions that can require malloc.  We should modify the implementation of `error` to use `write`.

```suggestion
      {
        error("Failed to get system randomness");
      }
```

src/pal/pal_linux.h
```diff
@@ -85,6 +94,42 @@ namespace snmalloc
...
+      auto target = std::end(buffer);
+      while (auto length = target - current)
+      {
+        auto ret = syscall(SYS_getrandom, current, length, 0);
+        if (ret < 0)
+        {
+          if (errno == EINTR)
+            continue;
+          else
+            error("Failed to get system randomness");
```
davidchisnall, line 125:
```suggestion
          if (errno == EINTR)
          {
            continue;
          }
          else
          {
            error("Failed to get system randomness");
          }
```

SchrodingerZhu:
@davidchisnall will the following the code improve the situation of initialisation failure?
```diff
diff --git a/src/pal/pal_posix.h b/src/pal/pal_posix.h
index 4c8dff2..ca45731 100644
--- a/src/pal/pal_posix.h
+++ b/src/pal/pal_posix.h
@@ -151,10 +151,10 @@ namespace snmalloc
       constexpr int SIZE = 1024;
       void* buffer[SIZE];
       auto nptrs = backtrace(buffer, SIZE);
-      fflush(stdout);
+      fsync(STDOUT_FILENO);
       backtrace_symbols_fd(buffer, nptrs, STDOUT_FILENO);
-      puts("");
-      fflush(stdout);
+      write(STDOUT_FILENO, "\n", 1);
+      fsync(STDOUT_FILENO);
 #endif
     }
 
@@ -163,7 +163,12 @@ namespace snmalloc
      */
     [[noreturn]] static void error(const char* const str) noexcept
     {
-      puts(str);
+      /// by this part, the allocator is failed; so we cannot assume
+      /// subsequent allocation will work.
+      /// @attention: since the program is failing, we do not guarantee that
+      /// preious bytes in stdout will be flushed
+      write(STDOUT_FILENO, "\n", 1);
+      write(STDOUT_FILENO, str, strlen(str));
       print_stack_trace();
       abort();
     }
```

src/pal/pal_linux.h
```diff
@@ -85,6 +86,108 @@ namespace snmalloc
...
+      // In this case, it is not a good idea to fallback to std::random_device:
+      // 1. it may want to use malloc to create a buffer, which causes
+      // reentrancy problem during initialisation routine.
+      // 2. some implementations also require libstdc++ to be linked since
+      // its APIs are not exception-free.
+      int flags = O_RDONLY;
+#  if defined(O_CLOEXEC)
+      flags |= O_CLOEXEC;
+#  endif
+      auto fd = open("/dev/random", flags, 0);
```
davidchisnall, line 161: This should probably be `urandom`, since we don't want it to block (and it's not the end of the world if the entropy is weak).

```suggestion
      auto fd = open("/dev/urandom", flags, 0);
```

src/pal/pal_linux.h
```diff
@@ -85,6 +86,108 @@ namespace snmalloc
...
+      // protected routine.
+      if (false == syscall_not_working.load(std::memory_order_relaxed))
+      {
+        auto current = std::begin(buffer);
+        auto target = std::end(buffer);
+        while (auto length = target - current)
+        {
+          // reading data via syscall from system entropy pool.
+          // According to both MUSL and GLIBC implementation, getentropy uses
+          // /dev/random (blocking API) as the pool which is indicated by the
```
davidchisnall, line 121: This should also be urandom.  See [myths about urandom](https://www.2uo.de/myths-about-urandom/)

src/pal/pal_linux.h
```diff
@@ -4,8 +4,12 @@
 #  include "../ds/bits.h"
 #  include "pal_posix.h"
 
+#  include <fcntl.h>
 #  include <string.h>
 #  include <sys/mman.h>
+#  if __has_include(<syscall.h>)
```
davidchisnall, line 10: What are the situations when this can't exist?

davidchisnall (commented):
Thanks, this looks good apart from the random vs urandom thing.

davidchisnall:
I think we should be using the non-blocking interface.  The only time this actually matters for the quality of the random number is *very* early in the boot (when there may not be any entropy).  If `init` is using snmalloc then it would be good if we didn't deadlock the system by refusing to spawn the process that seeds the kernel's random number generator until after we have sufficient entropy.

mjp41:
I wonder if we should also be mixing in `rdrand` if the AAL supports it.  Not part of this PR, but something to consider.  Especially for these early bring up stages.  XOR as many sources of entropy seems like a good idea.

davidchisnall:
> @davidchisnall will the following the code improve the situation of initialisation failure?

I missed this comment, sorry.  Yes, that looks great, though I think the last two `write` lines are the wrong way around (we go from a trailing newline to a newline before the output).

davidchisnall:
> I wonder if we should also be mixing in rdrand if the AAL supports it. Not part of this PR, but something to consider. Especially for these early bring up stages. XOR as many sources of entropy seems like a good idea.

I don't think so, for two reasons:

 1. Mixing entropy sources without ending up with less entropy is a non-trivial problem.  The kernel typically implements something like Yarrow or Fortuna to do this for you.
 2. If the hardware supports RDRAND, it will be one of the entropy sources used by Fortuna in the kernel, mixing it in incorrectly again will end up with less entropy than we would get otherwise.
 3. [Not a very likely reason] There are several theoretical attacks for biased entropy in hardware random number generators that let an attacker who knows a secret shared with the hardware to reduce the entropy.  The kernel's use of Fortuna mitigates this (Fortuna is designed to deal with entropy sources of different quality and, unlike Yarrow, doesn't depend on accurate estimation of entropy quality for security), we would be partially reversing this mitigation.

mjp41:
> @davidchisnall will the following the code improve the situation of initialisation failure?
> 
> ```diff
> diff --git a/src/pal/pal_posix.h b/src/pal/pal_posix.h
> index 4c8dff2..ca45731 100644
> --- a/src/pal/pal_posix.h
> +++ b/src/pal/pal_posix.h
> @@ -151,10 +151,10 @@ namespace snmalloc
>        constexpr int SIZE = 1024;
>        void* buffer[SIZE];
>        auto nptrs = backtrace(buffer, SIZE);
> -      fflush(stdout);
> +      fsync(STDOUT_FILENO);
>        backtrace_symbols_fd(buffer, nptrs, STDOUT_FILENO);
> -      puts("");
> -      fflush(stdout);
> +      write(STDOUT_FILENO, "\n", 1);
> +      fsync(STDOUT_FILENO);
>  #endif
>      }
>  
> @@ -163,7 +163,12 @@ namespace snmalloc
>       */
>      [[noreturn]] static void error(const char* const str) noexcept
>      {
> -      puts(str);
> +      /// by this part, the allocator is failed; so we cannot assume
> +      /// subsequent allocation will work.
> +      /// @attention: since the program is failing, we do not guarantee that
> +      /// previous bytes in stdout will be flushed
> +      write(STDOUT_FILENO, "\n", 1);
> +      write(STDOUT_FILENO, str, strlen(str));
>        print_stack_trace();
>        abort();
>      }
> ```

Do you want to add this into this PR?  Or shall I merge as is, and this can be picked up in a new PR by someone.

SchrodingerZhu:
@mjp41 since it is to address another issue, I think it would be better to launch a separate one.

Issue #430, opened 2021-11-25
Program stuck at initialization (4.14.0-115.el7a.0.1.aarch64)


SchrodingerZhu:
Compiler: GCC-7
```
GNU gdb (GDB) Red Hat Enterprise Linux 7.6.1-115.el7
Copyright (C) 2013 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "aarch64-redhat-linux-gnu".
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>...
Reading symbols from  <discarded>
(gdb) r
Starting program: <discarded>
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
^C
Program received signal SIGINT, Interrupt.
0x00000000037a32b8 in snmalloc::LocalAllocator<snmalloc::Globals>::init() ()
Missing separate debuginfos, use: debuginfo-install glibc-2.17-292.el7.aarch64 libatomic-8.3.1-2.1.1.el7.aarch64 libgcc-4.8.5-44.el7.aarch64 libstdc++-4.8.5-44.el7.aarch64
(gdb) bt
#0  0x00000000037a32b8 in snmalloc::LocalAllocator<snmalloc::Globals>::init() ()
[...]

PR #432, opened 2021-11-26
Add ownership checkings for FlagLock under debug

SchrodingerZhu:

### DIFF
--- a/src/ds/flaglock.h
+++ b/src/ds/flaglock.h
@@ -112,31 +112,24 @@ namespace snmalloc
   public:
     FlagLock(FlagWord& lock) : lock(lock)
     {
-#ifdef __cpp_lib_atomic_flag_test
-      do
+      while (lock.flag.test_and_set(std::memory_order_acquire))
       {
-        // test once with acquire ordering at the beginning
-        if (!lock.flag.test_and_set(std::memory_order_acquire))
-        {
-          break;
-        }
         // assert_not_owned_by_current_thread is only called when the first
         // acquiring is failed; which means the lock is already held somewhere
         // else.
         lock.assert_not_owned_by_current_thread();
-        // only need to do relaxed ordering in the busy spin loop
-        while (lock.flag.test(std::memory_order_relaxed))
+#ifdef __cpp_lib_atomic_flag_test
+        // acquire ordering because we need other thread's release to be
+        // visible. This loop is better for spin-waiting because it won't issue
+        // expensive write operation (xchg for example).
+        while (lock.flag.test(std::memory_order_acquire))
         {
           Aal::pause();
         }
-      } while (true);
 #else
-      while (lock.flag.test_and_set(std::memory_order_acquire))
-      {
-        lock.assert_not_owned_by_current_thread();
         Aal::pause();
-      }
 #endif
+      }
       lock.set_owner();
     }
 
--- a/src/ds/flaglock.h
+++ b/src/ds/flaglock.h
@@ -4,6 +4,10 @@
 
 #include <atomic>
 
+#ifndef NDEBUG
+#  include <thread>
+#endif
+
 namespace snmalloc
 {
   class FlagLock
@@ -11,16 +15,30 @@ namespace snmalloc
   private:
     std::atomic_flag& lock;
 
+#ifndef NDEBUG
+    std::thread::id owner{};
+#endif
   public:
     FlagLock(std::atomic_flag& lock) : lock(lock)
     {
       while (lock.test_and_set(std::memory_order_acquire))
+      {
+#ifndef NDEBUG
+        SNMALLOC_ASSERT(std::this_thread::get_id() != owner);
+#endif
         Aal::pause();
+      }
+#ifndef NDEBUG
+      owner = std::this_thread::get_id();
+#endif
     }
 
     ~FlagLock()
     {
       lock.clear(std::memory_order_release);
+#ifndef NDEBUG
+      owner = {};
+#endif
     }
   };
 } // namespace snmalloc
--- a/src/ds/flaglock.h
+++ b/src/ds/flaglock.h
@@ -6,7 +6,6 @@
 
 namespace snmalloc
 {
-
   struct DebugFlagWord
   {
     std::atomic_flag flag = ATOMIC_FLAG_INIT;
@@ -39,8 +38,8 @@ namespace snmalloc
 
     inline ThreadIdentity get_thread_identity()
     {
-      static thread_local int SNMALLOC_THREAD_INDENTITY = 0;
-      return &SNMALLOC_THREAD_INDENTITY;
+      static thread_local int SNMALLOC_THREAD_IDENTITY = 0;
+      return &SNMALLOC_THREAD_IDENTITY;
     }
   };
 
--- a/src/ds/flaglock.h
+++ b/src/ds/flaglock.h
@@ -18,11 +18,13 @@ namespace snmalloc
 
     void set_owner()
     {
+      SNMALLOC_ASSERT(nullptr == owner);
       owner = get_thread_identity();
     }
 
     void clear_owner()
     {
+      SNMALLOC_ASSERT(get_thread_identity() == owner);
       owner = nullptr;
     }
 
--- a/src/ds/flaglock.h
+++ b/src/ds/flaglock.h
@@ -6,8 +6,17 @@
 
 namespace snmalloc
 {
+  /**
+   * @brief The DebugFlagWord struct
+   * Wrapper for std::atomic_flag so that we can examine
+   * the re-entrancy problem at debug mode.
+   */
   struct DebugFlagWord
   {
+    /**
+     * @brief flag
+     * The underlying atomic field.
+     */
     std::atomic_flag flag = ATOMIC_FLAG_INIT;
 
     constexpr DebugFlagWord() = default;
@@ -16,18 +25,30 @@ namespace snmalloc
     constexpr DebugFlagWord(Args&&... args) : flag(std::forward<Args>(args)...)
     {}
 
+    /**
+     * @brief set_owner
+     * Record the identity of the locker.
+     */
     void set_owner()
     {
       SNMALLOC_ASSERT(nullptr == owner);
       owner = get_thread_identity();
     }
 
+    /**
+     * @brief clear_owner
+     * Set the identity to null.
+     */
     void clear_owner()
     {
       SNMALLOC_ASSERT(get_thread_identity() == owner);
       owner = nullptr;
     }
 
+    /**
+     * @brief assert_not_owned_by_current_thread
+     * Assert the lock should not be held already by current thread.
+     */
     void assert_not_owned_by_current_thread()
     {
       SNMALLOC_ASSERT(get_thread_identity() != owner);
@@ -36,8 +57,18 @@ namespace snmalloc
   private:
     using ThreadIdentity = int const*;
 
+    /**
+     * @brief owner
+     * We use a pointer to TLS field as the thread identity.
+     * std::thread::id can be another solution but it does not
+     * support `constexpr` initialisation on some platforms.
+     */
     ThreadIdentity owner = nullptr;
 
+    /**
+     * @brief get_thread_identity
+     * @return The identity of current thread.
+     */
     inline ThreadIdentity get_thread_identity()
     {
       static thread_local int SNMALLOC_THREAD_IDENTITY = 0;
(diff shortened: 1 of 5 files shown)


SchrodingerZhu:
oops, owner = {} is wrongly placed out of scope.

mjp41:
@achamayou, @ihaller I think this is a good thing to add, but I am unsure what we should do for the more constrained platforms.  Do you think is is reasonable to take a dependence on `std::this_thread::get_id();`?

Alternatively, we could effectively build our own that just depends on `thread_local`, and take the address something like:

```C++
static thread_local int for_id;

int* owner;

void set_owner()
{
  owner = &for_id;
}
```

achamayou:
@achamayou as far as I can tell from a quick local attempt to use it in our code base, and from https://github.com/openenclave/openenclave/blob/master/tests/threadcxx/enc/enc.cpp, `std::this_thread::get_id()` is supported in Open Enclave, so I have no objection.

src/ds/flaglock.h
```diff
@@ -4,22 +4,40 @@
...
+    std::thread::id owner{};
+#endif
   public:
     FlagLock(std::atomic_flag& lock) : lock(lock)
     {
       while (lock.test_and_set(std::memory_order_acquire))
+      {
+#ifndef NDEBUG
+        SNMALLOC_ASSERT(std::this_thread::get_id() != owner);
+#endif
```
davidchisnall, line 28: Does this need to be in `NDEBUG`, isn't `SNMALLOC_ASSERT` already guarded on `NDEBUG`?

davidchisnall: This test can never succeed.  Even if the lock is held by the current thread, the thing that you're looking at is a field that has, on entry to this function, been initialised to invalid and which cannot be concurrently modified.

davidchisnall (changes requested):
I'm not sure that I understand how this is used.  The `FlagLock` is an RAII object held by whoever owns the lock, so to be able to debug a case where another thread holds the lock, you need to find the thread that has the `FlagLock` on its stack and then look at the owner field.  But if you've found the stack that contains the `FlagLock` object then you already know which thread owns the lock: it's the one that you're looking at.

I think the intent here is to be able to go from the lock word to the owner and to detect attempts to recursively acquire the thread.  I believe that this needs to be done by exposing a type something like:

```c++
struct DebugFlagWord
{
  std::atomic_flag flag;
  std::thread_id owner;
  void set_owner()
  {
    owner = std::this_thread::get_id();
  }
  void clear_owner()
  {
    owner = {};
  }
  void assert_not_owned_by_current_thread()
  {
     SNMALLOC_ASSERT(std::this_thread::get_id() != owner);
  }
};
struct ReleaseFlagWord
{
  std::atomic_flag flag;
  void set_owner() {}
  void clear_owner()  {}
  void assert_not_owned_by_current_thread() {}
};
using FlagWord = std::conditional_t<SNMALLOC_DEBUG, DebugFlagWord, ReleaseFlagWord>;
```

You'd then need to modify all callers to use `FlagLock::FlagWord` instead of `std::atomic_flag`.

SchrodingerZhu:
Thanks for review. It is indeed a silly implementation here. I will correct it ASAP. 


#430 is the context of the PR. I am hoping to give reentrant problem an explicit error.

ihaller:
I am okay with a dependency on `std::this_thread::get_id()`.

SchrodingerZhu:
I end up with using `static thread_local int for_id;` approach. The problem for `std::thread::id` is that it is not constant initialized on some platform.

src/ds/flaglock.h
```diff
@@ -6,21 +6,73 @@
...
+      owner = get_thread_identity();
+    }
+    void clear_owner()
+    {
+      owner = nullptr;
+    }
+    void assert_not_owned_by_current_thread()
+    {
+      SNMALLOC_ASSERT(get_thread_identity() != owner);
+    }
```
mjp41, line 35: I find the lack of new lines hard to read.
```suggestion
    std::atomic_flag flag = ATOMIC_FLAG_INIT;
    ThreadIdentity owner = nullptr;

    constexpr DebugFlagWord() = default;

    template<typename... Args>
    constexpr DebugFlagWord(Args&&... args) : flag(std::forward<Args>(args)...)
    {}

    void set_owner()
    {
      owner = get_thread_identity();
    }

    void clear_owner()
    {
      owner = nullptr;
    }

    void assert_not_owned_by_current_thread()
    {
      SNMALLOC_ASSERT(get_thread_identity() != owner);
    }
```

src/ds/flaglock.h
```diff
@@ -6,21 +6,73 @@
 
 namespace snmalloc
 {
+  using ThreadIdentity = int const*;
+  inline ThreadIdentity get_thread_identity()
+  {
+    static thread_local int SNMALLOC_THREAD_INDENTITY = 0;
+    return &SNMALLOC_THREAD_INDENTITY;
+  }
```
mjp41, line 14: Could you nest this inside `DebugFlagWord` as a private static method.

src/ds/flaglock.h
```diff
@@ -6,21 +6,73 @@
...
+    template<typename... Args>
+    constexpr DebugFlagWord(Args&&... args) : flag(std::forward<Args>(args)...)
+    {}
+    void set_owner()
+    {
+      owner = get_thread_identity();
+    }
+    void clear_owner()
+    {
+      owner = nullptr;
```
mjp41, line 30: It shouldn't mater, because a programmer shouldn't be able to get it wrong, but it feels natural to add an assert here:
```suggestion
      SNMALLOC_ASSERT(get_thread_identity() == owner);
      owner = nullptr;
```

mjp41 (approved):

davidchisnall (commented):
The code looks good, thank you for working on this.  The load of `owner` in the assert function could probably use a weaker ordering but I think it's fine for debug builds.

Please can you add doc comments (`/** ... */`) on the new classes, methods, and fields?

SchrodingerZhu:
I did some extra change on the existence of `atomic_flag_test`. Presumably, it should reduce strict ordering atomic ops when busy spinning. Please let me know if you don't like the change and I will undo that part.

https://godbolt.org/z/vehTKEPxj

SchrodingerZhu:
So I experiments with the following program:
```
#include <iostream>
#include <atomic>
#include <chrono>
#include <thread>
#include <vector>

std::atomic_flag word = ATOMIC_FLAG_INIT;
constexpr int limit = 10000000;
int counter = 0;

__attribute__((always_inline)) inline void lock1() {
    while (true) {
        if (!word.test_and_set(std::memory_order_acquire)) {
            break;
        }
        while (word.test(std::memory_order_relaxed)) {
            asm volatile("pause\n":: :"memory");
        }
    }
}

__attribute__((always_inline)) inline void lock2() {
    while (word.test_and_set(std::memory_order_acquire)) {
        asm volatile("pause\n":: :"memory");
    }
}

__attribute__((always_inline)) inline void unlock() {
    word.clear(std::memory_order_release);
}

__attribute__((noinline)) void operation(const std::chrono::high_resolution_clock::time_point start) {
    if ((++counter % 1000) == 0) {
        std::cout << counter << ": " <<
                  std::chrono::duration_cast<std::chrono::nanoseconds>(
                          std::chrono::high_resolution_clock::now() - start).count() << std::endl;
    }
}


int main(int argc, char **argv) {
    auto start = std::chrono::high_resolution_clock::now();
    std::vector<std::thread> thd;
    if (argc < 2 || argv[1][0] == '1') {
        for (int i = 0; i < 100; ++i) {
            thd.emplace_back([&]() {
                while (counter < limit) {
                    lock1();
                    operation(start);
                    unlock();
                }
            });
        }
    } else {
        for (int i = 0; i < 100; ++i) {
            thd.emplace_back([=]() {
                while (counter < limit) {
                    lock2();
                    operation(start);
                    unlock();
                }
            });
        }
    }
    for (auto &i: thd) {
        i.join();
    }
    return 0;
}
```
It astonishes me that `lock1` is so much faster than `lock2`:
image

SchrodingerZhu:
So I test the following programs:
```c++
#include <iostream>
#include <atomic>
#include <chrono>
#include <thread>
#include <vector>

std::atomic_flag word = ATOMIC_FLAG_INIT;
constexpr int limit = 10000000;
int counter = 0;

__attribute__((always_inline)) inline void lock1() {
    while (true) {
        if (!word.test_and_set(std::memory_order_acquire)) {
            break;
        }
        while (word.test(std::memory_order_relaxed)) {
            asm volatile("pause\n":: :"memory");
        }
    }
}

__attribute__((always_inline)) inline void lock2() {
    while (word.test_and_set(std::memory_order_acquire)) {
        asm volatile("pause\n":: :"memory");
    }
}

__attribute__((always_inline)) inline void unlock() {
    word.clear(std::memory_order_release);
}

__attribute__((noinline)) void operation(const std::chrono::high_resolution_clock::time_point start) {
    if ((++counter % 1000) == 0) {
        std::cout << counter << ": " <<
                  std::chrono::duration_cast<std::chrono::nanoseconds>(
                          std::chrono::high_resolution_clock::now() - start).count() << std::endl;
    }
}


int main(int argc, char **argv) {
    auto start = std::chrono::high_resolution_clock::now();
    std::vector<std::thread> thd;
    if (argc < 2 || argv[1][0] == '1') {
        for (int i = 0; i < 100; ++i) {
            thd.emplace_back([&]() {
                while (counter < limit) {
                    lock1();
                    operation(start);
                    unlock();
                }
            });
        }
    } else {
        for (int i = 0; i < 100; ++i) {
            thd.emplace_back([&]() {
                while (counter < limit) {
                    lock2();
                    operation(start);
                    unlock();
                }
            });
        }
    }
    for (auto &i: thd) {
        i.join();
    }
    return 0;
}
```
It astonishes me that `lock1` is much faster than `lock2`:

image

src/ds/flaglock.h
```diff
@@ -6,21 +6,144 @@
...
+        }
+      } while (true);
+#else
+      while (lock.flag.test_and_set(std::memory_order_acquire))
+      {
+        lock.assert_not_owned_by_current_thread();
         Aal::pause();
+      }
+#endif
+      lock.set_owner();
```
mjp41, line 140: Could you push the `ifdef` in a little: 
```suggestion
      do
      {
        // test once with acquire ordering at the beginning
        if (!lock.flag.test_and_set(std::memory_order_acquire))
        {
          break;
        }
        // assert_not_owned_by_current_thread is only called when the first
        // acquiring is failed; which means the lock is already held somewhere
        // else.
        lock.assert_not_owned_by_current_thread();
#ifdef __cpp_lib_atomic_flag_test
        // only need to do relaxed ordering in the busy spin loop
        while (lock.flag.test(std::memory_order_relaxed))
        {
          Aal::pause();
        }
#else
        Aal::pause();
#endif
      } while (true);
      lock.set_owner();
```

The results for this are great.  It should never appear in the critical path for `snmalloc`, but still good to do.

mjp41 (commented):
Thanks for doing this.  If @davidchisnall is happy, then I think we can merge this.

src/ds/flaglock.h
```diff
@@ -6,21 +6,144 @@
...
+        // test once with acquire ordering at the beginning
+        if (!lock.flag.test_and_set(std::memory_order_acquire))
+        {
+          break;
+        }
+        // assert_not_owned_by_current_thread is only called when the first
+        // acquiring is failed; which means the lock is already held somewhere
+        // else.
+        lock.assert_not_owned_by_current_thread();
+        // only need to do relaxed ordering in the busy spin loop
```
davidchisnall, line 127: Is this true?  If I understand the spec correctly (and, to be clear, I definitely do not fully understand the C++11 atomics spec), the relaxed read does not establish a happens-before relationship with the release store.  This means that the spin loop is allowed to spin forever (and a compiler is free to transform it into a single load followed by an infinite loop).

SchrodingerZhu: You are definitely write. x86_64 is generating `mov` anyway so I wrongly placed relaxed ordering here. But it is because x86_64's mov itself has a lot of restrictions. I re-checked the standard:

> **memory_order_relaxed** Relaxed operation: there are no synchronization or ordering  constraints imposed on other reads or writes, only this operation's  atomicity is guaranteed

> **memory_order_acquire** A load operation with this memory order performs the acquire operation on the affected memory location: no reads or writes in the current thread can be reordered before this load. All writes in other threads that release the same atomic variable are visible in the current thread.

So yes, the ordering here should still be `acquire` but this optimisation will also work since there will be no `xchg` while spinning.

SchrodingerZhu: A relatex discussion from the following link:
https://rigtorp.se/spinlock/

mjp41: I am not convinced anything in the standard actually makes progress guarantees.  I was fairly convinced you couldn't optimise away a relaxed atomic, but there are lots of re-orderings that can occur.  Anyway, I am happy with the current slightly more pessimistic code.  And this should not be on a performance critical path, so is fine to be less than optimal.

mjp41 (approved):

PR #433, opened 2021-11-28
export netbsd's reallocarr proposal.

devnexen:
acts subtly differently from reallocarray, returns an error code
and first argument as receiver.


src/override/malloc.cc
```diff
@@ -116,6 +116,36 @@ extern "C"
...
 #endif
 
+#if !defined(__NetBSD__)
+  SNMALLOC_EXPORT int
+    SNMALLOC_NAME_MANGLE(reallocarr)(void* ptr, size_t nmemb, size_t size)
+  {
+    bool overflow = false;
+    auto& a = ThreadAlloc::get();
+    size_t sz = bits::umul(size, nmemb, overflow);
+    if (SNMALLOC_LIKELY(sz == 0))
```
mjp41, line 126: I am surprised this is the likely case.

src/override/malloc.cc
```diff
@@ -116,6 +116,36 @@ extern "C"
   }
 #endif
 
+#if !defined(__NetBSD__)
+  SNMALLOC_EXPORT int
+    SNMALLOC_NAME_MANGLE(reallocarr)(void* ptr, size_t nmemb, size_t size)
```
mjp41, line 121: I was sure this signature must be wrong.  Surely, it should be `void**`, but I read the docs and your right.

src/override/malloc.cc
```diff
@@ -116,6 +116,36 @@ extern "C"
...
+    if (SNMALLOC_LIKELY(sz == 0))
+    {
+      return 0;
+    }
+    if (overflow)
+    {
+      return EOVERFLOW;
+    }
+    size_t osz = a.alloc_size(ptr);
+    void* p = a.alloc(osz);
```
mjp41, line 135: What happens if this allocation fails?

src/override/malloc.cc
```diff
@@ -116,6 +116,36 @@ extern "C"
...
+    memcpy(p, ptr, osz);
```
mjp41, line 136: This copy seems dubious?  Does this corrupt the stack in the common usage?
* `p` - is a pointer to an allocation
* `ptr` - is a pointer to a pointer to an allocation.

src/override/malloc.cc
```diff
@@ -116,6 +116,36 @@ extern "C"
...
+    void* np = SNMALLOC_NAME_MANGLE(realloc)(p, sz);
+    if (SNMALLOC_LIKELY(p != nullptr))
+    {
+      sz = bits::min(size, sz);
+      a.dealloc(ptr);
```
mjp41, line 141: This is attempting to free a pointer to pointer to an allocation?

src/override/malloc.cc
```diff
@@ -116,6 +116,36 @@ extern "C"
   }
 #endif
 
+#if !defined(__NetBSD__)
```
mjp41, line 119: So this isn't defined on NetBSD?

mjp41 (changes requested):
I think my understanding of this is pretty bad.  It seems the type of `ptr` is pretty misleading, and I think means this code is incorrect. 

Could you add a test of a few examples to ensure this code is exercised?

src/override/malloc.cc
devnexen: reallocarr cares about this case.

src/override/malloc.cc
devnexen: ok so I got it backward, I thought, based on reallocarray above was not to export for systems having (e.g. freebsd has reallocarray).

mjp41: Yeah, I am not sure why `reallocarray` is not on `freebsd`,... in this file.  @davidchisnall is this intentional?

src/override/malloc.cc
mjp41: By "cares", do you mean this is a common case, or just a supported case.  Adding `SNMALLOC_LIKELY` is making this do the compiler will optimise for `sz==0`, and make code slower if it isn't.

devnexen: sorry I misread you earlier indeed likely is unneeded here.

davidchisnall:
My memory is slowly coming back.  I disabled `reallocarray` for the in-libc build of snmalloc because libc already provides it (as a wrapper around realloc) and we got duplicate symbols if we also exposed it.  I think that the correct thing to do is have an explicit flag for including these, which defaults to enabled.  We can then as `-DSNMALLOC_NO_REALLOCARRAY` in the FreeBSD libc version (which probably has to be used because otherwise `LD_PRELOAD`ing another malloc that doesn't add this function will be quite exciting (with jemalloc in libc it will work fine, with snmalloc in libc it would fail).

src/test/func/malloc/malloc.cc
```diff
@@ -138,6 +138,23 @@ void test_memalign(size_t size, size_t align, int err, bool null)
   check_result(size, align, p, err, null);
 }
 
+#if defined(__NetBSD__)
+void test_reallocarr(void* p, size_t nmemb, size_t size, int err, bool null)
+{
+  printf("reallocarr(%zu, %zu)\n", nmemb, size);
+  errno = SUCCESS;
+  int r = our_reallocarr(p, nmemb, size);
```
mjp41, line 146: This doesn't look like the correct use case.  The Man page gives the following use case:
```C
           int *data = NULL;
           int ret = 0;

           ret = reallocarr(&data, INITSIZE, sizeof(*data));
           if (ret)
               errc(1, ret, "reallocarr failed");

           ret = reallocarr(&data, NEWSIZE, sizeof(*data));
           if (ret)
               errc(1, ret, "reallocarr failed on resize");
```
https://man.netbsd.org/reallocarr.3

src/test/func/malloc/malloc.cc
```diff
@@ -229,6 +246,14 @@ int main(int argc, char** argv)
     test_posix_memalign(0, align + 1, EINVAL, true);
   }
 
+#if defined(__NetBSD__)
+  for (smallsizeclass_t sc = 0; sc < NUM_SMALL_SIZECLASSES; sc++)
+  {
+    const size_t size = sizeclass_to_size(sc);
```
mjp41, line 252: It would be good to have two loops. So the src and dst sizes are different.  If they are the same, you will hit a fast path on every test.

src/test/func/malloc/malloc.cc
devnexen: good catch

src/override/malloc.cc
```diff
@@ -116,6 +116,65 @@ extern "C"
...
+    }
+    if (overflow)
+    {
+      errno = err;
+      return EOVERFLOW;
+    }
+
+    if (ptr == nullptr)
+    {
+      ptr = a.alloc(sz);
```
mjp41, line 140: I think you have the level of indirection here incorrect. 
```suggestion
    if (*ptr == nullptr)
    {
      *ptr = a.alloc(sz);
```

src/override/malloc.cc
```diff
@@ -116,6 +116,65 @@ extern "C"
   }
 #endif
 
+#if !defined(SNMALLOC_NO_REALLOCARR)
+  SNMALLOC_EXPORT int
+    SNMALLOC_NAME_MANGLE(reallocarr)(void* ptr, size_t nmemb, size_t size)
+  {
```
mjp41, line 122:
```suggestion
    SNMALLOC_NAME_MANGLE(reallocarr)(void* ptr_, size_t nmemb, size_t size)
  {
     // Signature has a misleading level of indirection, cast to correct level of indirection.
     void** ptr = reinterpret_cast<void**>(ptr_);
```

src/override/malloc.cc
```diff
@@ -116,6 +116,65 @@ extern "C"
...
+        r = 0;
+      }
+      else
+      {
+        r = ENOMEM;
+      }
+      return r;
+    }
+
+    void* p = a.alloc(sz);
```
mjp41, line 153: Sorry this is still really confusing me.  You are allocating to `sz` objects here.  One directly, and then a second with realloc?  Why do we need two?  It looks like this code is leaking memory as 
```C++
void* p = a.alloc(sz);
```
appears to be neither returned nor deallocated.

src/override/malloc.cc
```diff
@@ -116,6 +116,65 @@ extern "C"
...
+    if (SNMALLOC_LIKELY(p != nullptr))
+    {
+      memcpy(p, &ptr, sizeof(ptr));
```
mjp41, line 156: This `memcpy` does not appear to be sensible. It is copying the contents of `ptr` into the fresh allocation `p`.

src/test/func/malloc/malloc.cc
```diff
@@ -138,6 +138,22 @@ void test_memalign(size_t size, size_t align, int err, bool null)
...
+void test_reallocarr(void* p, size_t nmemb, size_t size, int err, bool null)
+{
+  size_t old_size = 0;
+  if (p != nullptr)
+    old_size = our_malloc_usable_size(p);
+
+  printf("reallocarr(%p(%zu), %zu)\n", p, old_size, nmemb * size);
+  errno = SUCCESS;
+  int r = our_reallocarr(&p, nmemb, size);
+  if (r != SUCCESS && size != 0)
```
mjp41, line 151: Drop `size != 0` I don't think the spec requires `size == 0` to deallocate.  It doesn't for `realloc`, so I assume it doesn't here.

src/test/func/malloc/malloc.cc
```diff
@@ -138,6 +138,22 @@ void test_memalign(size_t size, size_t align, int err, bool null)
   check_result(size, align, p, err, null);
 }
 
+#if !defined(SNMALLOC_NO_REALLOCARR)
+void test_reallocarr(void* p, size_t nmemb, size_t size, int err, bool null)
+{
+  size_t old_size = 0;
```
mjp41, line 144: This is only ever nullptr from the calling context.
```suggestion
void test_reallocarr(size_t nmemb, size_t size, int err, bool null)
{
  void* p;
  size_t old_size = 0;
```
Alternatively, did you mean
```suggestion
void test_reallocarr(void*& p, size_t nmemb, size_t size, int err, bool null)
{
  size_t old_size = 0;
```

src/test/func/malloc/malloc.cc
```diff
@@ -229,6 +245,22 @@ int main(int argc, char** argv)
     test_posix_memalign(0, align + 1, EINVAL, true);
   }
 
+#if !defined(SNMALLOC_NO_REALLOCARR)
```
mjp41, line 248: The tests should cover size==0,

src/test/func/malloc/malloc.cc
```diff
@@ -138,6 +138,22 @@ void test_memalign(size_t size, size_t align, int err, bool null)
...
+void test_reallocarr(void* p, size_t nmemb, size_t size, int err, bool null)
+{
+  size_t old_size = 0;
+  if (p != nullptr)
+    old_size = our_malloc_usable_size(p);
+
+  printf("reallocarr(%p(%zu), %zu)\n", p, old_size, nmemb * size);
+  errno = SUCCESS;
+  int r = our_reallocarr(&p, nmemb, size);
```
mjp41, line 150: I think it would be good to store a bit pattern in the `reallocarr`, so that we can test we have actually copied the contents.  I liked what @schrodingerzhu did in his current PR for this on Rust.

https://github.com/microsoft/snmalloc/pull/426/files#diff-53e72133956e7966c55e6991080b396e2764f36321e757d4e2cee72bb959530bR17-R27

I think this would catch some bugs in your current code.

mjp41 (changes requested):

src/override/malloc.cc
devnexen: the tests did not return 'not empty allocator' exception (realloc on success ?). ptr is to be modified in the last moment too.

src/override/malloc.cc
devnexen: that would simplify things.

src/test/func/malloc/malloc.cc
mjp41: For instance, I just added 
```C++
    size_t* s = (size_t*)our_malloc(16);
    *s = 123456;
    if (SUCCESS == our_reallocarr(&s, 16, 16))
    {
      if (*s != 123456)
      {
        puts("reallocarr: bad value");
        fflush(stdout);
        abort();
      }
    }
    else
    {
      our_free(p);
    }
```
and it prints bad value.  I think the issue in your current code is either use `realloc` and attempt to get the correct error codes out, or don't use `realloc` and expand the copying.

I would probably lift the end of `realloc` and rewrite something like:
```C++
    void* p = a.alloc(size);
    if (SNMALLOC_LIKELY(p == nullptr))
    {
       return ENOMEM;
    }

    sz = bits::min(size, sz);
    // Guard memcpy as GCC is assuming not nullptr for ptr after the memcpy
    // otherwise.
    if (sz != 0)
      memcpy(p, *ptr, sz);
    a.dealloc(*ptr);
    *ptr = p;
    return SUCCESS;
```

src/override/malloc.cc
```diff
@@ -116,6 +116,59 @@ extern "C"
...
+      return r;
+    }
+
+    void* p = a.alloc(size);
+    if (p == nullptr)
+    {
+      return ENOMEM;
+    }
+
+    sz = bits::min(size, sz);
```
mjp41, line 160:
```suggestion
    sz = bits::min(sz, a.alloc_size(*ptr));
```

src/override/malloc.cc
```diff
@@ -116,6 +116,59 @@ extern "C"
...
+        r = 0;
+      }
+      else
+      {
+        r = ENOMEM;
+      }
+      return r;
+    }
+
+    void* p = a.alloc(size);
```
mjp41, line 154:
```suggestion
    void* p = a.alloc(sz);
```

src/test/func/malloc/malloc.cc
```diff
@@ -229,6 +265,25 @@ int main(int argc, char** argv)
...
 
+#if !defined(SNMALLOC_NO_REALLOCARR)
+  void* a = nullptr;
+  test_reallocarr(a, 1, 0, SUCCESS, false);
+
+  for (smallsizeclass_t sc = 0; sc < (MAX_SMALL_SIZECLASS_BITS + 4); sc++)
+  {
+    const size_t size = bits::one_at_bit(sc);
+    void* p = nullptr;
+    test_reallocarr(p, 1, size, SUCCESS, false);
```
mjp41, line 276: I still don't get the purpose of `p` here?  It is set to `nullptr` and only passed by value, so it is never modified.

mjp41:
> should we add `reallocarray` tests too then ?

Ideally, but that was a pre-existing problem, so it can be done separately later, or as part of this PR if you are keen.

src/test/func/malloc/malloc.cc
```diff
@@ -138,6 +138,61 @@ void test_memalign(size_t size, size_t align, int err, bool null)
...
+}
+#endif
+
+#if !defined(SNMALLOC_NO_REALLOCARR)
+void test_reallocarr(size_t nmemb, size_t size, int err, bool null)
+{
+  void* p = nullptr;
+
+  errno = SUCCESS;
+  int r = our_reallocarr(&p, nmemb, size);
```
mjp41, line 164: I think this is only going to test reallocarr on a `nullptr` reference for the first argument.

Perhaps pass the starting size:
```C++
void test_reallocarr(size_t size_old, size_t nmemb_new, size_t size_new, int err, bool null)
{
   void* p = nullptr;
   // Pass size_old == ~0 for nullptr test. 
   if (size != ~0)
     our_malloc(size_old);
   errno = SUCCESS;
   int r = our_reallocarr(&p, nmemb_new, size_new);
   ...
}
```

mjp41 (changes requested):

mjp41 (changes requested):
@devnexen thanks for all the changes, just a couple of minors bits and then I can merge it.

src/override/malloc.cc
```diff
@@ -116,6 +116,61 @@ extern "C"
...
+  {
+    int err = errno, r;
+    auto& a = ThreadAlloc::get();
+    bool overflow = false;
+    size_t sz = bits::umul(size, nmemb, overflow);
+    if (sz == 0)
+    {
+      errno = err;
+      return 0;
+    }
```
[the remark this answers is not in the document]
devnexen: I would need to catch it later with 0 alloc the pointer is null.

src/override/malloc.cc
```diff
@@ -116,6 +116,61 @@ extern "C"
...
+        errno = err;
+        r = 0;
+      }
+      else
+      {
+        r = ENOMEM;
+        errno = r;
+      }
+      return r;
+    }
```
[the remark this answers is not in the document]
devnexen: sure

devnexen:
should we keep windows 2016 matrix then ?

mjp41:
> should we keep windows 2016 matrix then ?

I am guessing that the hanging is a assert failure launching a Dialog box, and then hanging as the is no interaction.  I have had issues with this on other projects.  I have added a timeout in #438, once that lands, we can rebase your changes on top and see what is going wrong. I don't want to disable a platform until I am sure that it is a good idea.

SchrodingerZhu:
it seems that github is deprecating windows 2016.

src/override/malloc.cc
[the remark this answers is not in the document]
mjp41: So currently we only return nullptr if we are out of memory. So we could check with an assert, and make it such that a test would fail, if we changed that behaviour.  However, your code is much less fragile.  So lets keep it.

mjp41:
> it seems that github is deprecating windows 2016. [actions/virtual-environments#4312](https://github.com/actions/virtual-environments/issues/4312)

Thanks, I hadn't heard about this.  I'll disable those in a separate PR.

src/test/func/malloc/malloc.cc
```diff
@@ -229,6 +285,51 @@ int main(int argc, char** argv)
     test_posix_memalign(0, align + 1, EINVAL, true);
   }
 
+  test_reallocarray(nullptr, 1, 0, SUCCESS, false);
+  for (smallsizeclass_t sc = 0; sc < (MAX_SMALL_SIZECLASS_BITS + 4); sc++)
+  {
+    const size_t size = bits::one_at_bit(sc);
+    test_reallocarray(our_malloc(size), 1, size, SUCCESS, false);
```
mjp41, line 292:
```suggestion
    test_reallocarray(our_malloc(size), 1, size, SUCCESS, false);
    test_reallocarray(our_malloc(size), 1, 0, SUCCESS, false);
```

src/test/func/malloc/malloc.cc
```diff
@@ -229,6 +285,51 @@ int main(int argc, char** argv)
...
+    }
+  }
+
+  test_reallocarr((size_t)~0, 1, 0, SUCCESS, false);
+  test_reallocarr((size_t)~0, 1, 16, SUCCESS, false);
+
+  for (smallsizeclass_t sc = 0; sc < (MAX_SMALL_SIZECLASS_BITS + 4); sc++)
+  {
+    const size_t size = bits::one_at_bit(sc);
+    test_reallocarr(size, 1, size, SUCCESS, false);
```
mjp41, line 309:
```suggestion
    test_reallocarr(size, 1, size, SUCCESS, false);
    test_reallocarr(size, 1, 0, SUCCESS, false);
```

mjp41 (approved):

mjp41:
Thanks for doing this.  It still looks like a zero target size is not covered in the tests. But otherwise great.

mjp41:
Something went a bit wrong with the rebase. I had merged the changes from snmalloc2 onto this, and then it did something a bit weird.  I have manually patch the rebase, and once CI says yes. I'll merge this.

Issue #371, opened 2021-08-09
Pagemap 


davidchisnall:
The pagemap is not allocated with the no-core flag, so when a program crashes it tries to generate a huge core file.  This probably requires a small tweak to the PAL concept to fix.

davidchisnall:
Fixed on FreeBSD by #469 and on Linux by #471.  Not a problem on Windows or OpenEnclave because of differences in the underlying systems.  NetBSD / Solaris / OpenBSD / Haiku may still have this problem but happy to close this as a QOI thing for their PALs.  At least NetBSD and OpenBSD don't appear to provide the necessary system calls to fix the issue.

PR #434, opened 2021-11-30
override add malloc_conceal api proposal

devnexen:


davidchisnall:
Hi, I've thought about adding this API.  I think the correct way of doing it is:

 - Extend the PALs to support no-core memory as a class-template parameter.
 - Create a separate allocator instance that allocates slabs using the no-core Pal.
 - Have these APIs use the second allocator instance.

They can share the pagemap with other allocators and they should be returned to the pools.

Libsodium's secure allocator also supports rounding up allocations to page size, allocating them at the end of the page and putting guard pages on either side.  Once the PAL can provide no-core pages, this should also be possible through the large-object paths in the allocator.  The allocator rounds up to the page size so that you can mprotect the memory read-only or no-access later.

mjp41:
This is a really interesting idea.  I am not sure what the granularity of NODUMP should be.  
```
       The Linux implementation requires that the address addr be page-
       aligned, and allows length to be zero.
```
If this is at a Page granularity, then we probably need to allocate whole slabs with NODUMP set.  This is doable, but would involve adding a completely different set of sizeclasses for concealed allocations.

Adding NODUMP to the PAL is something I have been meaning to do, to address #371.

devnexen:
> This is a really interesting idea. I am not sure what the granularity of NODUMP should be.
> 
> ```
>        The Linux implementation requires that the address addr be page-
>        aligned, and allows length to be zero.
> ```
> 
> If this is at a Page granularity, then we probably need to allocate whole slabs with NODUMP set. This is doable, but would involve adding a completely different set of sizeclasses for concealed allocations.

you re perfectly right it has to, I was just assessing if the overall idea would hit home before going deeper.

PR #435, opened 2021-12-01
build fix proposal for GCC 11.x (spotted with 11.2.0).

devnexen:
build error due to access none attribute on the pthread_setspecific's
value argumenti (glibc).

### DIFF
--- a/src/mem/threadalloc.h
+++ b/src/mem/threadalloc.h
@@ -71,7 +71,6 @@ namespace snmalloc
 #    pragma warning(pop)
 #  endif
 #else
-
   /**
    * Holds the thread local state for the allocator.  The state is constant
    * initialised, and has no direct dectructor.  Instead snmalloc will call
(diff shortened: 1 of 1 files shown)


mjp41:
@devnexen, could you post the error/warning message that GCC was raising.  The value in `set_specific` just has to be non-nullptr for it to function correctly.

devnexen:
I understand it seems useless fix at first glance but it is only a modern glibc/gcc case.

`/home/dcarlier/Contribs/snmalloc/src/test/func/two_alloc_types/../../../override/../mem/threadalloc.h:140:24: error: ‘int pthread_setspecific(pthread_key_t, const void*)’ expecting 1 byte in a region of size 0 [-Werror=stringop-overread]
  140 |     pthread_setspecific(p_key.get(), reinterpret_cast<void*>(1))`

devnexen:
as an alternative we can just make a static var just confined with the function since it s never accessed afterwards anyway.

src/mem/threadalloc.h
```diff
@@ -137,7 +138,7 @@ namespace snmalloc
     Singleton<pthread_key_t, &pthread_create> p_key;
     // We need to set a non-null value, so that the destructor is called,
     // we never look at the value.
-    pthread_setspecific(p_key.get(), reinterpret_cast<void*>(1));
+    pthread_setspecific(p_key.get(), reinterpret_cast<void*>(p_teardown_val));
```
mjp41, line 141: Okay, I prefer your suggestion to make it a static of this function. 

I think you might want to change it to:
```suggestion
    pthread_setspecific(p_key.get(), &p_teardown_val);
```
I think your fix has not addressed the compilers concern just masked it by breaking the analysis pass.

devnexen: in that case we would need to pass the key from a more "meaningful" source as function argument. however no use of this tough.

mjp41 (approved):

mjp41:
Thanks for fixing.

PR #440, opened 2021-12-06
Making PAL numa aware when the underlying system supports it proposal.

devnexen:
Focusing on Linux and Windows.
Here letting the system choosing the proper node id and binding the page
 to it.

### DIFF
--- a/src/pal/pal.h
+++ b/src/pal/pal.h
@@ -28,7 +28,7 @@ namespace snmalloc
 #if !defined(OPEN_ENCLAVE) || defined(OPEN_ENCLAVE_SIMULATION)
   using DefaultPal =
 #  if defined(_WIN32)
-    PALWindows;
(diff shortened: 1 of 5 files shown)


mjp41:
Apologies, I broke CI with the branch name change, can you rebase onto the latest `main` to get the checks to run.

davidchisnall:
Do you have some benchmarks showing an improvement from this change?  It doesn't look as if it's pinning threads to cores, so don't we end up with memory associated with a core that the thread isn't necessarily running on.  If a thread is deallocated then the `Alloc` that owns the memory will be reallocated for another thread, potentially on a completely different NUMA domain, and will now be using memory with the wrong location.

devnexen:
I do not know if it s worthy to make snmalloc numa aware yet, more a POC at the moment ; I did not benchmark it really thoroughly still.

PR #361, opened 2021-08-02
DNM: tentative AddressSpace doc for review

nwf:
I found myself wanting notes about where the new design <s>buried the bodies</s>cached free memory, and what exactly `MetaEntry`-s held.  I won't pretend that this is useful as it stands to anyone but me, but perhaps it's a start on something that might one day be.

PR #441, opened 2021-12-13
Post large deallocations to original thread

mjp41:
This change sets all large allocations to be owned by the originating
thread. This means they will be messaged back to the original thread
before they can be reused.

The following reason for making this change:
* This will improve producer/consumer apps involving large allocations.
* It enables the implementation of a more complex chunk allocator that
reassembles chunks.
* It addresses an issue with compartmentalisation where the handling of
large allocations can result in meta-data ownership changing.

### DIFF
--- a/src/mem/metaslab.h
+++ b/src/mem/metaslab.h
@@ -60,6 +60,12 @@ namespace snmalloc
      */
     bool sleeping_ = false;
 
+    /**
+     * Flag to indicate this is actually a large allocation rather than a slab
+     * of small allocations.
+     */
+    bool large_ = false;
+
     uint16_t& needed()
     {
       return needed_;
@@ -82,6 +88,25 @@ namespace snmalloc
       // allocated from. Hence, the bump allocator slab will never be returned
       // for use in another size class.
       set_sleeping(sizeclass, 0);
+
+      large_ = false;
+    }
+
+    /**
+     * Make this a chunk represent a large allocation.
+     *
+     * Set needed so immediately moves to slow path.
+     */
+    void initialise_large()
+    {
+      // We will push to this just to make the fast path clean.
+      free_queue.init();
+
+      // Flag to detect that it is a large alloc on the slow path
+      large_ = true;
+
+      // Jump to slow path on first deallocation.
+      needed() = 1;
     }
 
     /**
@@ -104,6 +129,11 @@ namespace snmalloc
     bool is_sleeping()
     {
       return sleeping();
+    }
+
+    bool is_large()
+    {
+      return large_;
     }
 
     /**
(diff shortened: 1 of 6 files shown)


src/mem/metaslab.h
```diff
@@ -60,6 +60,12 @@ namespace snmalloc
      */
     bool sleeping_ = false;
 
+    /**
+     * Flag to indicate this is actually a large allocation rather than a slab
+     * of small allocations.
+     */
```
nwf, line 66: I am going to say some things that are probably false, but a corrected version should probably end up somewhere in `docs/`.  I feel like I have written this before but I cannot find it.

Address space is in one of five states:
* Not ours, in which the `Pagemap` is entirely zero and the address space is not tracked by any `AddressSpaceCore`.
* Ours and in an `AddressSpaceCore`: the `Pagemap` `Metaslab*` points to a `FreeChunk`.  The `Pagemap` `remote` and `sizeclass` fields are both zero.
* Ours and in a local allocator's stacks as a large allocation: the Pagemap Metaslab* points to a `ChunkRecord`.  The Pagemap size class is a *large* size class and the remote pointer points at the owning allocator.
* Ours and allocated as a large allocation: the `Pagemap` `Metaslab*` points to a `Metaslab` whose `needed` is `1` to force us off the fast path and whose `large_` flag is `true`.  The `Pagemap` size class is a *large* size class and the remote pointer points at the owning allocator.
* Ours and in use as a slab of some flavor: the `Pagemap` `Metaslab*` points to a `Metaslab` whose `large_` flag is `false`.  The `Pagemap` size class is a *small* size class and the remote pointer points at the owning allocator.

Is all that right?  If it is, should we make `FreeChunk` also have a `MetaCommon` prefix like the others?

nwf (commented):
Two inline comments and also... are we `notify_not_using` things after the first page of a large chunk we've posted?

src/mem/metaslab.h
mjp41: Yeah, that is right. #361 is where you started documenting this.

src/mem/sizeclasstable.h
```diff
@@ -377,13 +379,24 @@ namespace snmalloc
...
+      return static_cast<uint32_t>(offset % sizeclass_full_to_size(sc)) == 0;
+  }
+
+  inline static bool is_start_of_object(sizeclass_t sc, address_t addr)
+  {
+    size_t offset = addr & (sizeclass_full_to_slab_size(sc) - 1);
+    return divisible_by_sizeclass(sc, offset);
+  }
+
+  inline static bool is_start_of_object(smallsizeclass_t sc, address_t addr)
```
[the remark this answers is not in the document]
mjp41: Fixed

mjp41:
> are we notify_not_using things after the first page of a large chunk we've posted?

No, I have been delaying the `notify_not_using` with the Pal timer code.  We could change if we think it makes sense. I am concerned that we should let the original allocator decide how much memory it wants to keep around.

nwf (approved):

PR #443, opened 2021-12-14
use ::write and ::fsync in error path

SchrodingerZhu:

### DIFF
--- a/src/pal/pal_posix.h
+++ b/src/pal/pal_posix.h
@@ -151,10 +151,10 @@ namespace snmalloc
       constexpr int SIZE = 1024;
       void* buffer[SIZE];
       auto nptrs = backtrace(buffer, SIZE);
-      fsync(STDOUT_FILENO);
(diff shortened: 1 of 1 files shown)


src/pal/pal_posix.h
```diff
@@ -151,10 +151,10 @@ namespace snmalloc
       constexpr int SIZE = 1024;
       void* buffer[SIZE];
       auto nptrs = backtrace(buffer, SIZE);
-      fflush(stdout);
+      UNUSED(fsync(STDOUT_FILENO));
```
mjp41, line 154: I would sooner this one be inside error after the three calls to write.  Then you have the pattern some output followed by a flush.  Then it is easier to see you have all the flushes.

mjp41 (approved):
LGTM thanks for doing this.  One minor nit.

mjp41 (approved):
LGTM, thanks for fixing

PR #444, opened 2021-12-14
Three fixes found by running TSan.

mjp41:
These are three issues highlighted by TSan.  This PR also enables TSan and UBSan in CI.

### DIFF
--- a/src/mem/remoteallocator.h
+++ b/src/mem/remoteallocator.h
@@ -106,9 +106,12 @@ namespace snmalloc
       invariant();
       freelist::Object::atomic_store_null(last, key);
 
-      // exchange needs to be a release, so nullptr in next is visible.
+      // Exchange needs to be acq_rel.
+      // *  It needs to be a release, so nullptr in next is visible.
+      // *  Needs to be acquire, so linking into the list does not race with
+      //    the other threads nullptr init of the next field.
       freelist::QueuePtr prev =
-        back.exchange(capptr_rewild(last), std::memory_order_release);
+        back.exchange(capptr_rewild(last), std::memory_order_acq_rel);
 
       freelist::Object::atomic_store_next(domesticate_head(prev), first, key);
     }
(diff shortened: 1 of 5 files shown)


src/mem/localalloc.h
```diff
@@ -426,7 +426,7 @@ namespace snmalloc
       // make pass through call aligned_alloc with the alignment snmalloc
       // would guarantee.
       void* result = external_alloc::aligned_alloc(
-        natural_alignment(size | (sizeof(void*) - 1)), round_size(size));
+        natural_alignment(size), round_size(size));
```
nwf: Did you mean to revert this hunk from 966fe9df74f4b337b11c0ce0e7e08cacfe1716ce?

mjp41: Yeah, this was actually incorrect.  I don't know how it passed locally.

src/mem/remoteallocator.h
```diff
@@ -106,9 +106,12 @@ namespace snmalloc
...
       freelist::Object::atomic_store_null(last, key);
 
-      // exchange needs to be a release, so nullptr in next is visible.
+      // Exchange needs to be acq_rel.
+      // *  It needs to be a release, so nullptr in next is visible.
+      // *  Needs to be acquire, so linking into the list does not race with
+      //    the other threads nullptr init of the next field.
       freelist::QueuePtr prev =
-        back.exchange(capptr_rewild(last), std::memory_order_release);
+        back.exchange(capptr_rewild(last), std::memory_order_acq_rel);
```
mjp41, line 115: So my investigation of a TSAN error can be found here:
https://gist.github.com/mjp41/3747d988209ca1d0d18a5f09ef9d67c7

I think `acquire` is much safer here.  I think the coherence order on `back` might induce sufficient order to make this actually correct, but moving to `acquire` for now.  We can revisit this choice at a later point if perf is harmed on weaker platforms than x64.

nwf (approved):
All LGTM.

PR #449, opened 2021-12-21
Optimise guarded memcpy

mjp41:
One optimisations and some perf testing for guarded memcpy.

The optimisation is a to use a slightly different reciprocal division/modulus calculation, so that large allocations can be a degenerate case.

### DIFF
--- a/src/mem/sizeclasstable.h
+++ b/src/mem/sizeclasstable.h
@@ -148,7 +148,7 @@ namespace snmalloc
     // the slab.
     size_t slab_mask;
     // Table of constants for reciprocal division for each sizeclass.
-    size_t mod_mult;
+    size_t div_mult;
     // Table of constants for reciprocal modulus for each sizeclass.
     size_t mod_zero_mult;
   };
@@ -167,6 +167,8 @@ namespace snmalloc
   {
     ModArray<SIZECLASS_REP_SIZE, sizeclass_data_fast> fast_;
     ModArray<SIZECLASS_REP_SIZE, sizeclass_data_slow> slow_;
+
+    size_t DIV_MULT_SHIFT{0};
 
     [[nodiscard]] constexpr sizeclass_data_fast& fast(sizeclass_t index)
     {
@@ -199,8 +201,10 @@ namespace snmalloc
       return slow_[index.raw()];
     }
 
-    constexpr SizeClassTable() : fast_(), slow_()
+    constexpr SizeClassTable() : fast_(), slow_(), DIV_MULT_SHIFT()
     {
+      size_t max_capacity = 0;
+
       for (sizeclass_compress_t sizeclass = 0;
            sizeclass < NUM_SMALL_SIZECLASSES;
            sizeclass++)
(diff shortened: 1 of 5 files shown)


src/backend/pagemap.h
```diff
@@ -40,19 +40,49 @@ namespace snmalloc
...
+     * with zero, all requests will be handled by the default value until
+     * initialisation.
+     */
+    size_t index_mask{0};
+
+    /**
+     * Optimised way to get index when not necessarily initialised.
+     *
+     * The best path had a shift and mask.  By making the mask value a global
+     * we can collapse the init check into the masking.
```
nwf, line 76:
```suggestion
     * While C/C++ conventionally encodes an optional T as a T*,
     * this requires loading the T* and branching on zero.  Instead,
     * because what we're really after is either the default_value
     * or indexing into an array, we initialize our pointer to the array
     * to point to the default_value and swap it out for the real
     * backing store upon initialization.  Initially, the mask value here
     * is zero, so all accesses are crushed to the zeroth element,
     * but later the mask will allow access to the whole backing store.
     * Thus, we avoid a branch in favor of a bitwise and ALU op.
```

It doesn't _look_ like you're combining into an existing mask instruction so much as avoiding a branch, so maybe something like this instead?  (I could be missing something!)

src/mem/sizeclasstable.h
```diff
@@ -328,40 +334,36 @@ namespace snmalloc
       .capacity;
   }
 
-  inline static size_t mod_by_sizeclass(sizeclass_t sc, size_t offset)
+  inline static address_t start_of_object(sizeclass_t sc, address_t addr)
```
nwf, line 337: Oh, that should be convenient for the CHERI+MTE work's need to find granule 0.

nwf (commented):
Some minor nits about commentary.  The idea LGTM.

src/backend/pagemap.h
mjp41: The code originally (in the dissassembly) had a shift and mask to calculate the aligned address.  The change makes that particular mask a global, so we aren't increasing the instruction stream by much.  But I will add commentary similar to what you proposed with a few addtions.

src/override/memcpy.cc
```diff
@@ -193,15 +193,12 @@ namespace
...
-}
 
-extern "C"
-{
   /**
    * Snmalloc checked memcpy.
    */
-  SNMALLOC_EXPORT void*
-    SNMALLOC_NAME_MANGLE(memcpy)(void* dst, const void* src, size_t len)
+  template<bool checked = true>
```
davidchisnall, line 200: This probably wants to be in an anonymous namespace, not the snmalloc one, since it shouldn't be exposed outside of this compilation unit.

src/test/func/memcpy/func-memcpy.cc
```diff
@@ -82,7 +82,7 @@ void check_size(size_t size)
     {
       dst[i] = 0;
     }
-    my_memcpy(dst, src, sz);
+    snmalloc::memcpy(dst, src, sz);
```
davidchisnall, line 85: I don't really like this because it means that we need to do extra work to hide the symbol from the .so.  We can't use `snmalloc::memcpy<false>` in any other compilation unit.

PR #450, opened 2021-12-24
PAL netbsd (temporary until some time 2022) build fix.

devnexen:
std::random_device seems unimplemented on this platform
 thus falling back to the usual /dev/urandom device for
the time being.

### DIFF
--- a/src/pal/pal_netbsd.h
+++ b/src/pal/pal_netbsd.h
@@ -3,6 +3,8 @@
 #ifdef __NetBSD__
 #  include "pal_bsd_aligned.h"
 
+#  include <fcntl.h>
+
 namespace snmalloc
 {
   /**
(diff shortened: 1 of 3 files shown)


nwf:
Hm.  Is there no feature test macro for `std::random_device` we could use to gate this fallback code?

src/pal/pal_netbsd.h
```diff
@@ -26,7 +28,44 @@ namespace snmalloc
...
+
+    /**
+     * Temporary solution while waiting getrandom support for the next release
+     * random_device seems unimplemented in clang for this platform
+     */
+    static uint64_t get_entropy64()
+    {
+      uint64_t result;
+      auto flags = O_RDONLY | O_CLOEXEC;
+      auto fd = open("/dev/urandom", flags, 0);
```
davidchisnall, line 42: This looks like it's the same as the Linux version - can we factor it out into a function that other operating systems that provide a /dev/[u]random can use?

devnexen:
> Hm. Is there no feature test macro for `std::random_device` we could use to gate this fallback code?

the issue is at linkage time and as it s just a temporary fix I just went to the direct route.

src/pal/pal_netbsd.h
devnexen: almost :-) but I favored linux version regardless.

mjp41 (approved):
LGTM

PR #452, opened 2022-01-07
Remove fake_large_remote

davidchisnall:
Since #441 was merged, pagemap entries are no longer ever set to
fake_large_remote.


mjp41 (approved):
LGTM

PR #453, opened 2022-01-10
Improve check_bounds init check.

mjp41:
There was a slightly awkward jump in the guarded memcpy implementation were if it was uninitialised, it then used the address of the default value.  This commit lifts some of the initialisation checking up a few levels, so that it can be detected, and we just succeed the bounds check if snmalloc has not been initialised.

The resulting ASM is:
```asm x86
<memcpy_guarded>:
    mov    rax,QWORD PTR [rip+0xbfa]        # Load Chunk map base
    test   rax,rax                          # Check if chunk map is initialised
    je     DONE                             #  |
    mov    rcx,rdi                          # Get chunk map entry
    shr    rcx,0xa                          #  |
    and    rcx,0xfffffffffffffff0           #  |
    mov    rax,QWORD PTR [rax+rcx*1+0x8]    # Load sizeclass
    and    eax,0x7f                         #  |
    shl    rax,0x5                          #  |
    lea    r8,[sizeclass_meta_data]         #  |
    mov    rcx,QWORD PTR [rax+r8*1]         # Load object size
    mov    r9,QWORD PTR [rax+r8*1+0x8]      # Load slab mask
    and    r9,rdi                           # Offset within slab
    mov    rax,QWORD PTR [rax+r8*1+0x10]    # Load modulus constant
    imul   rax,r9                           # Perform recripocal modulus
    shr    rax,0x36                         #  |
    imul   rax,rcx                          #  |
    sub    rcx,r9                           # Find distance to end of object.
    add    rcx,rax                          #  |
    cmp    rax,rdx                          # Compare to length of memcpy.
    jb     ERROR                            #  |
DONE:
    jmp    <memcpy>
ERROR:
    ud2                                     # Trap
```

### DIFF
--- a/src/backend/backend.h
+++ b/src/backend/backend.h
@@ -335,6 +335,12 @@ namespace snmalloc
         UNUSED(local_state);
         return concretePagemap.get_bounds();
       }
+
+      static bool is_initialised(LocalState* ls)
+      {
+        UNUSED(ls);
+        return concretePagemap.is_initialised();
+      }
     };
 
   private:
--- a/src/backend/pagemap.h
+++ b/src/backend/pagemap.h
@@ -279,6 +279,14 @@ namespace snmalloc
     }
 
     /**
+     * Check if the pagemap has been initialised.
+     */
+    [[nodiscard]] bool is_initialised() const
+    {
+      return body_opt != nullptr;
+    }
+
+    /**
      * Return the starting address corresponding to a given entry within the
      * Pagemap. Also checks that the reference actually points to a valid entry.
      */
--- a/src/mem/localalloc.h
+++ b/src/mem/localalloc.h
@@ -772,6 +772,16 @@ namespace snmalloc
 #endif
     }
 
+    bool check_bounds(const void* p, size_t s)
+    {
+      auto ls = core_alloc->backend_state_ptr();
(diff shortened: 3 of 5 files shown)


davidchisnall:
Looks good to me.  I lost track of why the pagemap couldn't live in BSS anymore - that would avoid the check entirely if it were possible.

davidchisnall (approved):

mjp41:
> Looks good to me. I lost track of why the pagemap couldn't live in BSS anymore - that would avoid the check entirely if it were possible.

So because we have 16 bytes per Chunk (16KiB) on a 48 bit address space that is a 2^38 byte allocation in BSS.  Various linker things seem to explode if there was more than 32bit addresses, I tried some settings but didn't get it working across enough platforms to be happy. 

Just replaying the changes I get
```
relocation truncated to fit R_X86_64_PC32
```
so then I add the
```
-mcmodel=large
```
option, and then it links, but I get a segfault at start up.  There were then compilation issue where things took a very long time, and I think the C++20 semantics of atomic init also messed stuff up.

Ultimately, it was feeling to fragile to work nicely, so I dropped in favour of actually allocating not in BSS.  

Being a mmap call also makes it easier to do the randomisation, and guard pages around the structure.

mjp41:
Also, I left a version compiling and filled 450GB of disk space, and my laptop became very unhappy.

PR #456, opened 2022-02-01
Add jemalloc compat functions.

davidchisnall:
This adds the full set of jemalloc functions that FreeBSD's libc
exposes, including some (the `*allocm` family) that are gone from newer
versions of jemalloc and the `*allocx` family that replaced them.  These
are not necessarily efficient implementations but they should allow
snmalloc to replace jemalloc without any ABI breakage (in the loosest
possible sense).

Jemalloc provides a very generic sysctl-like mechanism for setting and
getting some values.  These are all implemented to return the
not-supported error code.  This may break code that expects that they
will succeed.

In particular, these APIs are used to register custom backing-store
allocators and to manage caches and arenas.  These concepts don't map
directly onto snmalloc and attempting to do so would almost certainly
not provide the same performance characteristics and so it's better to
`LD_PRELOAD` jemalloc (or explicitly link to it) for programs that gain
a significant speedup from this.


davidchisnall:
Note: We are currently missing tests for all of these.  I intend to fix that before it is merged.

src/test/func/jemalloc/jemalloc.cc
```diff
@@ -0,0 +1,393 @@
...
+   */
+#define START_TEST(msg) \
+  function = __PRETTY_FUNCTION__; \
+  fprintf(stderr, "Starting test: " msg "\n");
+
+  /**
+   * An assertion that fires even in debug builds.  Uses the value set by
+   * START_TEST.
+   */
+#define EXPECT(x, msg, ...) \
```
mjp41, line 108: Could we use `SNMALLOC_CHECK`
https://github.com/microsoft/snmalloc/blob/main/src/ds/defines.h#L125

src/override/malloc.cc
```diff
@@ -221,16 +297,267 @@ extern "C"
...
+   * alignment is requested.
+   */
+  int SNMALLOC_NAME_MANGLE(allocm)(
+    void** ptr, size_t* rsize, size_t size, int flags)
+  {
+    auto f = JEMallocFlags(flags);
+    size = f.aligned_size(size);
+    if (rsize != nullptr)
+    {
+      *rsize = size;
```
mjp41, line 347: Looking at the code size could still be below the size that is actually allocated by snmalloc.

snmalloc::aligned_size gives a size that when translated to a size class is guaranteed to meet the alignment, but it isn't necessarily the actual size.

```
snmalloc::aligned_size(16, 66) = 
   (66 -  1) | (16 - 1) + 1 = 80
```
But I think the next size class is 96. 
```suggestion
      *rsize = round_size(size);
```
should fix this.

src/override/malloc.cc
```diff
@@ -221,16 +297,267 @@ extern "C"
...
+    }
+    return (*ptr != nullptr) ? allocm_success : allocm_err_oom;
+  }
+
+  /**
+   * Jemalloc 3 experimental API.  Reallocates the allocation in `*ptr` to be at
+   * least `size` bytes and returns the result in `*ptr`, if `rsize` is not null
+   * then writes the allocated size into `*rsize`.  `flags` controls whether the
+   * memory is zeroed and what alignment is requested and whether reallocation
+   * is permitted.  If reallocating, the size will be at least `size` + `extra`
```
mjp41, line 365: Not sure I understand the use of `extra` here?  Is the intended use, I need it to be `size` big, but if you need to reallocate, make it `extra` bigger.  I think your comment says that, but it took me a couple of passes.

src/override/malloc.cc
```diff
@@ -221,16 +297,267 @@ extern "C"
...
+      // otherwise.
+      if (sz != 0)
+      {
+        memcpy(p, *ptr, sz);
+      }
+      a.dealloc(*ptr);
+      *ptr = p;
+      if (rsize != nullptr)
+      {
+        *rsize = size;
```
mjp41, line 411:
```suggestion
        *rsize = alloc_size;
```

src/override/malloc.cc
```diff
@@ -221,16 +297,267 @@ extern "C"
...
+    {
+      return ptr;
+    }
+
+    if (size == (size_t)-1)
+    {
+      return nullptr;
+    }
+
+    void* p = f.should_zero() ? a.alloc<YesZero>(size) : a.alloc(size);
```
mjp41, line 494: Should we memset the difference?  Might be cheaper.  Though, not really important and keeping simple is good.

src/override/malloc.cc
```diff
@@ -221,16 +297,267 @@ extern "C"
...
+    if (SNMALLOC_LIKELY(p != nullptr))
+    {
+      sz = bits::min(size, sz);
```
mjp41, line 497: Here you are using `size` which has been expanded by `aligned_size`, but not `round_size`.  This seems strange, you probably want the original size.

mjp41 (changes requested):

mjp41:
@davidchisnall I also wonder if this would be better in a 
```
override/jemalloc_compat.cc
```
or
```
override/freebsd_libc_extras.cc
```

Just to make it more separated out.  There is already quite a lot of functionality in `malloc.cc`.

src/override/malloc.cc
davidchisnall: The actual requirement (from the half-sentence 'specification' in the jemalloc manual) for `extra` is that the caller would *like* `size + extra` bytes, but will accept anything above `size` bytes.  Since we're a sizeclass allocator, we always either return the current size if it's `>= size` or allocate to be at least `size+extra` here.

src/override/malloc.cc
```diff
@@ -221,16 +297,267 @@ extern "C"
...
+
+    auto& a = ThreadAlloc::get();
+    size_t sz = a.alloc_size(ptr);
+    // Keep the current allocation if the given size is in the same sizeclass.
+    if (sz == round_size(size))
+    {
+      return ptr;
+    }
+
+    if (size == (size_t)-1)
```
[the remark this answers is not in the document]
davidchisnall: This is copied from realloc.  I'm not actually sure what it is for.

src/test/func/jemalloc/jemalloc.cc
```diff
@@ -0,0 +1,393 @@
...
+      EXPECT(
+        (ret == ALLOCM_SUCCESS),
+        "nallocm(%zx, %d) failed with error %d",
+        size,
+        flags,
+        ret);
+      void* ptr;
+      size_t allocated;
+      ret = Allocm(&ptr, &allocated, expected, flags);
+      EXPECT(
```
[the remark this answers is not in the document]
davidchisnall: `malloc_usable_size` is spelled `sallocm` in this version of the API and is tested below.  This test runs with both our implementation and the one from jemalloc if it is also linked, so needs to have the malloc-implementation-specific things as template parameters.

src/test/func/jemalloc/jemalloc.cc
mjp41: Can you improve SNMALLOC_CHECK?  It is only used in tests. I just would like to not have a proliferation of different bits for doing basically the same thing.

src/override/malloc.cc
```diff
@@ -221,16 +297,266 @@ extern "C"
...
+      // otherwise.
+      if (sz != 0)
+      {
+        memcpy(p, *ptr, sz);
+      }
+      a.dealloc(*ptr);
+      *ptr = p;
+      if (rsize != nullptr)
+      {
+        *rsize = alloc_size;
```
mjp41, line 411: Should this be:
```suggestion
        *rsize = round_size(alloc_size);
```

src/test/func/jemalloc/jemalloc.cc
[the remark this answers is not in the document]
mjp41: I'm surprised the missing `round_size` didn't get caught by this check then?

src/override/malloc.cc
mjp41: Agreed.  Let's leave.  Perhaps leave a comment for future readers, that there is a choice.

src/override/malloc.cc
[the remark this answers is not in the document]
mjp41: Okay, I'll take a look at that code when I get some time.

src/test/func/jemalloc/jemalloc.cc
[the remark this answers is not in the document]
davidchisnall: Not sure.  Possibly the sizes I test are always sufficiently close to a size class that the two are always the same?

src/test/func/jemalloc/jemalloc.cc
davidchisnall: I don't think so easily.  `SNMALLOC_CHECK` calls `Pal::error`, which carefully avoids depending on `malloc`.  The `EXPECT` macro doesn't because it is never used in a situation where we expect `malloc` to be too broken for `fprintf` to work.  If we want to use `SNMALLOC_CHECK` for tests where the allocator under test is the process' `malloc` implementation then we can't make it do this.

Once we can depend on a C++ standard library that has `std::format` and `std::source_location` then we can make much nicer versions of both of these and so I'd prefer to limit the scope of a macro that uses `printf` format strings until we're willing to depend on a C++20 stdlib for the tests (even on platforms where we have a C++20 compiler, we don't always have a C++20 stdlib at the moment).  I could write a C++20 version that used `constexpr` if things to avoid memory allocation if the message didn't use use any format arguments, which would give the right behaviour in both cases, but it would break CI and a bunch of our not-tested-in-CI platforms.

I'm happy to do that in another PR that we don't merge until we're happy that platforms we care about have a sufficiently modern stdlib, if you like.

mjp41: Okay. Makes sense.  Lets perhaps look at lifting something into `tests/utils.h` or something of that nature, so we can share across tests. (As part of a future PR.)

mjp41 (approved):

PR #457, opened 2022-02-07
CHERI bounds bug

rmn30:
Add check in malloc test that CHERI capability length matches `alloc_size` and fix a bug in `address_space_core.h` were not being set correctly.


nwf:
LGTM.  I'll feel better once we get CHERI CI.

mjp41 (approved):
LGTM

PR #458, opened 2022-02-09
Clear freelist pointers on allocation for CHERI or CHECK_CLIENT builds.

rmn30:
This is especially important on CHERI to avoid leaking capabilities to
the freelist. Also add a check in the malloc functional test that
there are no valid capabilities in the return allocation.

There is scope in future to add a dimension to `CapPtr` to track zeroing of allocations but this fixes the obvious big leak for now.

### DIFF
--- a/src/mem/corealloc.h
+++ b/src/mem/corealloc.h
@@ -339,6 +339,17 @@ namespace snmalloc
         address_cast(start_of_slab) ==
         address_cast(chunk_record->meta_common.chunk));
 
+#if defined(__CHERI_PURE_CAPABILITY__) && !defined(SNMALLOC_CHECK_CLIENT)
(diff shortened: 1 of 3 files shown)


mjp41 (approved):
LGTM, just a couple of `clangformat` issues.  If you run the `clangformat` built target it will fix locally, e.g.
```
  ninja clangformat
```

mjp41:
Actually, 
https://github.com/microsoft/snmalloc/blob/79ea08779bb8a6b1ded2b86c1ee9bf38c3e95b19/src/mem/corealloc.h#L304-L331

This code should also be run on Cheri to scrub the pointers.  This might be cheaper to memset the slab you don't need to walk it in order like you do with the CHECK_CLIENT checks.

mjp41 (approved):
LGTM

PR #459, opened 2022-02-10
Series of refactoring for introducing new backend allocator

mjp41:
A series of refactorings to make it easier to replace the current naïve backend allorator with a slightly less naïve buddy allocator.

### DIFF
--- a/src/backend/pagemap.h
+++ b/src/backend/pagemap.h
@@ -236,12 +236,12 @@ namespace snmalloc
      * read/write.
      */
     template<bool potentially_out_of_range>
-    const T& get(address_t p)
+    T& get_mut(address_t p)
     {
       if constexpr (potentially_out_of_range)
       {
         if (SNMALLOC_UNLIKELY(body_opt == nullptr))
-          return default_value;
+          return const_cast<T&>(default_value);
       }
 
       if constexpr (has_bounds)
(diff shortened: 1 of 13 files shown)


src/backend/pagemap.h
```diff
@@ -236,12 +236,12 @@ namespace snmalloc
      * read/write.
      */
     template<bool potentially_out_of_range>
-    const T& get(address_t p)
+    T& get_mut(address_t p)
```
nwf, line 239: In the past, I think there was objection to such an accessor because it couldn't be used across the sandbox boundary.

Now, I'm a little jittery about returning mutable references to the default value.  That seems risky?

src/ds/redblacktree.h
```diff
@@ -8,6 +8,38 @@
...
+      }
+    }
+
+    template<bool new_line = true>
+    static void msg()
+    {
+      if constexpr (TRACE && new_line)
+      {
+#ifdef SNMALLOC_TRACING
+        if (new_line)
```
nwf, line 36: Redundant conditional?

nwf (commented):
Nits, but generally LGTM.

src/backend/pagemap.h
mjp41: So to use the pagemap storage for the buddy allocator, then we need to a way to modify it.  This is currently done in `set_next` in the address space manager, but it gets more complex for the RBTree as it does rotations and a lot more mutations.  Using the pagemap across a boundary like that is not a good idea.

There will be two versions of the buddy allocator, one using in band meta-data and one using the pagemap.  In the compartment version, we could just us in band meta-data.

It is a tricky balance between making the best allocator for X and the best allocator for Y.

src/ds/redblacktree.h
mjp41: Whoops.

PR #460, opened 2022-02-11
Return a small allocation from realloc(ptr, 0).

davidchisnall:
An annoying amount of real-world code (e.g. mandoc, BSD sort) treats a
NULL return from `realloc` as a failure, even when requesting a size of
0.  This code is wrong (the standard explicitly permits a return of NULL
from realloc when given a size 0) but working around it in snmalloc is
easier than fixing it everywhere.

### DIFF
--- a/src/mem/sizeclasstable.h
+++ b/src/mem/sizeclasstable.h
@@ -482,7 +482,7 @@ namespace snmalloc
     }
     if (size == 0)
     {
-      return 0;
+      return 1;
     }
     return sizeclass_to_size(size_to_sizeclass(size));
   }
--- a/src/test/func/malloc/malloc.cc
+++ b/src/test/func/malloc/malloc.cc
@@ -36,12 +36,19 @@ void check_result(size_t size, size_t align, void* p, int err, bool null)
     failed = true;
   }
   const auto alloc_size = our_malloc_usable_size(p);
-  const auto expected_size = round_size(size);
+  auto expected_size = round_size(size);
 #ifdef SNMALLOC_PASS_THROUGH
   // Calling system allocator may allocate a larger block than
   // snmalloc. Note, we have called the system allocator with
   // the size snmalloc would allocate, so it won't be smaller.
   const auto exact_size = false;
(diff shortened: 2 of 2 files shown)


src/mem/sizeclasstable.h
```diff
@@ -482,7 +482,7 @@ namespace snmalloc
     }
     if (size == 0)
     {
-      return 0;
+      return 1;
```
mjp41, line 485: Could you add a comment about returning `nullptr` versus minimum size allocation? And this line would need changing if we changed to `nullptr` return from `realloc`.  It is really useful to document this.

Also, I would prefer 
 ```suggestion
       return sizeclass_to_size(size_to_sizeclass(1));
```
which will compile to 
```
   return 16;
```
But it means we have
```
 malloc_usable_size(malloc(x, size)) == round_size(size)
```

**Review by mjp41 — APPROVED** [review]
You want to change one line of code, rather than fix the world.  I don't know :-)

LGTM.

File: src/test/func/malloc/malloc.cc [hunk 2]
```diff
@@ -36,12 +36,19 @@ void check_result(size_t size, size_t align, void* p, int err, bool null)
...
-  const auto expected_size = round_size(size);
+  auto expected_size = round_size(size);
 #ifdef SNMALLOC_PASS_THROUGH
   // Calling system allocator may allocate a larger block than
   // snmalloc. Note, we have called the system allocator with
   // the size snmalloc would allocate, so it won't be smaller.
   const auto exact_size = false;
+  // We allocate 1 byte for 0-sized allocations (and so round_size will tell us
+  // that the minimum size is 1), but the system allocator may return a 0-sized
+  // allocation.
```
mjp41, line 47:
```suggestion
  // We allocate MIN_ALLOC_SIZE byte for 0-sized allocations (and so round_size will tell us
  // that the minimum size is MIN_ALLOC_SIZE), but the system allocator may return a 0-sized
  // allocation.
```
Probably won't pass clangformat with my change.

PR #462, opened 2022-02-15
shared lib linkage build fix for OpenBSD.

devnexen:
disable purposely --no-undefined for this platform.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -233,7 +233,9 @@ else()
   endif()
 endif()
 
-check_linker_flag(CXX "-Wl,--no-undefined" SNMALLOC_LINKER_SUPPORT_NO_ALLOW_SHLIB_UNDEF)
+if (NOT ${CMAKE_SYSTEM_NAME} MATCHES "OpenBSD")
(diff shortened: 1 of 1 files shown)


devnexen:
ping :)

CMakeLists.txt
```diff
@@ -233,7 +233,9 @@ else()
   endif()
 endif()
 
-check_linker_flag(CXX "-Wl,--no-undefined" SNMALLOC_LINKER_SUPPORT_NO_ALLOW_SHLIB_UNDEF)
+if (NOT ${CMAKE_SYSTEM_NAME} MATCHES "OpenBSD")
+  check_linker_flag(CXX "-Wl,--no-undefined" SNMALLOC_LINKER_SUPPORT_NO_ALLOW_SHLIB_UNDEF)
```
mjp41, line 237: Shouldn't this just failed to find the flag on OpenBSD if it doesn't support the flag?

devnexen: in fact it s supported but disabled on purpose here, it does not have the same behavior than on Linux.

mjp41 (approved):

PR #464, opened 2022-02-24
Refactor error reporting for bounds checks.

davidchisnall:
This introduces a very limited formatter that can embed strings and hex
representations of pointers / integers in an internal buffer. This is
used to format error strings for passing to `Pal::error`.  This is used,
in turn, by a wrapper for reporting bounds checks, which can be used by
external functions to implement bounds checks.

This removes the sprintf_l usage from the bounds checks.

This provides enough of a format implementation that the tests
introduced in #465 can be refactored to use this, instead of their
custom `printf` wrapper and that can be used by SNMALLOC_CHECK.  This
will be a follow-on PR.

### DIFF
--- a/src/ds/helpers.h
+++ b/src/ds/helpers.h
@@ -5,6 +5,7 @@
 
 #include <array>
 #include <atomic>
+#include <string_view>
 #include <type_traits>
 
 namespace snmalloc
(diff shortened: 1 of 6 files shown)


src/ds/helpers.h
```diff
@@ -249,4 +250,144 @@ namespace snmalloc
...
+     */
+    void append(size_t s)
+    {
+      append_char('0');
+      append_char('x');
+      std::array<char, sizeof(size_t) * 2> buf;
+      const char hexdigits[] = "0123456789abcdef";
+      // Length of string including null terminator
+      static_assert(sizeof(hexdigits) == 0x11);
+      for (ssize_t i = buf.size() - 1; i >= 0; i--)
```
mjp41, line 339: I don't think ssize_t is C standard type.

mjp41 (approved):
LGTM

PR #465, opened 2022-02-25
Add test helper macros.

davidchisnall:
- Refactor the existing SNMALLOC_ASSERT and SNMALLOC_CHECK.  These now
   use the FatalErrorBuilder to format the output if a format string is
   provided.
 - Extend the FatalErrorBuilder to print decimal integers for signed
   values.
 - Rewrite the macros used in the jemalloc tests to use
   FatalErrorBuilder and move them into a header.
 - Refactor some of the tests to use the new macros.

### DIFF
--- a/src/ds/defines.h
+++ b/src/ds/defines.h
@@ -1,4 +1,5 @@
 #pragma once
+#include <cstddef>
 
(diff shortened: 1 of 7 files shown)


mjp41 (approved):
LGTM

PR #466, opened 2022-03-02
Improve the `memcpy` implementation.

davidchisnall:
This now either outperforms, or performs as well within the region of
measurement noise as, FreeBSD's libc memcpy, which is hand-written
assembly.

The checked version still incurs a noticeable overhead.

### DIFF
--- a/src/override/memcpy.cc
+++ b/src/override/memcpy.cc
@@ -88,7 +88,7 @@ namespace
       {
         copy_one<Word>(dst, src);
         small_copy<Size - Word, Word>(
-          pointer_offset(dst, Size), pointer_offset(src, Size));
+          pointer_offset(dst, Word), pointer_offset(src, Word));
       }
       else
       {
--- a/src/override/memcpy.cc
+++ b/src/override/memcpy.cc
@@ -77,7 +77,7 @@ namespace
   SNMALLOC_FAST_PATH_INLINE void small_copy(void* dst, const void* src)
   {
     static_assert(
-      bits::next_pow2_const(Word) == Word, "Word size must be a power of two!");
+      bits::is_pow2(Word), "Word size must be a power of two!");
     if constexpr (Size != 0)
     {
       if constexpr (Size >= Word)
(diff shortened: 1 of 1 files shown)


src/override/memcpy.cc
```diff
@@ -79,18 +62,189 @@ namespace
...
+
+  /**
+   * Generate small copies for all sizes up to `Size`, using `WordSize` as the
+   * largest size to copy in a single operation.
+   */
+  template<size_t Size, size_t WordSize = Size>
+  SNMALLOC_FAST_PATH_INLINE void
+  small_copies(void* dst, const void* src, size_t len)
+  {
+    if (len == Size)
```
mjp41, line 109: Could we add an assert here that len <= Size?

src/override/memcpy.cc
```diff
@@ -197,18 +193,21 @@ namespace
...
+    copy(void* dst, const void* src, size_t len)
     {
       // The Intel optimisation manual recommends doing this for sizes >256
       // bytes on modern systems and for all sizes on very modern systems.
       // Testing shows that this is somewhat overly optimistic.
       if (SNMALLOC_UNLIKELY(len >= 512))
       {
         // Align to cache-line boundaries if possible.
-        unaligned_start<64, LargestRegisterSize>(dst, src, len);
+        // unaligned_start<64, LargestRegisterSize>(dst, src, len);
```
mjp41, line 204: Do you want to put this back in?

mjp41 (approved):
LGTM

src/override/memcpy.cc
davidchisnall: We'd end up with 16 copies of that assertion in the fast path for `memcpy`, which would make debug builds really slow.

src/override/memcpy.cc
davidchisnall: Ooops!

LifeIsStrange:
Note that GLIBC provide AVX 256 and AVX 512 accelerated memcpy. I suppose this implementation should only be used when AVX is not supported?

davidchisnall:
You can enable these by setting the register size to 32 and compiling with an AVX-enabled target (I don't have AVX-512 hardware to test on).  I experimented with this but it ended up being slower.  `rep movsb` is faster than AVX for large copies and the cost of the extra branches and getting the alignment right exceeded the benefits relative to SSE for medium-sized copies.  The nice thing about having a C++ implementation is that you can easily experiment with different tweaks.  I'm very happy to review a PR if you can make it faster with AVX!

PR #467, opened 2022-03-02
Add Pal::message.

davidchisnall:
This provides a single place for reporting messages to the user.

While here, be consistent about using stderr for things that should go
to stderr.  We were previously using a mix of stderr and stdout.


mjp41 (approved):
LGTM

PR #468, opened 2022-03-02
Buddy

mjp41:
Missing the following features

- [ ] Lazy Decommit. The features of he `ChunkAllocator` to use time to gradually deallocate memory are not currently implemented.  The `ChunkAllocator` code needs turning into a new backend `Range`.
- [x] Custom OE support.  This needs a slightly different backend configuration
- [x] Blocking merging of blocks from Pal.  Some platforms should not merge blocks from the Pal. For instance, you cannot commit contiguous VirtualAllocs with a single call, so we need to prevent the buddy allocator merging them.  This is also important for CHERI support.
- [x]  CHERI support.  The capabilities for Chunks need to be stored and passed around on CHERI.  That is currently not done.
- [x]  Coarse-grained statistics.  That was previously done by the Chunk Allocator.  Now, we need to add a level to perform that, perhaps add to the GlobalRange to track the current and peak usage. 

Pushing this draft PR, to get some early feedback, and also to get some CI feedback.


SchrodingerZhu:
Not closely related to this PR. But I wonder whether we should do sth like [jemalloc](https://github.com/jemalloc/jemalloc/blob/eabe88916290fec452048eaa1abe1cd52a794339/include/jemalloc/internal/jemalloc_preamble.h.in#L57) for linux PAL to improve its compatibility.

SchrodingerZhu:
> @SchrodingerZhu, what Linux environments support the flag in the kernel but don't provide it in the headers?

Centos 7.9, for example?

src/backend/largebuddyrange.h
```diff
@@ -0,0 +1,287 @@
...
+
+    static constexpr uintptr_t RED_BIT = 2;
+
+    static constexpr Contents null = 0;
+
+    static void set(Holder* ptr, Contents r)
+    {
+      SNMALLOC_ASSERT((r & (MIN_CHUNK_SIZE - 1)) == 0);
+      // Preserve lower bits.
+      *ptr = r | (*ptr & (MIN_CHUNK_SIZE - 1)) | BACKEND_MARKER;
```
nwf, line 33: The CHERI compiler is again a little upset about this expression: `*ptr` is a capability, `*ptr & ...` is a capability, and `r` is a capability, so `r | (*ptr & ...)` again triggers its uncertainty of provenance.
```suggestion
      *ptr = r | statc_cast<address_t>(*ptr & (MIN_CHUNK_SIZE - 1)) | BACKEND_MARKER;
```

src/backend/largebuddyrange.h
```diff
@@ -0,0 +1,287 @@
...
+
+    static bool can_consolidate(Contents k, size_t size)
+    {
+      // Need to know both entries exist in the pagemap.
+      // This must only be called if that has already been
+      // ascertained.
+      // The buddy could be in a part of the pagemap that has
+      // not been registered and thus could segfault on access.
+      auto larger = bits::max(k, buddy(k, size));
+      MetaEntry& entry = Pagemap::template get_metaentry_mut<false>(larger);
```
nwf, line 99:
```suggestion
      MetaEntry& entry = Pagemap::template get_metaentry_mut<false>(static_cast<address_t>(larger));
```

src/backend/smallbuddyrange.h
```diff
@@ -0,0 +1,226 @@
...
+
+    static constexpr Contents null = nullptr;
+
+    static constexpr address_t MASK = 1;
+    static void set(Holder* ptr, Contents r)
+    {
+      SNMALLOC_ASSERT((address_cast(r) & MASK) == 0);
+      if (r == nullptr)
+        *ptr = capptr::Chunk<FreeChunk>(
+          reinterpret_cast<FreeChunk*>(address_cast(*ptr) & MASK));
```
nwf, line 36: The CHERI compiler correctly flags this as the construction of a pointer from a provenance-free integer.
```suggestion
          reinterpret_cast<FreeChunk*>((*ptr).unsafe_uintptr() & MASK));
```

src/mem/metaslab.h
```diff
@@ -227,7 +227,26 @@ namespace snmalloc
...
+    /**
+     * Bit used to indicate this should not be considered part of the previous
+     * PAL allocation.
+     *
+     * Some platforms cannot treat different PalAllocs as a single allocation.
+     * This is true on CHERI as the combined permission might not be
+     * representable.  It is also true on Windows as you cannot Commit across
+     * multiple continuous VirtualAllocs.
+     */
+    static constexpr uintptr_t BOUNDARY_BIT = 1;
```
nwf, line 249:
```suggestion
    static constexpr address_t BOUNDARY_BIT = 1;
```

src/mem/metaslab.h
```diff
@@ -312,7 +323,32 @@ namespace snmalloc
...
-        (alignof(RemoteAllocator) - 1));
+        (REMOTE_WITH_BACKEND_MARKER_ALIGN - 1));
+    }
+
+    MetaEntry(const MetaEntry&) = delete;
+
+    MetaEntry& operator=(const MetaEntry& other)
+    {
+      // Don't overwrite the boundary bit with the other's
+      meta = (other.meta & ~BOUNDARY_BIT) | (meta & BOUNDARY_BIT);
```
nwf, line 334: Another ambiguous provenance case.

```suggestion
      meta = (other.meta & ~BOUNDARY_BIT) | static_cast<address_t>(meta & BOUNDARY_BIT);
```

nwf (changes requested):
Some issues identified by the CHERI compiler.  Beyond the suggestions here, we're tickling a deprecation warning (`CHERI_PERM_CHERIABI_VMMAP` has been renamed to `CHERI_PERM_SW_VMEM`; I'll raise a separate PR) and there appears to be a missing set-bounds in a large alloc path.  Will debug the latter and get back to you, but wanted to flag these for your review.

mjp41:
I am going to leave migrating the slow decommit behaviour to a future PR as this doesn't seem to be an issue with perf currently.

And adding new Range types should be much nicer now.

src/backend/palrange.h
```diff
@@ -0,0 +1,57 @@
...
+  template<SNMALLOC_CONCEPT(ConceptPAL) PAL>
+  class PalRange
+  {
+  public:
+    class State
+    {
+    public:
+      PalRange* operator->()
+      {
+        static PalRange range{};
```
rmn30, line 16: I guess this is some C++ism? Maybe worth a comment to at least name the pattern being used?

nwf (approved):
If CI's happy, I think I'm happy.  Thanks for folding in the CHERI bounds.

PR #469, opened 2022-03-03
[FreeBSD] Exclude unused memory from core dumps.

davidchisnall:
We currently include 256+GiB zeroes in core dumps.  This is a QoI issue
because it causes core dumps to take a long time (even if ZFS is able to
compress large runs of zeros down to almost nothing, they still take a
long time to load into debuggers).  This patch removes anything marked
as not-using or using-but-read-only (pagemap zero pages) as excluded
from core dumps.

Adding __builtin_trap() to the end of `func-malloc-fast` now gives a 21
MiB core dump rather than a 256 GiB one.

### DIFF
--- a/src/pal/pal_bsd_aligned.h
+++ b/src/pal/pal_bsd_aligned.h
@@ -44,7 +44,8 @@ namespace snmalloc
         nullptr,
         size,
         prot,
-        MAP_PRIVATE | MAP_ANONYMOUS | MAP_ALIGNED(log2align) | OS::extra_mmap_flags(state_using),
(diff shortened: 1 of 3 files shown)


mjp41 (approved):
LGTM

davidchisnall:
@SchrodingerZhu, a load of other platforms have some variations of this mechanism, patches to extend this behaviour elsewhere are very welcome!

devnexen:
> @devnexen, a load of other platforms have some variations of this mechanism, patches to extend this behaviour elsewhere are very welcome!

Besides Linux I recall openbsd can do it too.

PR #471, opened 2022-03-03
Follow-up on #469 for Linux this time, overriding read only pages cod…

devnexen:
…e path too.

### DIFF
--- a/src/pal/pal_linux.h
+++ b/src/pal/pal_linux.h
@@ -34,6 +34,14 @@ namespace snmalloc
      */
     static constexpr int default_mmap_flags = MAP_NORESERVE;
 
+    static void* reserve(size_t size) noexcept
+    {
+      void* p = PALPOSIX<PALLinux>::reserve(size);
+      if (p)
+        madvise(p, size, MADV_DONTDUMP);
+      return p;
+    }
+
     /**
      * OS specific function for zeroing memory.
      *
--- a/src/pal/pal_linux.h
+++ b/src/pal/pal_linux.h
@@ -68,20 +68,40 @@ namespace snmalloc
     {
       SNMALLOC_ASSERT(is_aligned_block<page_size>(p, size));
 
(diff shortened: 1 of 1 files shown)


src/pal/pal_linux.h
```diff
@@ -68,20 +68,40 @@ namespace snmalloc
...
+    }
+
+    /**
+     * Notify platform that we will be using these pages.
+     */
+    template<ZeroMem zero_mem>
+    static void notify_using(void* p, size_t size) noexcept
+    {
+      PALPOSIX<PALLinux>::notify_using<zero_mem>(p, size);
+      madvise(p, size, MADV_DONTDUMP);
```
mjp41, line 104: DODUMP?

devnexen: ah yes copy/paste sorry :)

davidchisnall (approved):
Looks good to me, is it marking memory returned by `reserve` as not being included in core dumps?

mjp41:
> Looks good to me, is it marking memory returned by `reserve` as not being included in core dumps?

I think Linux requires a separate call to `madvise`, there is no equivalent of `MAP_NOCORE` that I could see.

devnexen:
> > Looks good to me, is it marking memory returned by `reserve` as not being included in core dumps?
> 
> I think Linux requires a separate call to `madvise`, there is no equivalent of `MAP_NOCORE` that I could see.

thanks good catch I missed that one.

devnexen:
> > Looks good to me, is it marking memory returned by `reserve` as not being included in core dumps?
> 
> I think Linux requires a separate call to `madvise`, there is no equivalent of `MAP_NOCORE` that I could see.

makes me think ... openbsd is literally the other way around ..

devnexen:
Do you know if windows can too ? It seems to me it is not possible looking at virtualloc2* pages

PR #473, opened 2022-03-03
Follow-up on #469 for openbsd but only on pure reservations.

devnexen:


src/pal/pal_openbsd.h
```diff
@@ -22,6 +22,15 @@ namespace snmalloc
...
      */
     static constexpr uint64_t pal_features = PALBSD::pal_features;
+
+    /**
+     * Extra mmap flags.  Exclude mappings from core files if they are
+     * pure reservations.
+     */
+    static int extra_mmap_flags(bool state_using)
+    {
+      return state_using ? 0 : MAP_PRIVATE;
```
davidchisnall, line 32: This doesn't look like the right flag.  `MAP_CONCEAL` is the right thing on OpenBSD.  Without a `madvise` to undo it though, this means that most heap allocations will be excluded from cores, which probably isn't what you want.

devnexen: you re right it s MAP_CONCEAL indeed.

PR #474, opened 2022-03-04
PowerPC64 memcpy arch tuning

nwf:
Here's `perf-memcpy-fast` modified to run with this in place (`unchecked` and `checked`), the two implementations from FreeBSD (`ppc-asm` and `ppc-vsx`), and glibc (Debian package 2.33-7, `platform`).  Hopefully results are representative and I haven't flubbed anything.

20220304-perf-memcpy-fast-mr16-block128-large128

### DIFF
--- a/src/mem/memcpy.h
+++ b/src/mem/memcpy.h
@@ -220,9 +220,35 @@ namespace snmalloc
   };
 #endif
 
+#if defined(__powerpc64__)
+  struct PPC64Arch
+  {
+    /* Modern POWER machines have vector registers */
+    static constexpr size_t LargestRegisterSize = 16;
+
+    /*
+     * For large copies (128 bytes or above), use a copy loop that moves up to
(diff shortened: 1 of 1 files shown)


mjp41:
Looks fast

src/mem/memcpy.h
```diff
@@ -220,9 +220,35 @@ namespace snmalloc
   };
 #endif
 
+#if defined(__powerpc64__)
+  struct PPC64Arch
+  {
+    /* Modern POWER machines have vector registers */
```
davidchisnall, line 226:
```suggestion
    /**
     * Modern POWER machines have vector registers 
     */

File: src/mem/memcpy.h [hunk 2]
```diff
@@ -220,9 +220,35 @@ namespace snmalloc
...
+    static constexpr size_t LargestRegisterSize = 16;
+
+    /*
```
davidchisnall, line 229:
```suggestion
    /**
```

PR #476, opened 2022-03-07
g++ 12 unit test build fix.

devnexen:
test_random_allocation, g++ sees the removal from the list of the test case
as UAF.


mjp41 (approved):
LGTM, thanks

PR #477, opened 2022-03-13
Further disentangle front- and back-ends

nwf:
This is starting to sprawl a bit, but it's all sort of thematically related towards the end goal of removing includes of src/mem from src/backend.

By and large NFCI, except for "clobber MetaEntry-s in dealloc_chunk", which more aggressively zeros out `MetaEntry::meta` pointers and came up while writing docs.

### DIFF
--- a/docs/AddressSpace.md
+++ b/docs/AddressSpace.md
@@ -1,0 +1,157 @@
+# How snmalloc Manages Address Space
+
+Like any modern, high-performance allocator, `snmalloc` contains multiple layers of allocation.
+We give here some notes on the internal orchestration.
+
+## From platform to malloc
+
+Consider a first, "small" allocation (typically less than a platform page); such allocations showcase more of the machinery.
+For simplicity, we assume that
+
+- this is not an `OPEN_ENCLAVE` build,
+- the `BackendAllocator` has not been told to use a `fixed_range`,
+- this is not a `SNMALLOC_CHECK_CLIENT` build, and
+- (as a consequence of the above) `SNMALLOC_META_PROTECTED` is not `#define`-d.
+
+Since this is the first allocation, all the internal caches will be empty, and so we will hit all the slow paths.
+For simplicity, we gloss over much of the "lazy initialization" that would actually be implied by a first allocation.
+
+1. The `LocalAlloc::small_alloc` finds that it cannot satisfy the request because its `LocalCache` lacks a free list for this size class.
+   The request is delegated, unchanged, to `CoreAllocator::small_alloc`.
+
+2. The `CoreAllocator` has no active slab for this sizeclass, so `CoreAllocator::small_alloc_slow` delegates to `BackendAllocator::alloc_chunk`.
+   At this point, the allocation request is enlarged to one or a few chunks (a small counting number multiple of `MIN_CHUNK_SIZE`, which is typically 16KiB); see `sizeclass_to_slab_size`.
+
+3. `BackendAllocator::alloc_chunk` at this point splits the allocation request in two, allocating both the chunk's metadata structure (of size `PAGEMAP_METADATA_STRUCT_SIZE`) and the chunk itself (a multiple of `MIN_CHUNK_SIZE`).
+   Because the two exercise similar bits of machinery, we now track them in parallel in prose despite their sequential nature.
+
+4. The `BackendAllocator` has a chain of "range" types that it uses to manage address space.
+   By default (and in the case we are considering), that chain begins with a per-thread "small buddy allocator range".
+
+   1. For the metadata allocation, the size is (well) below `MIN_CHUNK_SIZE` and so this allocator, which by supposition is empty, attempts to `refill` itself from its parent.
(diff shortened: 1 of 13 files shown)


nwf:
Well rats.  I seem to have broken something, but I'm not sure what and it's nothing on any platform I have both easy access to and understanding of.  Forgive some bisection on the PR.

nwf:
OK, so, so far so good, but something about my understanding is incomplete because https://github.com/nwf/snmalloc/commit/879b4bf9ccc6f35ecb94f7ebaddc271a7ee9fd98 (atop this series) causes some mysterious breakages on some Windows builds and, apparently, in arm on qemu.  I spent a while chasing after a missing `META_BOUNDARY_BIT`, but I don't think it's that, so I'm at a bit of a loss.

docs/AddressSpace.md
```diff
@@ -0,0 +1,153 @@
...
+For chunks owned by a *small buddy allocator*, the remainder of the `MetaEntry` is zero.
+That is, it appears to have small sizeclass 0 and an implausible `RemoteAllocator*`.
+
+For chunks owned by a *large buddy allocator*, the `MetaEntry` is instead a node in a red-black tree of all such chunks.
+Its contents can be decoded as follows:
+
+1. The `meta` field's `META_BOUNDARY_BIT` is preserved, with the same meaning as in the frontend case, above.
+
+2. `meta` (resp. `remote_and_sizeclass`) includes a pointer to the left (resp. right) *chunk* of address space.
+   (The corresponding child *node* in this tree is found by taking the *address* of this chunk and looking up the `MetaEntry` in the Pagemap.)
```
rmn30, line 114: Why use the address of the chunk and not the `MetaEntry`? Wouldn't that make traversing the tree faster?

docs/AddressSpace.md
```diff
@@ -0,0 +1,153 @@
...
+3. The address is part of a free chunk inside a backend's Small Buddy Allocator:
+   Here, the `MetaEntry` is zero aside from the asserted `REMOTE_BACKEND_MARKER` bit, and so it...
+   * has "small" sizeclass 0, which has size 0.
+   * has no associated metadata structure.
+
+4. The address is part of a live large allocation (spanning one or more 16KiB chunks):
+   Here, the `MetaEntry`...
+   * has `REMOTE_BACKEND_MARKER` clear in `remote_and_sizeclass`.
+   * has a *large* sizeclass value.
+   * has an associated `RemoteAllocator*` and metadata structure.
```
rmn30, line 148: What is the type of the metadata structure?

rmn30 (commented):
Very useful documentation, I have not yet looked at the code. Might be useful to say something about the difference between CoreAlloc and LocalAlloc, perhaps in a footnote? I understand it is related to lazy initialization?

docs/AddressSpace.md
mjp41: If you use the address of the chunk, then it actually works better with CHERI.  We can get permission on the pagemap from a global, so we don't need a capability to access it.  But if you wanted direct pointers into the pagemap, either you would have to rederive them to follow, or you would need three capabilities, one for each direction, and one for the underlying allocation.  With pointers like this, we have access to the full capability for the chunks.  This makes the code almost unchanged for CHERI (@nwf spotted this, and saved me jumping through hoops for CHERI).

This representation adds a shift and add to each level of the tree, which given it is balanced and consolidated is not going to be many levels, over the direct pointers.  Also, this is not perf critical as it is many levels off the fast path.

nwf: Thanks for calling that out; I'll add it to the documentation.

It's also probably worth noting that there's a *lot* of redundancy in the pagemap in CHERI at the moment: everyone, either in the Pagemap directly (as here) or in the `MetaCommon::chunk` for slabs is carrying around full `Arena` bounds to the entire underlying `mmap` region so that we can `madvise` and `Pal::zero` with `mmap` and, most challenging, recombine within the buddy allocators.  I'm not exactly thrilled with that, but I don't see an easy way to make anything smaller, since we do need pointers in the pagemap and so may as well use the space when we've got it, and while it might be good to more tightly bound internally (perhaps with yet more points in the `capptr::bounds::Spatial` taxonomy and maybe just in debug builds), it's not clear that there's a threat model where doing that actually distinguishes adversaries, since if an adversary can leak one high-authority pointer it can probably choose which one it gets.

docs/AddressSpace.md
nwf: I think it's currently a `ChunkRecord`, but that type isn't long for the world, and I think the `next` pointer therein is unused, making it equivalent to a naked `MetaCommon`.  If that's right, @mjp41, should I add a commit that does away with the use of `ChunkRecord` for this case?

mjp41: Correct. It is a Metaslab, that always triggers a slow path on deallocation, so we can detect large off the fast path.

docs/AddressSpace.md
rmn30: Makes sense, thanks. I was partly asking to check my understanding. I will keep @nwf's points in mind as I browse code.

PR #481, opened 2022-03-17
unsafe_{to,from}_uintptr<T> and a quick audit of casts

nwf:
After utter silence from the compiler as part of #477 spooked both @mjp41 and me (everything was fine, but it really should have been a warning), here's something that tries to induce the compiler to speak up a little more should similar things happen again.

### DIFF
--- a/src/mem/metaslab.h
+++ b/src/mem/metaslab.h
@@ -272,7 +272,7 @@ namespace snmalloc
      */
     SNMALLOC_FAST_PATH
     MetaEntry(Metaslab* meta, uintptr_t remote_and_sizeclass)
-    : meta(reinterpret_cast<uintptr_t>(meta)),
+    : meta(unsafe_to_uintptr<Metaslab>(meta)),
       remote_and_sizeclass(remote_and_sizeclass)
     {}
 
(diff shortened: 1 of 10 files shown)


mjp41 (approved):
LGTM, thanks for fixing this.

src/mem/metaslab.h
```diff
@@ -272,7 +272,7 @@ namespace snmalloc
      */
     SNMALLOC_FAST_PATH
     MetaEntry(Metaslab* meta, uintptr_t remote_and_sizeclass)
-    : meta(reinterpret_cast<uintptr_t>(meta)),
+    : meta(unsafe_to_uintptr<Metaslab>(meta)),
```
davidchisnall, line 275: Would it be better to use a union of the allowed types here, rather than an implicit union in the form of `uintptr_t`?

mjp41: It needs to be `uintptr_t` as the borrowing bottom bits and allowing nullptr leads to UB if it is real pointer.
For instance, a redblack tree node with a null right child would store 258 (I think), and adding 258 to nullptr is UB.

PR #483, opened 2022-03-18
Fix Debug symbols for Windows CI.

mjp41:
The /DEBUG is a linker flag and not a "compile_option".


mjp41:
This should make the Windows CI report stack traces correctly. This would have helped with https://github.com/microsoft/snmalloc/pull/477#issuecomment-1069738510

nwf (commented):
I know nothing about compiling on Windows.  If you say this is the right thing to do, I'm happy to take your word for it. :)

PR #488, opened 2022-03-23
CheriBSD renamed VMEM SW perm bit

nwf:
CheriBSD 00d71bd4d11af448871d196f987c2ded474f3039 changes
"CHERI_PERM_CHERIABI_VMMAP" to be spelled "CHERI_PERM_SW_VMEM" and deprecated
the old form.  Follow along.


rmn30:
The change looks trivial but I wonder if we could ease the transition a bit with something like:
```
#ifndef CHERI_PERM_SW_VMEM
#define CHERI_PERM_SW_VMEM CHERI_PERM_CHERIABI_VMMAP
#endif
```
Which would allow you to build snmalloc against an old or new kernel.

rmn30:
I suppose if it is deprecated not removed that is probably not necessary.

rmn30:
Actually, it would be useful to have the define for building snmalloc against an old FreeBSD that doesn't define `CHERI_PERM_SW_VMEM`.

rmn30 (approved):

PR #489, opened 2022-03-23
RFC Backend *Range plumbing

nwf:
This sets the stage for bounding metadata allocations more tightly than the whole underlying arena (on CHERI) though it does not actually do so yet.  Specifically, this series serves to get the `SmallBuddyRange` out of the chain taken by ordinary objects and only on the chain (at the head, in fact) for metadata objects; we can therefore have these two chains have different return values.

While it seems attractive to introduce a `Range` type that manages single chunks and caches the arena-bounded pointer in the `PageMap` on `alloc` and substitutes the cached version back on `dealloc`, I think that this is insufficient for `SmallBuddyRange`'s needs once it begins enforcing bounds on allocations, and so I fear that `SmallBuddyRange` will come to directly know about the `PageMap`... but I haven't gotten there yet, so am not sure.

### DIFF
--- a/src/backend/commitrange.h
+++ b/src/backend/commitrange.h
@@ -25,11 +25,13 @@ namespace snmalloc
       }
     };
 
+    using B = typename ParentRange::B;
+
     static constexpr bool Aligned = ParentRange::Aligned;
 
     constexpr CommitRange() = default;
 
-    capptr::Chunk<void> alloc_range(size_t size)
+    CapPtr<void, B> alloc_range(size_t size)
     {
       auto range = parent->alloc_range(size);
       if (range != nullptr)
@@ -38,7 +40,7 @@ namespace snmalloc
     }
 
     template<SNMALLOC_CONCEPT(ConceptBackendRange_Dealloc) _pr = ParentRange>
-    void dealloc_range(capptr::Chunk<void> base, size_t size)
+    void dealloc_range(CapPtr<void, B> base, size_t size)
     {
       static_assert(
         std::is_same<_pr, ParentRange>::value,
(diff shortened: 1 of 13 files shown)


src/backend/indirectrange.h
```diff
@@ -0,0 +1,105 @@
...
+{
+  template<typename T>
+  static T& IndirectRange_default_kparent(T* x)
+  {
+    return *x;
+  }
+
+  template<
+    typename ParentRange,
+    typename KArg_ = typename ParentRange::State*,
```
mjp41, line 15: Could the following be made to work:
```suggestion
    typename KArg_ = typename ParentRange::State&,
```
Then it would ensure that callers had to provide one?

src/backend/commitrange.h
```diff
@@ -23,22 +25,29 @@ namespace snmalloc
...
 
+    using B = typename ParentRange::B;
+    using KArg = typename ParentRange::KArg;
+
     static constexpr bool Aligned = ParentRange::Aligned;
 
     constexpr CommitRange() = default;
 
-    capptr::Chunk<void> alloc_range(size_t size)
+    CapPtr<void, B> alloc_range(KArg ka, size_t size)
```
mjp41, line 35: Making `KArg ka` the last parameter might be a useful trick, so we can add defaults and not change the API.

mjp41 (commented):
This wasn't as bad as I was expecting.  I wonder if `GlobalRange` could partially be replaced by `IndirectRange`.  We would still need something to do locking for shared access, but the referencing a global could be lifted to `IndirectRange`.

It is still unclear to me precisely why we need this.  I see you want to switch the signature so the ranges do the bounding, but could we do a rederive in this case, with something stored in the pagemap for sub-chunk meta-data?

nwf-msr:
Obsolesced by #529.

PR #490, opened 2022-03-23
RFC split backend alloc_chunk to add _premeta

nwf:
As discussed today, this may be more useful for the process sandbox work.

### DIFF
--- a/src/backend/backend.h
+++ b/src/backend/backend.h
@@ -246,6 +246,48 @@ namespace snmalloc
     }
 
     /**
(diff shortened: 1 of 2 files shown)


nwf:
In the interest of further disentangling front and backend, it may make sense to move the current notion of `MetaCommon` out to being a frontend concern and have the backend specify what it believes is analogous to the current `MetaCommon::chunk` field.

mjp41:
I'm not sure I see the usecase for this.  Let's see what is needed for the sandbox use case, and add something like this, if it makes sense.

PR #491, opened 2022-03-23
Remove ChunkAllocator

mjp41:
This is the first commits for adding the DecayRange.  It involves

* removing the redundant ChunkAllocator
* Adding a field representing if a Range is concurrency safe
* Fixing OE Stats
* Improving the message printing.


src/mem/corealloc.h
```diff
@@ -282,7 +281,7 @@ namespace snmalloc
       bumpptr = slab_end;
     }
 
-    ChunkRecord* clear_slab(Metaslab* meta, smallsizeclass_t sizeclass)
+    Metaslab* clear_slab(Metaslab* meta, smallsizeclass_t sizeclass)
```
nwf, line 284: The return here was always just a type pun and can probably be done away with now.

src/pal/pal.h
```diff
@@ -168,10 +168,23 @@ namespace snmalloc
...
+    static thread_local size_t tid{0};
+    static std::atomic<size_t> tid_source{1};
+
+    if (tid == 0)
+    {
+      tid = tid_source++;
+    }
+    return tid;
+  }
+
```
nwf, line 182: Generally approve of the idea, but should this actually be in the PAL so we can actually get the platform notion of a thread ID (when it's sufficiently inexpensive to do so)?

src/backend/palrange.h
```diff
@@ -24,6 +24,8 @@ namespace snmalloc
 
     static constexpr bool Aligned = pal_supports<AlignedAllocation, PAL>;
 
+    static constexpr bool ConcurrencySafe = true;
```
nwf, line 27: While true at the moment, do we want to promise that all PALs have concurrency-safe (de)alloc routines?

nwf (commented):
Some minor nits; generally looks good.

Do you want to add a `static_assert`s in `backend.h` that `GlobalRange` is `ConcurrencySafe`?

src/backend/palrange.h
mjp41: Good question.  I could say false instead?  Or we could add a feature flag to each Pal that says if it is true?

src/pal/pal.h
nwf: This is a strict improvement, so no reason to drop it.  Was mostly asking in case there was reason not to do the PAL thing.

src/mem/corealloc.h
mjp41: Good point. I had removed the use-site, but not the return.

src/pal/pal.h
mjp41: I was trying to trace a bug in `DecayRange` which only exhibited with concurrency, and this was the quickest thing I could do to get a trace I could understand.  I think we can engineer something better.

mjp41: > I think we can engineer something better.

But not tonight ;-)

src/backend/palrange.h
mjp41: The Pals have always been assumed to be concurrency safe so far.

nwf (approved):
LGTM; I'll leave it to you to squash into one or many commits.

PR #492, opened 2022-03-24
Expose a way of constructing MetaslabMetaEntry instances.

davidchisnall:


mjp41 (approved):
LGTM, @nwf any objections?

nwf:
As per additional discussion with @davidchisnall, I think `MetaslabMetaEntry` is going away?

davidchisnall:
The current use of `MetaslabMetaEntry` relies on down-casts that are UB.  The current `MetaCommon` has a private field that every back end must use but can't.

I think the correct fix is probably:

 - Remove `MetaslabMetaEntry` and make the current methods free functions on `MetaEntry`.
 - Make `MetaCommon` a class inside the back end, allowing it to be extended by the back end if the back-end wants to store extra state there.
 - Provide `MetaEntry` as a template parameter to the back end and `static_assert` that `MetaEntry` is a subclass of `Backend::MetaCommon` (which may be empty if a back end doesn't need or want to carry the chunk pointer there).
 - Make `dealloc_chunk` take an address.  If the back end needs to look up the authorising capability in the pagemap, it can because it has a reference to the pagemap anyway.

nwf:
Are we going this way or #493?

PR #493, opened 2022-03-24
Stop playing OO games with MetaEntry

nwf:
David points out that the downcasts I had introduced were UB.  Instead, go back
to passing MetaEntry-s around and make MetaslabMetaEntry just a namespace of
static methods.

This partially reverts 7940fee00cf0ce43c3c40584a3b3e4fb8d77cfd8


nwf:
CI seems lost.  Forgive some violence.

davidchisnall (approved):

PR #494, opened 2022-03-25
Make MetaCommon state public.

davidchisnall:
C++ has no way of exposing it only to things that are backend
implementations.  Eventually this class should be made purely part of
the CHERI back end.


mjp41 (approved):
I like the idea of this not occurring unless we are on Cheri.

PR #495, opened 2022-03-25
Allow POSIX PALs to overwrite writev and fsync.

davidchisnall:
On FreeBSD (possibly elsewhere) the normal versions of these go via an
indirection layer because they are pthread cancellation points.  This
indirection layer does not work correctly if pthreads can't allocate
memory and so we can't get debug output until malloc is working, at
least a little bit.

With this version, we can call the __sys_ variants, which skip any libc
/ libthr interposition.

### DIFF
--- a/src/pal/pal_bsd.h
+++ b/src/pal/pal_bsd.h
@@ -10,8 +10,8 @@ namespace snmalloc
    * Generic *BSD PAL mixin.  This provides features that are common to the BSD
    * family.
    */
-  template<typename OS>
-  class PALBSD : public PALPOSIX<OS>
+  template<typename OS, auto... Args>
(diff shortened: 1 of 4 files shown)


mjp41 (approved):

devnexen:
would it bother you if I commit to your branch ?

mjp41:
> would it bother you if I commit to your branch ?

@devnexen, I am not sure you will have the ability to directly commit to this branch, but you should be able to PR onto the underlying branch @davidchisnall is using. 

We're always super happy for your contributions ;-)

mjp41:
Or you can use the 
```
```suggestions
```
notation, but that can be painful, and only allows you to change bits that already have changes.

devnexen:
ah I did not notice it was not a fork :)

PR #497, opened 2022-03-27
Following up using writev/fsync direct syscalls to avoid pthread inte…

devnexen:
…rpositions

### DIFF
--- a/src/pal/pal_bsd.h
+++ b/src/pal/pal_bsd.h
@@ -10,8 +10,8 @@ namespace snmalloc
    * Generic *BSD PAL mixin.  This provides features that are common to the BSD
    * family.
    */
-  template<typename OS>
-  class PALBSD : public PALPOSIX<OS>
+  template<typename OS, auto... Args>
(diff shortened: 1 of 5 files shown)


davidchisnall:
This looks fine to me.  Please can you add a doc comment explaining why it's necessary for NetBSD?  On FreeBSD, libc contains an interposition layer that allows libthr to replace certain functions that POSIX defines as cancellation points.   `write` goes through this layer.  It's not obvious to me that NetBSD would have the same structure (on FreeBSD, this is legacy of the 5.x series, which supported 1:1 and N:M threading models and so needed an interposition layer that did different things depending on which pthread implementation was used).  This interposition layer requires `malloc` to be working and deadlocks on the first call if it isn't.  Did you check that it's necessary (patch the PAL to return `nullptr` from the reserve functions and see if it prints an error)?

davidchisnall:
Sorry, this was closed automatically when the branch was merged.  Please reopen it against main!

PR #499, opened 2022-03-28
Following up using writev/fsync direct syscalls to avoid pthread inte…

devnexen:
…rpositions

### DIFF
--- a/src/pal/pal_netbsd.h
+++ b/src/pal/pal_netbsd.h
@@ -5,6 +5,16 @@
 
 #  include <fcntl.h>
 
+/**
+ * We skip the pthread cancellation checkpoints by reaching directly
+ * the following syscalls so we avoid the possible pthread
+ * allocation initialization timing issues.
+ * @{
+ */
+extern "C" ssize_t _sys_writev(int fd, const struct iovec* iov, int iovcnt);
+extern "C" int _sys_fsync(int fd);
+/// @}
+
 namespace snmalloc
 {
   /**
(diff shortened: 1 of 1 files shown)


mjp41:
@davidchisnall has the following comment on the earlier PR.  Just posting here so it doesn't get lost:

> This looks fine to me. Please can you add a doc comment explaining why it's necessary for NetBSD? On FreeBSD, libc contains an interposition layer that allows libthr to replace certain functions that POSIX defines as cancellation points. `write` goes through this layer. It's not obvious to me that NetBSD would have the same structure (on FreeBSD, this is legacy of the 5.x series, which supported 1:1 and N:M threading models and so needed an interposition layer that did different things depending on which pthread implementation was used). This interposition layer requires `malloc` to be working and deadlocks on the first call if it isn't. Did you check that it's necessary (patch the PAL to return `nullptr` from the reserve functions and see if it prints an error)?

davidchisnall:
Thanks @devnexen.  I'm happy with this if you can confirm that we don't see error messages on NetBSD with the PAL hacked to fail all allocations, but do with this change (looking at the NetBSD libc sources, I don't see any equivalent of the problematic code on FreeBSD).

devnexen:
No indeed the issue is not the same but NetBSD's pthread lib can provoke issue through writev with thread specific data, so that s a particular case, let s say it.s more an extra precaution rather than an actual fix.

davidchisnall (approved):

PR #500, opened 2022-03-28
Refactor MetaSlab / MetaCommon.

davidchisnall:
MetaCommon is now gone.  The back end must provide a BackendMetadata,
which must be a subtype of MetaSlab (i.e. MetaSlab or a subclass of
MetaSlab).  It may add additional state here.

The MetaEntry is now templated on the concrete subclass of MetaSlab that
the back-end uses.  The MetaEntry still stores this as a `uintptr_t` to
allow easier toggling of the boundary bit but the interfaces are all in
terms of stable types now.

Also some tidying of names (SharedStateHandle is now called Backend).

In a follow-on PR, we can then remove the chunk field from the
BackendMetadata in the non-CHERI back end and allow back ends that don't
require extra state to use MetaSlab directly.


src/backend/backend_concept.h
```diff
@@ -15,20 +17,29 @@ namespace snmalloc
...
+      Meta::set_metaentry(addr, sz, t)
+    }
+    ->ConceptSame<void>;
 
-    { Meta::template get_metaentry<true>(addr) }
-      -> ConceptSame<const MetaEntry&>;
+    {
+      Meta::template get_metaentry<true>(addr)
+    }
+    ->ConceptSame<const MetaEntry<BackendMetadata>&>;
```
ihaller, line 37: This does not return a `MetaEntry<BackendMetadata>&`, but a `PageMapEntry<BackendMetadata>&`.

src/backend/backend_concept.h
```diff
@@ -15,20 +17,29 @@ namespace snmalloc
...
   concept ConceptBackendMeta =
-    requires(
-      address_t addr,
-      size_t sz,
-      const MetaEntry& t)
+    requires(address_t addr, size_t sz, const MetaEntry<BackendMetadata>& t)
   {
-    { Meta::set_metaentry(addr, sz, t) } -> ConceptSame<void>;
+    {
+      Meta::set_metaentry(addr, sz, t)
```
ihaller, line 30: set_metaentry is not used by the frontend anymore, so it could be removed from the concept. Right @mjp41 ?

src/backend/metatypes.h
```diff
@@ -163,9 +117,9 @@ namespace snmalloc
      * assert that this chunk is being used as a slab (i.e., has an associated
      * owning allocator).
      */
-    [[nodiscard]] SNMALLOC_FAST_PATH MetaCommon* get_meta() const
+    [[nodiscard]] SNMALLOC_FAST_PATH BackendMetadata* get_meta() const
     {
-      return reinterpret_cast<MetaCommon*>(meta & ~META_BOUNDARY_BIT);
+      return reinterpret_cast<BackendMetadata*>(meta & ~META_BOUNDARY_BIT);
```
ihaller, line 122: Should this use unsafe_from_uintptr for symmetry with the constructor?

src/backend/metatypes.h
```diff
@@ -120,7 +46,35 @@ namespace snmalloc
...
      */
     static constexpr address_t REMOTE_BACKEND_MARKER = 1 << 7;
+  };
+
+  /**
+   * Entry stored in the pagemap.  See docs/AddressSpace.md for the full
+   * MetaEntry lifecycle.
+   */
+  template<typename BackendMetadata>
+  class MetaEntry : public MetaEntryBase
```
ihaller, line 56: If PageMapEntry should be a derived class from MetaEntry, then it needs some way to update the underlying meta and remote_and_sizeclass fields.

ihaller: Also having the copy-constructor deleted and needing to implement the assignment operator is more awkward for inheritance.

nwf (commented):
Broadly, I think I can summarize this PR as the consequence of some other design decisions:

1. We want the backend to have a policy that distinguishes metadata and data allocations, and we want the backend to manage the association of metadata with data locations, though the backend does not, itself, _use_ that association for its own end in most cases.  The exception, of course, is CHERI, where Arena-bounded pointers must persist somewhere, and we have elected to use the metadata structure (rather than another design like the older `AuthMap`).

  As a consequence, the backend must at least know the _size_ of the frontend metadata structure.

2. But beyond size, we want the interface between the frontend and backend to speak in terms of the _real_ metadata type  (what this PR calls `BackendMetadata<Metaslab>`), which is a function of both front- and back-ends.  In fact, further, we want the backend to pervasively know this type, so that its `MetaEntry`s constructor and `get_meta` accessors speak in terms of it internally (though it is not always true that `get_meta` returns a pointer to such a type -- see the assertion in `FrontendMetaEntry::get_metaslab`).

3. The frontends are templatized on their backends, but the thought of plumbing the name `Metaslab` as a parameter to that template argument instils fear, because it would have to start at `snmalloc::Globals` and work its way down.

While all of those seem sensible, I still can't quite get over my distaste of `BackendAllocator::BackendMetadata` extending a frontend type by name without indirection (i.e., without it being a parameter to the backend template).  But I haven't been able to come up with anything significantly better, either.

The best I had was what prompted #490: a different design in which `MetaCommon` also disappeared, replaced by a backend-defined type for "cookies" passed to the frontend to be managed and passed back on `dealloc`.  While the backend would no longer need to know about frontend metadata (at all), the requirement to manage cookies may unnecessarily complicate frontend code (and perhaps there's risk of handing the wrong cookie back with a chunk, though this PR is also not immune to corrupting the `BackendMetadata` fields).  (On non-CHERI, the cookies could be made empty, though perhaps not zero-size, because C++.)  This alternative design would also require two separate {,meta}data allocation functions on the backend and perhaps some overly-elaborate knot-tying in the frontend (when a metadata chunk must hold its own cookie).  This alternative design does not achieve point 2 above; the interface between front- and back-ends does not speak in terms of frontend metadata structures at all.

src/backend/backend_concept.h
davidchisnall: @nwf, can you help with the concepts goo?

src/backend/metatypes.h
davidchisnall: It should, though I'd rather that we had a generic IntPtrPair class like LLVM.  This code is inherited from the existing code though.

src/backend/metatypes.h
davidchisnall: That is unchanged from the previous version.  I agree that this class isn't the easiest to extend, but since you're currently the only person with a use case for extending it then I'd rather that we had an orthogonal PR to address those concerns, after you've tried to use it so that we have something that definitely works and not something that I hope works.

src/mem/localalloc.h
```diff
@@ -59,20 +59,29 @@ namespace snmalloc
...
 
   private:
-    using CoreAlloc = CoreAllocator<SharedStateHandle>;
+    /**
+     * Define local names for specialised versions of various types that are
+     * specialised for the back-end that we are using.
+     * @{
+     */
+    using CoreAlloc = CoreAllocator<Backend>;
+    using Metaslab = typename Backend::BackendMetadata;
```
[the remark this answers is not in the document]
davidchisnall: Happy to, if you can suggest better names.  It was on my to-do list but I failed to think of better names.

src/backend/backend_concept.h
davidchisnall: I believe this is fixed now.

src/backend/backend_concept.h
davidchisnall: Gone.

src/backend/metatypes.h
davidchisnall: Done.

src/backend/metatypes.h
davidchisnall: The copy assignment operator is used only by the pagemap.  The new version demonstrates a back end that extends this class and makes all constructors private / friends, so you can see exactly what you need to support for copying / construction in an extended version.

src/backend/backend_concept.h
nwf: Yes, though it looks like you let clangformat get its hands on the concept syntax.  I don't really object to its layout of compound `require`ments, even if it's not how I have been writing them, but the asymmetry of `foo&& bar` is unfortunate.

src/backend/backend_concept.h
```diff
@@ -1,34 +1,45 @@
...
-    requires(
-      address_t addr,
-      size_t sz,
-      const MetaEntry& t)
+  concept ConceptBackendMeta = requires(
+    address_t addr, size_t sz, const typename Meta::Entry& t)
   {
-    { Meta::set_metaentry(addr, sz, t) } -> ConceptSame<void>;
+    {
+      Meta::template get_metaentry_mut<true>(addr)
```
ihaller, line 25: Is get_metaentry_mut used by the front-end? My understanding was that these are the pagemap accessors that any backend needs to implement and sandbox implementations don't normally implement get_metaentry_mut right?

src/backend/backend.h
```diff
@@ -37,20 +37,102 @@ namespace snmalloc
...
+         * Copy assignment is used only by the pagemap.
+         */
+        Entry& operator=(const Entry& other)
+        {
+          PageMapEntry<SlabMetadata>::operator=(other);
+          return *this;
+        }
+      };
+
+    private:
```
ihaller, line 119: Could this be changed to protected instead? I promise to keep dependencies on it to a minimum. Also I think that get_metaentry_mut could be moved here.

src/backend/backend_concept.h
davidchisnall: It's used by LargeBuddyRange, so you need it to exist if you use LargeBuddyRange.  I should probably separate it into a separate concept.

src/backend/backend.h
davidchisnall: I assume that you'll implement your own backend, rather than subclassing this one.  If that isn't too much work, then we can remove some template parameters from this one.  I am not subclassing this for either of the backends in the process-sandboxing code.

src/mem/metaslab.h
```diff
@@ -1,20 +1,189 @@
...
+      return *this;
+    }
+
+    /**
+     * Return the Metaslab metadata associated with this chunk, guarded by an
+     * assert that this chunk is being used as a slab (i.e., has an associated
+     * owning allocator).
+     */
+    [[nodiscard]] SNMALLOC_FAST_PATH BackendMetadata* get_metaslab() const
+    {
```
nwf, line 146: The assert mentioned in the documentation here has gone missing.  `meta` only encodes a `BackendMetadata*` if there's an associated `RemoteAllocator*` (which should be equivalent to `MetaEntryBase::REMOTE_BACKEND_MARKER` being clear in `remote_and_sizeclass`).

```suggestion
    {
      SNMALLOC_ASSERT(get_remote(me) != nullptr);
```

src/backend/backend_concept.h
ihaller: Hmm, that is very useful to be aware of, but I am not sure if it can be captured in a concept, since backends are not required to use LargeBuddyRange, it is an internal implementation decision they make.

davidchisnall: Done in the latest version.

src/mem/metaslab.h
davidchisnall: Ooops, this got lost in the merging of `get_metadata` and `get_meta`.

src/backend/backend_concept.h
davidchisnall: Sorry, I forgot clang-format didn't do a good job here.

src/backend/backend_concept.h
davidchisnall: I've pulled this out into a separate concept which LargeBuddyRange uses but nothing else.

src/backend/backend.h
ihaller: To clarify a bit, set_metaentry is ok to stay private as long as get_metaentry_mut is at least protected (it is public right now, which is good). It would help if concretePagemap was protected since I need access to the get_address method and a pointer to it to pass into a compartment.

ihaller: Without subclassing I would need to copy alloc_chunk, dealloc_chunk and alloc_meta_data as they are and continue copying it over as changes happen to those methods.

davidchisnall:
Uh, what?  I didn't press a merge button here!

src/backend/backend.h
davidchisnall: If you want to subclass, then I'm inclined to remove the definition of `Entry` here.  It was intended as an example for you.  You're going to need to override `alloc_chunk` and `dealloc_chunk` if you want other things in `Entry` anyway, so the total shared code that you'll have with the superclass is a small amount of boilerplate.  I strongly suspect that this is goign to end up with more complexity.

davidchisnall:
Oh, ooops.  Branch protection is not on for `main` branch.

src/backend/metatypes.h
davidchisnall: If you want to subclass, then I should remove the `Pagemap::Entry`.  It's there as an example of what to do if you want a custom metadata entry, but it's not intended to be extensible.

PR #501, opened 2022-03-29
Refactor MetaSlab / MetaCommon.

davidchisnall:
MetaCommon is now gone.  The back end must provide a SlabMetadata,
which must be a subtype of MetaSlab (i.e. MetaSlab or a subclass of
MetaSlab).  It may add additional state here.

The MetaEntry is now templated on the concrete subclass of MetaSlab that
the back-end uses.  The MetaEntry still stores this as a `uintptr_t` to
allow easier toggling of the boundary bit but the interfaces are all in
terms of stable types now.

Also some tidying of names (SharedStateHandle is now called Backend).

In a follow-on PR, we can then remove the chunk field from the
BackendMetadata in the non-CHERI back end and allow back ends that don't
require extra state to use MetaSlab directly.

Other cleanups:

 - Remove backend/metatypes, define the types that the front end expects
   in mem/metaslab.  The back end may extend them but these types define
   part of the contract between the front and back ends.
 - Remove FrontendMetaEntry and fold its methods into MetaEntry.
 - For example purposes, the default back end now extends MetaEntry.
   This also ensures that nothing in the front end depends on the
   specific type of MetaEntry.
 - Some things now have more sensible names.


src/backend/backend.h
```diff
@@ -37,20 +37,102 @@ namespace snmalloc
   template<
     SNMALLOC_CONCEPT(ConceptPAL) PAL,
     bool fixed_range,
-    typename PageMapEntry = MetaEntry>
+    template<typename> class PageMapEntry = MetaEntry>
```
ihaller, line 40: Maybe this is the right time to remove this templating. As discussed, subclassing is not the right path here and it is the only use-case for this template.

src/mem/metaslab.h
```diff
@@ -1,20 +1,191 @@
...
+
+  /**
+   * Entry stored in the pagemap.  See docs/AddressSpace.md for the full
+   * MetaEntry lifecycle.
+   */
+  template<typename BackendMetadata>
+  class MetaEntry : public MetaEntryBase
+  {
+    template<SNMALLOC_CONCEPT(ConceptBuddyRangeMeta) Pagemap>
+    friend class BuddyChunkRep;
```
nwf, line 74: Isn't the need for the _frontend_ here to `friend` a _backend_ class closing the door on other backends providing new range types that use the `MetaEntry` in ways other than `BuddyChunkRep` (the large buddy range)?

Perhaps #493 was premature; it would have sufficed, I think, to ensure that `MetaEntry` and its children classes were standard layout and that the children added no new fields, making them pointer-interconvertible, and so removing any UB around casts between them.

src/mem/metaslab.h
```diff
@@ -163,14 +176,30 @@ namespace snmalloc
...
+    /**
+     * Return the sizeclass.
+     *
+     * This is called by `external_pointer` on arbitrary memory and so cannot
+     * assert that the meta entry is owned by the front end.  In the future, it
+     * may provide some stronger guarantees on the value that is returned in
+     * this case.
+     */
     [[nodiscard]] SNMALLOC_FAST_PATH sizeclass_t get_sizeclass() const
     {
```
nwf, line 204:
```suggestion
    {
      SNMALLOC_ASSERT(!is_backend_owned());
```

src/mem/metaslab.h
```diff
@@ -89,20 +74,51 @@ namespace snmalloc
...
+
+    /**
+     * Default constructor, zero initialises.
+     */
+    constexpr MetaEntryBase() : MetaEntryBase(0, 0) {}
+
   public:
-    constexpr MetaEntry() = default;
+    /**
+     * When a meta entry is in use by the back end, it exposes tow words of
```
nwf, line 92:
```suggestion
     * When a meta entry is in use by the back end, it exposes two words of
```

src/mem/metaslab.h
```diff
@@ -179,6 +208,189 @@ namespace snmalloc
...
+        return address_cast(val);
+      }
+    };
+
+    /**
+     * Get a proxy that allows the back end to read from and write to (some bits
+     * of) a word in the meta entry.  The meta entry must either be unowned or
+     * explicitly claimed by the back end before calling this.
+     */
+    BackendStateWord get_backend_word(Word w)
```
nwf, line 328: Is it worth moving this to `template`-d on `w`, since the only callers know statically which they use?

nwf (commented):
Some initial comments.

src/mem/metaslab.h
davidchisnall: See the comment on this method for an explanation of why this method does not contain that assert.

src/mem/metaslab.h
davidchisnall: I made it dynamic because it's data-dependent in the caller currently and it could be simplified to a single call.  If it's inlined (which it should be, since it's tiny), the caller can then inline it once and the compiler will find it easier to generate better code than if it has to do ICF.

src/mem/metaslab.h
```diff
@@ -89,20 +74,51 @@ namespace snmalloc
...
+        REMOTE_BACKEND_MARKER;
+    }
+
+    /**
+     * Returns true if this metaentry has not been claimed by the front or back
+     * ends.
+     */
+    bool is_unowned() const
+    {
+      return (meta == 0) && (remote_and_sizeclass == 0);
```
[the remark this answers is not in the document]
davidchisnall: That would have hidden the bug that CI found.  Setting the boundary bit is explicitly claiming the meta entry for either the front or back end.

src/mem/metaslab.h
nwf: Sorry, missed that.  Can we assert, then, that it's in one of three non-overlapping states: unowned, owned by backend and sizeclass zero, or owned by frontend?  Or do we not care if we generate nonsensical results from `external_pointer` for things that are backend owned if we allow the backend to reuse the sizeclass bits?

src/test/func/redblack/redblack.cc
```diff
@@ -41,37 +67,38 @@ class Rep
...
   {
-    ptr->value = ((r + Wrapper::offset) << 1) + (ptr->value & 1);
+    ptr.set(r);
   }
 
-  static Contents get(Holder* ptr)
+  static Contents get(Holder ptr)
   {
-    return (ptr->value >> 1) - Wrapper::offset;
+    return (Contents)ptr;
```
mjp41, line 82: Best not to use C-style casts.  Though I don't think Clang-tidy checks the tests, so they don't all meet this bar.

mjp41 (commented):
This are my comments just on the refactor of the red-black tree and the protection of the `BackendMetaentryWord`

.github/workflows/main.yml
```diff
@@ -58,8 +58,8 @@ jobs:
           - os: "ubuntu-latest"
             build-type: Debug
             self-host: true
-            extra-cmake-flags: "-DSNMALLOC_MEMCPY_BOUNDS=ON -DSNMALLOC_CHECK_LOADS=ON"
-            dependencies: "sudo apt install ninja-build"
+            extra-cmake-flags: "-DSNMALLOC_MEMCPY_BOUNDS=ON -DSNMALLOC_CHECK_LOADS=ON -DCMAKE_CXX_COMPILER=g++-10"
```
mjp41, line 61: When you change us to using GCC it was so that people using the defaults on Ubuntu would get a well tested implementation.  I'm little concerned this will just mean we have many issues raised in the future.

src/mem/metaslab.h
davidchisnall: That assert would check that one of the following holds:

 - All bits are zero.
 - All bits except bit N are in an undefined state, bit N is 1.
 - All bits except bit N are in an undefined state, bit N is 0.

For any bit pattern, exactly one out of conditions 2 or 3 will always hold.

src/mem/metaslab.h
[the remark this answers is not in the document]
davidchisnall: Yes, probably.  That could be asserted.

.github/workflows/main.yml
davidchisnall: We can add something to CMake to warn that gcc 9 miscompiles snmalloc.  My experience with Ubuntu 18.04 is that it ships a very buggy compiler.  I've wasted time elsewhere with code that it miscompiles.  I'm happy if it can be made to work but it seems to just generate invalid binaries and I don't see how to fix that.  GCC 10 doesn't have this problem and is in the apt repo.

After some more changes, gcc 9.4 seems to work again.  So, uh, yay?  Reverted.

src/ds/redblacktree.h
```diff
@@ -11,16 +11,40 @@
 namespace snmalloc
 {
 #ifdef __cpp_concepts
+  /**
+   * The representation must define two types.  `Contents` defines some
+   * identifier that can be mapped to a node as a value type.  `Holder` defines
```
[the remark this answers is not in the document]
davidchisnall: With `Holder*` and `Holder&`, you required the holders to be things that were allocated and had a lifetime that lasted beyond the lifetime of the pointers / references to them, which didn't allow them to be wrapper types.  With the new abstraction, `Holder` (which, I agree, is a bad name.  I prefer `Handle` because `Reference` implies that I can't set the thing that it refers to), the assumption that the storage location and the contents types are the same is in precisely one place, and that's easy to remove (done in the latest revision).

src/ds/redblacktree.h
```diff
@@ -77,26 +109,35 @@ namespace snmalloc
...
+      ChildRef& operator=(const K t)
       {
         // Use representations assigment, so we update the correct bits
         // color and other things way also be stored in the Holder.
         Rep::set(ptr, t);
         return *this;
       }
 
+      /**
+       * Comparison operators.  Note that these are address-comparison, not
```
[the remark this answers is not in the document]
davidchisnall: It compares that two Hs have aliased storage (i.e. writing to one will be visible to reads through the other), not that they refer to the same value.  This confused me a lot because I initially assumed comparing two holders would compare the values that they pointed to.

src/ds/redblacktree.h
```diff
@@ -106,20 +147,26 @@ namespace snmalloc
...
+      auto printable()
       {
-        return Rep::get(ptr) == Rep::null;
+        return Rep::printable(ptr);
       }
     };
 
     // Root field of the tree
-    H root{};
+    K root_val{Rep::root};
```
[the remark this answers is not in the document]
davidchisnall: Good point.  Fixed.

src/backend/backend.h
```diff
@@ -363,10 +347,11 @@ namespace snmalloc
     }
 
     static void dealloc_chunk(
-      LocalState& local_state, SlabMetadata& meta_common, size_t size)
+      LocalState& local_state,
+      SlabMetadata& meta_common,
```
nwf, line 351:
```suggestion
      SlabMetadata& meta,
```

src/backend/backend.h
```diff
@@ -363,10 +347,11 @@ namespace snmalloc
...
+      capptr::Alloc<void> alloc,
```
nwf, line 352: Now that these are separated, should we assert that the `MetaEntry` for this `alloc` is indeed pointed at this `meta`-data field?

src/ds/redblacktree.h
```diff
@@ -106,20 +147,26 @@ namespace snmalloc
...
+    H root{&root_val};
```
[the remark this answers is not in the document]
davidchisnall: It needs to be `{H{&root_val}}`, but yes, you're right.  I was confused by the equality comparison over references and assumed that the root node reference needed to be stable, rather than the storage for the root node.  Fixed.

src/ds/redblacktree.h
```diff
@@ -77,26 +109,35 @@ namespace snmalloc
...
         // Use representations assigment, so we update the correct bits
         // color and other things way also be stored in the Holder.
         Rep::set(ptr, t);
         return *this;
       }
 
+      /**
+       * Comparison operators.  Note that these are address-comparison, not
+       * comparison of the values held in these child references.
+       * @{
```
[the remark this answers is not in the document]
davidchisnall: It's doxygen syntax for associating a comment with a block of things.  Eventually, I'd like us to have a CI check that doxygen doesn't report any undocumented methods.  I am at least trying to not make the situation worse...

src/ds/redblacktree.h
```diff
@@ -194,6 +241,16 @@ namespace snmalloc
     {
       ChildRef node;
       bool dir = false;
+      void set(typename Rep::Holder r, bool direction)
+      {
+        node = ChildRef(r);
+        dir = direction;
+      }
+      void set(ChildRef r, bool direction)
```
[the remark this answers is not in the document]
davidchisnall: Also missing doc comments.  Fixed.

src/ds/redblacktree.h
```diff
@@ -207,9 +264,9 @@ namespace snmalloc
       std::array<RBStep, 128> path;
       size_t length = 0;
 
-      RBPath(typename Rep::Holder& root) : path{}
+      RBPath(typename Rep::Holder root) : path{}
       {
-        path[0] = {root, false};
+        path[0].set(root, false);
```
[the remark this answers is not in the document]
davidchisnall: Assignment from an initialiser list is complicated.  It ends up being copy assignment from a newly constructed thing.  This has a lot of implicit steps that were not working reliably.  Set removes all of the implicit things.

src/test/func/redblack/redblack.cc
davidchisnall: We currently run clang-tidy only on the includes.  We probably should extend it to the tests.

src/test/func/redblack/redblack.cc
```diff
@@ -21,16 +21,42 @@ struct Wrapper
   // the representation.
   static constexpr size_t offset = 10000;
 
-  size_t value = offset << 1;
+  size_t* ptr;
+  constexpr Wrapper(size_t* p) : ptr(p) {}
```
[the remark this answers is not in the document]
davidchisnall: Changed to `NodeRef`.  I've also extended the test so that the internal storage is `size_t` but the interface is in terms of `uint16_t`, so that we test both using a wrapper class for accessing and having a different internal representation and interface type.

src/backend/backend.h
davidchisnall: Good idea.  Done.

src/mem/metaslab.h
```diff
@@ -1,20 +1,408 @@
...
+   */
+  static constexpr size_t REMOTE_MIN_ALIGN =
+    bits::max<size_t>(CACHELINE_SIZE, SIZECLASS_REP_SIZE) << 1;
+
+  /**
+   * Base class for the templated MetaEntry.  This exists to avoid needing a
+   * template parameter to access constants that are independent of the
+   * template parameter.
+   */
+  class MetaEntryBase
```
nwf, line 26: Given that `metaslab` is not a great name for what's going on here and nothing about `MetaEntry` directly references `MetaSlab` (because it's indirected through its template for the backend), could this and its child be moved out to its own `src/mem/metaentry.h`?

src/mem/metaslab.h
```diff
@@ -1,20 +1,408 @@
...
+   * all the size classes, both large and small. An additional bit is required
+   * to separate backend uses.
+   */
+  static constexpr size_t REMOTE_MIN_ALIGN =
+    bits::max<size_t>(CACHELINE_SIZE, SIZECLASS_REP_SIZE) << 1;
+
+  /**
+   * Base class for the templated MetaEntry.  This exists to avoid needing a
+   * template parameter to access constants that are independent of the
+   * template parameter.
```
nwf, line 24: It now also includes the fields themselves, too (sort of like the short window from #477 up through #493, the very important difference being that the pagemap now contains the derived class -- or, one assumes, a descendant -- rather than the parent, so no concerns about UB and downcasting to the wrong type arise.  Yes?)

src/ds/seqset.h
```diff
@@ -20,25 +20,26 @@ namespace snmalloc
   template<typename T, bool Fifo = false>
   class SeqSet
   {
+    using NextPtr = decltype(std::declval<T>().next);
```
nwf, line 23: That's cute.

src/mem/metaslab.h
```diff
@@ -89,20 +74,51 @@ namespace snmalloc
...
+    /**
+     * When a meta entry is in use by the back end, it exposes tow words of
+     * state.  The low bits in both are reserved.  Bits in this bitmask must
+     * not be set by the back end in either word.
+     *
+     * During a major release, this constraint may be weakened, allowing the
+     * back end to set more bits.
+     */
+    static constexpr address_t BackendReservedMask =
+      (REMOTE_BACKEND_MARKER + 1);
```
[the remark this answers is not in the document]
nwf: It might also be worth a comment that this mask applies to both words, and so (at present) is a strict superset of actually reserved bits in each word, with the need for two reserved bits arising from the desire to make sizeclass extraction shift-less (otherwise we could put the `REMOTE_BACKEND_MARKER` at the same position as `META_BOUNDARY_BIT`)?

src/mem/metaslab.h
davidchisnall: I have rearranged this so that:

 - `MetaEntry` is now `FrontendMetaEntry`.
 - `Metaslab` is now `FrontendSlabMetadata`.
 - `FrontendMetaEntry` is declared after `FrontendSlabMetadata` so that it can check that its template argument is a subclass of `FrontendSlabMetadata`.
 - `mem/metaslab.h` is now called `mem/metadata.h` and includes all three classes that describe metadata for the front end.

src/mem/metaslab.h
davidchisnall: I've updated the comment.

src/ds/seqset.h
```diff
@@ -20,25 +20,26 @@ namespace snmalloc
...
     {
-      T* head{nullptr};
-      T** end{&head};
+      NextPtr head{nullptr};
+      NextPtr* end{&head};
     };
 
     static_assert(
-      std::is_same<decltype(T::next), T*>::value,
+      std::is_convertible_v<T*, decltype(T::next)>,
```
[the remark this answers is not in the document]
davidchisnall: Done, with an extended comment and an extra `static_assert` because we actually need the conversion to hold in both directions.

src/mem/metaslab.h
```diff
@@ -89,20 +74,51 @@ namespace snmalloc
...
-    constexpr MetaEntry() = default;
+    /**
+     * When a meta entry is in use by the back end, it exposes tow words of
+     * state.  The low bits in both are reserved.  Bits in this bitmask must
+     * not be set by the back end in either word.
+     *
+     * During a major release, this constraint may be weakened, allowing the
+     * back end to set more bits.
+     */
+    static constexpr address_t BackendReservedMask =
```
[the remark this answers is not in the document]
davidchisnall: We should probably have a style guide if we want to have rules like that because it's not a convention I've seen anywhere else.  The more common convention is that *macros* are all-caps, and I find it confusing in snmalloc that we have macros and non-macro constants all uppercase.  I've made the change but it makes me sad.

src/mem/metaslab.h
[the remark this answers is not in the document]
davidchisnall: Comment added.  The +1 is from my forgetting how maths works.  Fixed.

src/mem/metaslab.h
```diff
@@ -163,14 +176,30 @@ namespace snmalloc
...
+      SNMALLOC_ASSERT(!is_backend_owned());
       return reinterpret_cast<RemoteAllocator*>(
         pointer_align_down<REMOTE_WITH_BACKEND_MARKER_ALIGN>(
           get_remote_and_sizeclass()));
     }
 
+    /**
+     * Return the sizeclass.
+     *
+     * This is called by `external_pointer` on arbitrary memory and so cannot
```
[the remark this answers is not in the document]
davidchisnall: The assert or lack thereof is something that I want to know when reading the generated docs and is part of the interface contract (this is safe to call, without prior knowledge that the entry is owned by the front end).  The fact that asserts are present is in the doc comments of other functions.

src/mem/metaslab.h
```diff
@@ -179,6 +208,189 @@ namespace snmalloc
...
+      BackendStateWord& operator=(uintptr_t v)
+      {
+        SNMALLOC_ASSERT_MSG(
+          ((v & BackendReservedMask) == 0),
+          "The back end is not permitted to use the low bits in the meta "
+          "entry. ({} & {}) == {}.",
+          v,
+          BackendReservedMask,
+          (v & BackendReservedMask));
+        *val = v | (*val & BackendReservedMask);
```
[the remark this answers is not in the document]
davidchisnall: Fixed.

src/backend/largebuddyrange.h
```diff
@@ -23,68 +24,86 @@ namespace snmalloc
      * of) chunks of the address space; as such, bits in (MIN_CHUNK_SIZE - 1)
      * are unused and so the RED_BIT is packed therein.  However, in practice,
      * these are not "just any" uintptr_t-s, but specifically the uintptr_t-s
-     * inside the Pagemap's MetaEntry structures.  As such, there are some
-     * additional bit-swizzling concerns; see set() and get() below.
+     * inside the Pagemap's MetaEntry structures.
+     *
+     * The MetaEntry provides us with helpers that guarantee that we use only
```
nwf, line 29:
```suggestion
     * inside the Pagemap's BackendAllocator::Entry structures.
     *
     * The BackendAllocator::Entry provides us with helpers that guarantee that we use only
```

nwf (approved):
I think I'm happy for this PR to land modulo one nit.

src/mem/metaslab.h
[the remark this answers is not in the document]
mjp41: A style guide is a good idea.  Let's do it in a different PR ;-)

src/mem/metaslab.h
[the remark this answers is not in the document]
mjp41: I guess my objection was you are saying who calls it.  That isn't a specification, that is a justification for an implementation. I think the specification is 
```
Can be called on a MetaEntry associated with either the front or the back end.
```
The reason is important for stopping someone putting an assert in there in the future.  But that feels internal.  I think if you just reverse the comment to be spec followed by reason. I would be happy.

src/ds/redblacktree.h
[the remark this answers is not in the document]
mjp41: The previous solution allowed the storage location to have a more interesting type that could control the access.  This is different to perhaps the more standard smart pointer approach.  The original example in redblack.cc did perform interesting manipulations that related the stored value to the storage location.  I think your approach is less invasive on the storage locations type by using smartptr type instead.  It does require an additional type `decltype(root)`, which I think of as being `Holder` before this PR.

I think it would be good to switch `Holder` to `Handle` as I don't think `Holder` is the correct description now you have changed its meaning.

src/ds/redblacktree.h
```diff
@@ -41,7 +65,20 @@ namespace snmalloc
...
+    ->ConceptSame<typename Rep::Holder>;
+    {
+      Rep::null
+    }
+    ->ConceptSameModRef<const typename Rep::Contents>;
+    {
+      typename Rep::Holder
+      {
+        const_cast<
+          std::remove_const_t<std::remove_reference_t<decltype(Rep::root)>>*>(
```
mjp41, line 77: I see why you need to remove `const` but not reference types.

src/ds/redblacktree.h
```diff
@@ -106,20 +154,26 @@ namespace snmalloc
...
+      auto printable()
       {
-        return Rep::get(ptr) == Rep::null;
+        return Rep::printable(ptr);
       }
     };
 
     // Root field of the tree
-    H root{};
+    typename std::remove_const_t<std::remove_reference_t<decltype(Rep::root)>>
```
mjp41, line 175: I don't see why you need to remove reference types.

src/mem/corealloc.h
```diff
@@ -328,39 +339,32 @@ namespace snmalloc
...
-
 #if defined(__CHERI_PURE_CAPABILITY__) && !defined(SNMALLOC_CHECK_CLIENT)
       // Zero the whole slab. For CHERI we at least need to clear the freelist
       // pointers to avoid leaking capabilities but we do not need to do it in
       // the freelist order as for SNMALLOC_CHECK_CLIENT. Zeroing the whole slab
       // may be more friendly to hw because it does not involve pointer chasing
       // and is amenable to prefetching.
-      meta->meta_common.template zero_chunk<typename SharedStateHandle::Pal>(
-        snmalloc::sizeclass_to_slab_size(sizeclass));
+      // FIXME: This should be a back-end method guarded on a feature flag.
```
mjp41, line 348: Did you mean to delete the code from here?

mjp41 (approved):
Couple of minor bits.  But otherwise looks good.

src/ds/redblacktree.h
davidchisnall: If `Rep` declares a static `root` of type `T` then `decltype(Rep::root)` will be `T&`.

PR #503, opened 2022-04-04
Header layering proposal.

davidchisnall:
Straw-man header layering, mostly based on prior discussions with some
tweaks from actually trying to do it.

See src/snmalloc/README.md for an explanation of the layers.


src/test/func/jemalloc/jemalloc.cc
```diff
@@ -9,8 +9,8 @@
 #define SNMALLOC_BOOTSTRAP_ALLOCATOR
 #define SNMALLOC_JEMALLOC3_EXPERIMENTAL
 #define SNMALLOC_JEMALLOC_NONSTANDARD
-#include "../../../override/jemalloc_compat.cc"
-#include "../../../override/malloc.cc"
+#include "../../../snmalloc/override/jemalloc_compat.cc"
```
mjp41, line 12: Most tests seem to get away with `<snmalloc/...>`

src/test/func/malloc/malloc.cc
```diff
@@ -6,7 +6,7 @@
 #undef SNMALLOC_NO_REALLOCARRAY
 #undef SNMALLOC_NO_REALLOCARR
 #define SNMALLOC_BOOTSTRAP_ALLOCATOR
-#include "../../../override/malloc.cc"
+#include "../../../snmalloc/override/malloc.cc"
```
mjp41, line 9: Most tests seem to get away with `<snmalloc/...>`

src/test/func/domestication/domestication.cc
```diff
@@ -98,7 +98,7 @@ namespace snmalloc
 }
 
 #  define SNMALLOC_NAME_MANGLE(a) test_##a
-#  include "../../../override/malloc.cc"
+#  include "../../../snmalloc/override/malloc.cc"
```
mjp41, line 101: Most tests use `<snmalloc/...>`

src/snmalloc/backend_helpers/pagemap.h
```diff
@@ -1,9 +1,7 @@
 #pragma once
 
-#include "../ds/bits.h"
-#include "../ds/helpers.h"
+#include "../ds/ds.h"
 #include "../mem/entropy.h"
```
mjp41, line 4: Should this include `entropy.h` directly?

mjp41 (approved):
Like the changes.  I think this makes things much cleaner.  Couple of minor nits and CI isn't happy yet.  I can take a second pass when CI approves.

SchrodingerZhu (approved):

src/snmalloc/README.md
```diff
@@ -0,0 +1,40 @@
...
+This allows specific files to be moved to a lower layer if appropriate, without too much code churn.
+
+There is only one exception to this rule: `backend/globalconfig.h`.
+This file defines either the default configuration *or* nothing, depending on whether the user has defined `SNMALLOC_PROVIDE_OWN_CONFIG`.
+The layers above the back end should include only this file, so that there is a single interception point for externally defined back ends.
+
+External code should include only the following files:
+
+ - `snmalloc/snmalloc_core.h` includes everything up to `backend_helpers`.
+   This provides the building blocks required to assembly an snmalloc instance, but does not assume any global configuration.
```
nwf, line 37:
```suggestion
   This provides the building blocks required to assemble an snmalloc instance, but does not assume any global configuration.
```

src/test/setup.h
```diff
@@ -1,10 +1,9 @@
 #if defined(SNMALLOC_CI_BUILD)
-#  include <pal/pal.h>
+#  include <snmalloc/pal/pal.h>
 #  if defined(WIN32)
-#    include <ds/bits.h>
 #    include <iostream>
-#    include <pal/pal.h>
 #    include <signal.h>
+#    include <snmalloc/ds/bits.h>
```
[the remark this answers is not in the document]
davidchisnall: I was largely ignoring the tests for the purpose of layering:  Nothing should depend on the tests, so they can do whatever they want.

nwf (commented):
The `ds` directory's new contents confuse me.  I'd have imagined that `aba.h`, `csv.h`, `redblacktree.h`, `seqset.h`, and `singleton.h` were suitable for `ds_core` (at a glance, I don't see why they're not?).  `flaglock.h` might plausibly depend on AAL or PAL, but at present it does not seem to.  `mpmcstack.h` depends on `allocconfig.h` only for `CACHELINE_SIZE`, which seems like it could be templated out instead (and, while here, perhaps we should stop this file putting `SNMALLOC_THREAD_SANITIZER_ENABLED` in the namespace?).

davidchisnall:
> The `ds` directory's new contents confuse me. I'd have imagined that `aba.h`,

Depends on the AAL for `pause`.  We could make it templated on the AAL, but since the AAL is a singleton that seems painful to do unless we are going to need it at lower layers.  Even if it were templated, we wouldn't be able to use it below the AAL layer, so it's only a benefit to have it lower if we want to use it in the PAL.  

> `csv.h`,

Was part of the old pre-`MessageBuilder` stats infrastructure, deleted.

> `redblacktree.h`,

Agreed.

> `seqset.h`,

Agreed.

> and `singleton.h`

Depends on `FlagLock`.

> were suitable for `ds_core` (at a glance, I don't see why they're not?). `flaglock.h` might plausibly depend on AAL or PAL, but at present it does not seem to. 

`FlagLock` depends on the AAL for `pause`.

> `mpmcstack.h` depends on `allocconfig.h` only for `CACHELINE_SIZE`, which seems like it could be templated out instead (and, while here, perhaps we should stop this file putting `SNMALLOC_THREAD_SANITIZER_ENABLED` in the namespace?).

Depends on `ABA`, which depends on the AAL for `pause`.

mjp41 (approved):
LGTM

PR #504, opened 2022-04-05
Pagemap move

rmn30:
Hopefully this is roughly what you had in mind, David. Note that I had to make `PageMap::set_metaentry` public as well as the `concretePageMap` field so still things to do.

### DIFF
--- a/src/snmalloc/backend/backend.h
+++ b/src/snmalloc/backend/backend.h
@@ -24,135 +24,72 @@ namespace snmalloc
   class BackendAllocator : public CommonConfig
   {
   public:
+    class PageMapEntry;
     using Pal = PAL;
-
     using SlabMetadata = FrontendSlabMetadata;
 
-    class Pagemap
+  private:
+    using ConcretePagemap =
+      FlatPagemap<MIN_CHUNK_BITS, PageMapEntry, PAL, fixed_range>;
+
+  public:
+    /**
+     * Example of type stored in the pagemap.
+     * The following class could be replaced by:
+     *
+     * ```
+     * using PageMapEntry = FrontendMetaEntry<SlabMetadata>;
+     * ```
+     *
+     * The full form here provides an example of how to extend the pagemap
+     * entries.  It also guarantees that the front end never directly
+     * constructs meta entries, it only ever reads them or modifies them in
+     * place.
+     */
+    class PageMapEntry : public FrontendMetaEntry<SlabMetadata>
     {
+      /**
+       * The private initialising constructor is usable only by this back end.
+       */
       friend class BackendAllocator;
 
-    public:
       /**
-       * Export the type stored in the pagemap.
-       * The following class could be replaced by:
-       *
-       * ```
-       * using Entry = FrontendMetaEntry<SlabMetadata>;
-       * ```
-       *
-       * The full form here provides an example of how to extend the pagemap
-       * entries.  It also guarantees that the front end never directly
-       * constructs meta entries, it only ever reads them or modifies them in
-       * place.
+       * The private default constructor is usable only by the pagemap.
        */
-      class Entry : public FrontendMetaEntry<SlabMetadata>
-      {
-        /**
-         * The private initialising constructor is usable only by this back end.
-         */
-        friend class BackendAllocator;
-
-        /**
-         * The private default constructor is usable only by the pagemap.
-         */
-        friend class FlatPagemap<MIN_CHUNK_BITS, Entry, PAL, fixed_range>;
-
-        /**
-         * The only constructor that creates newly initialised meta entries.
-         * This is callable only by the back end.  The front end may copy,
-         * query, and update these entries, but it may not create them
-         * directly.  This contract allows the back end to store any arbitrary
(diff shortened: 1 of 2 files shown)


davidchisnall:
I think this approach is mostly right.  You've upset GCC because you're shadowing the name of a type: the class `Pagemap` should be called something else so that when you do `using Pagemap = ` you're not introducing a new type with the same name.  Maybe `BasicTemplate`.  

I think that you can avoid making `set_metadata` public by taking the back end as another template parameter and make that a `friend`.  Something like:

```
template<typename Backend, ...>
class DefaultPagemap
{
  friend class Backend;
  ...
};
```

src/snmalloc/backend/backend.h
```diff
@@ -28,132 +28,64 @@ namespace snmalloc
...
-      static bool is_initialised()
-      {
-        return concretePagemap.is_initialised();
-      }
+      SNMALLOC_FAST_PATH PageMapEntry() = default;
     };
 
+    using Pagemap = Pagemap<
+      PAL,
+      FlatPagemap<MIN_CHUNK_BITS, PageMapEntry, PAL, fixed_range>,
```
davidchisnall, line 85: Since this type is used in two places, it would make sense to have a (private) `using` declaration to name the type so that these two can't get out of sync.  Something like:
```
using ConcretePagemap = FlatPagemap<MIN_CHUNK_BITS, PageMapEntry, PAL, fixed_range>;
```
And then use `ConcretePagemap` in both places.

src/snmalloc/backend_helpers/pagemap.h
```diff
@@ -325,4 +325,81 @@ namespace snmalloc
       body[p >> SHIFT] = t;
     }
   };
+
+  template<
+    typename PAL,
+    typename ConcreteMapT,
+    typename EntryT,
```
davidchisnall, line 332: We don't use `T` suffixes for types anywhere else.  Names that start with upper-case letters are types.

src/snmalloc/backend_helpers/pagemap.h
```diff
@@ -325,4 +325,81 @@ namespace snmalloc
...
+  class Pagemap
+  {
+  public:
+    /**
+     * Export the type stored in the pagemap.
+     */
+    using Entry = EntryT;
+
+    SNMALLOC_REQUIRE_CONSTINIT
+    static inline ConcreteMapT concretePagemap;
```
davidchisnall, line 343: Please add a doc comment and a blank line after the end.

src/snmalloc/backend_helpers/pagemap.h
```diff
@@ -325,4 +325,81 @@ namespace snmalloc
...
+     * Set template parameter to true if it not an error
+     * to access a location that is not backed by a chunk.
+     */
+    template<bool potentially_out_of_range = false>
+    SNMALLOC_FAST_PATH static auto& get_metaentry_mut(address_t p)
+    {
+      return concretePagemap.template get_mut<potentially_out_of_range>(p);
+    }
+
+    static void register_range(address_t p, size_t sz)
```
davidchisnall, line 380:
```suggestion
    /**
     * Register a range in the pagemap as in-use, requiring it to allow writing to the underlying memory.
     */
    static void register_range(address_t p, size_t sz)
```

src/snmalloc/backend_helpers/pagemap.h
```diff
@@ -325,4 +325,81 @@ namespace snmalloc
...
+    static SNMALLOC_FAST_PATH
+      std::enable_if_t<fixed_range_, std::pair<address_t, address_t>>
+      get_bounds()
+    {
+      static_assert(fixed_range_ == fixed_range, "Don't set SFINAE parameter!");
+
+      return concretePagemap.get_bounds();
+    }
+
+    static bool is_initialised()
```
davidchisnall, line 400:
```suggestion
    /**
     * Return whether the pagemap is initialised, ready for access.
     */
    static bool is_initialised()
```

src/snmalloc/backend_helpers/pagemap.h
davidchisnall: This should also be `private`.  Only the back end class should touch it directly (and then only for initialisation).

src/snmalloc/backend_helpers/pagemap.h
```diff
@@ -325,4 +325,105 @@ namespace snmalloc
       body[p >> SHIFT] = t;
     }
   };
+
+  /**
+   * This is the backend's interface to the page map. It takes a concrete page
```
davidchisnall, line 330:
```suggestion
* This is a generic implementation of the backend's interface to the page map. It takes a concrete page
```
This isn't the interface, this is one thing that you can choose to use to implement the interface.  You are also free to use something else.

davidchisnall (commented):
Minor nit with a comment, but LGTM.  @ihaller, does this help simplify your use case?  Please approve it if it does (or, at least, doesn't make things worse!).

ihaller (approved):

mjp41 (approved):
LGTM

PR #505, opened 2022-04-08
Remove indirection of state in ranges.

davidchisnall:
This doesn't give any extra flexibility: the range itself can be either
a stateless class, a class with no per-instance state that stores all of
static fields, or a class with stateful instances.  It did add a
requirement that every range implementation added an indirection layer.

Removing this removes over 100LoC with no loss of functionality.

### DIFF
--- a/src/snmalloc/backend_helpers/globalrange.h
+++ b/src/snmalloc/backend_helpers/globalrange.h
@@ -11,27 +11,15 @@ namespace snmalloc
   template<typename ParentRange>
   class GlobalRange
   {
-    typename ParentRange::State parent{};
+    static inline ParentRange parent{};
 
     /**
      * This is infrequently used code, a spin lock simplifies the code
      * considerably, and should never be on the fast path.
      */
-    FlagWord spin_lock{};
+    static inline FlagWord spin_lock{};
 
   public:
-    class State
-    {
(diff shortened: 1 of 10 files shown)


src/snmalloc/backend_helpers/globalrange.h
```diff
@@ -11,28 +11,15 @@ namespace snmalloc
...
      * This is infrequently used code, a spin lock simplifies the code
      * considerably, and should never be on the fast path.
      */
-    FlagWord spin_lock{};
+    static inline FlagWord spin_lock{};
 
   public:
-    class State
-    {
-      SNMALLOC_REQUIRE_CONSTINIT static inline GlobalRange global_range{};
```
nwf, line 25: This drops the `SNMALLOC_REQUIRE_CONSTINIT` here; should we put it on the `GlobalMetaRange global_state` object in `backend.h`?

nwf (approved):
LGTM; one small question.

src/snmalloc/backend_helpers/globalrange.h
mjp41: Should probably be on line 14 and 20 of this file.

davidchisnall: Agreed, fixed.

nwf:
That works; LGTM.

PR #508, opened 2022-04-14
Check things in release builds with check-client.

davidchisnall:
Clang 15 doesn't build the release builds with CHECK_CLIENT enabled
because they are using `SNMALLOC_ASSERT` and so the values that we're
collecting to check are never actually checked.  This is probably a bug
- if we're turning on the checks, I imagine it's because we want them.


mjp41 (approved):
LGTM

PR #510, opened 2022-04-22
Create a 'StrictProvenanceBackend' that restores CHERI functionality

rmn30:
This current draft duplicates way too much code and will have to be
revised.  Essential all it does is provide a custom SlabMetadata class
that extends FrontendSlabMetadata with a chunk_ptr to store the
original pointer to the slab with the original bounds of the mmaped
region (this used to exist but was removed during
refactoring). alloc_range sets this in the SlabMetadata and
dealloc_range uses it instead of the pointer argument (which might
have restricted bounds).

### DIFF
--- a/src/snmalloc/backend/globalconfig.h
+++ b/src/snmalloc/backend/globalconfig.h
@@ -5,6 +5,7 @@
 #ifndef SNMALLOC_PROVIDE_OWN_CONFIG
 
 #  include "../backend/backend.h"
+#include "strictprovenancebackend.h"
 
 namespace snmalloc
 {
@@ -21,18 +22,23 @@ namespace snmalloc
   }
 #  endif
 
+  using BackendAlloc = std::conditional_t<
+    aal_supports<StrictProvenance, Aal>,
+    StrictProvenanceBackend<Pal>,
+    BackendAllocator<Pal, false>>;
+
   /**
    * The default configuration for a global snmalloc.  This allocates memory
    * from the operating system and expects to manage memory anywhere in the
    * address space.
    */
-  class Globals final : public BackendAllocator<Pal, false>
+  class Globals final : public BackendAlloc
   {
   public:
     using GlobalPoolState = PoolState<CoreAllocator<Globals>>;
 
   private:
-    using Backend = BackendAllocator<Pal, false>;
+    using Backend = BackendAlloc;
 
     SNMALLOC_REQUIRE_CONSTINIT
     inline static GlobalPoolState alloc_pool;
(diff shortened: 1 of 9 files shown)


src/snmalloc/backend/strictprovenancebackend.h
```diff
@@ -0,0 +1,241 @@
...
+        address_cast(alloc),
+        Pagemap::get_metaentry(address_cast(alloc)).get_slab_metadata());
+      Pagemap::set_metaentry(address_cast(alloc), size, t);
+
+      local_state.get_meta_range().dealloc_range(
+        capptr::Chunk<void>(&slab_metadata), sizeof(SlabMetadata));
+
+      // On CHERI platforms, we need to re-derive to get a pointer to
+      // the chunk.
+      auto diff = pointer_diff(slab_metadata.chunk_ptr, alloc);
```
mjp41, line 224: `slab_metadata` deallocated by this point (UAF)

nwf:
FWIW, the Morello cluster is very nearly ready to serve as a CI backend going forward; if you have an updated version of this you want me to try putting through that machinery, let me know.

src/snmalloc/backend/strictprovenancebackend.h
```diff
@@ -0,0 +1,241 @@
+#if defined(SNMALLOC_CHECK_CLIENT) && !defined(OPEN_ENCLAVE)
+/**
```
davidchisnall, line 2: We are missing a `#pragma once` at the start of this file.

src/snmalloc/backend/globalconfig.h
```diff
@@ -21,18 +22,23 @@ namespace snmalloc
   }
 #  endif
 
+  using BackendAlloc = std::conditional_t<
```
davidchisnall, line 25: Please add a doc comment for this.

src/snmalloc/backend/strictprovenancebackend.h
```diff
@@ -0,0 +1,241 @@
...
+
+  public:
+    using Pagemap = BasicPagemap<
+      StrictProvenanceBackend,
+      PAL,
+      ConcretePagemap,
+      PageMapEntry,
+      false>;
+
+    static constexpr bool CONSOLIDATE_PAL_ALLOCS = false;
```
davidchisnall, line 54: There are lots of missing doc comments here, but this one in particular is confusing.  Is it actually meant to be public or is it used only in the definition of the ranges below?

mjp41:
Should this be closed now?

rmn30:
I think it contains some changes to clear pointers that didn't make it into @nwf 's version so I should rebase on top of main and resubmit at some point.

mjp41:
It looks like the CLEAN_POINTER changes all made it into #528, and the rest has been superseeded by #537, so I am going to close this.

PR #511, opened 2022-04-27
Refill heuristic

mjp41:
Modify the large buddy allocator to gradually grow its refill size up to a set amount.  This is designed to work well on more limited address space platforms, but be able to grow to reduce contention on larger systems.

Primary use is for OE, where changes from CoffeeLake to IceLake mean the address space limits are very different.


mjp41:
I need to do some perf testing on this to check it hasn't regressed.

davidchisnall (approved):

PR #512, opened 2022-04-27
Refactor capptr_domesticate SFINAE to make more statically safe.

rmn30:
This refactoring was provided by David.  Previously if a backend
provided a capptr_domesticate function with the wrong type it would be
silently ignored.  This change requires backends to explicitly opt in
to domestication via a new Backend::Option and ensures the compiler
will loudly complain if there is a mismatch.

### DIFF
--- a/src/snmalloc/backend/fixedglobalconfig.h
+++ b/src/snmalloc/backend/fixedglobalconfig.h
@@ -24,7 +24,26 @@ namespace snmalloc
       return alloc_pool;
     }
 
-    static constexpr Flags Options{};
+    /*
+     * The obvious
+     * `static constexpr Flags Options{.HasDomesticate = true};` fails on
+     * Ubuntu 18.04 with an error "sorry, unimplemented: non-trivial
+     * designated initializers not supported".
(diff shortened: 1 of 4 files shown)


src/snmalloc/backend/fixedglobalconfig.h
```diff
@@ -24,7 +24,26 @@ namespace snmalloc
       return alloc_pool;
     }
 
-    static constexpr Flags Options{};
+    /*
+     * The obvious
+     * `static constexpr Flags Options{.HasDomesticate = true};` fails on
+     * Ubuntu 18.04 with an error "sorry, unimplemented: non-trivial
```
davidchisnall, line 30: @mjp41, this is another gcc 9.4 work-around.  I wonder if we should add:
```c++
#if defined(__GNUC__) && !defined(__clang__)
#  if __GNUC__ < 10
#    error GCC 9 is too broken, please update to GCC 10 or clang
#  endif
#endif
```

mjp41 (approved):
Looks okay to me.

PR #513, opened 2022-04-28
Simplify asserts introduced in #512.

rmn30:
Note that the second one was incorrect.


davidchisnall (approved):

rmn30:
On second thoughts I think they are equivalent, but simpler is probably better.

mjp41:
Could simplify to a single static assert with `==`, but the current gives more descriptive error messages

rmn30:
Yes, originally we just had the equality but needed the more detailed error message for debugging!

PR #514, opened 2022-04-29
Modified Metadata range to have separate pool

mjp41:
The Metadata range should not be shared with the object range.  This
change ensures that their are separate requests to the Pal for meta-data
and object data ranges.  The requests are never combined, and thus
memory cannot flow from being used in malloc to later be used in meta-
data.


mjp41:
@achamayou this PR should fix the memory reporting on OE.

PR #515, opened 2022-05-03
Enable memcpy checks for check shim

mjp41:
Build three levels of checking
  - None
  - Checks memcpy only
  - Checks (full)

Currently you can build checks without enabling the memcpy protection.
This PR fixes that.


davidchisnall (approved):
Looks fine to me.  The names might need a bit of work, because I read them as right-associative (so assumed that the memcpy-only library contained only memcpy, rather than that it did no checks other than memcpy, but as long as the docs are updated to explain what each is then it's probably fine).

mjp41:
I'll make sure the docs I am writing for 0.6.0 cover this.

PR #517, opened 2022-05-05
Preparation for 0.6.0

mjp41:
* Update documents for 0.6.0
* Add some explanations of the security features

Mostly, this adds a series of descriptions of the security features
we've added to 0.6.0.  There are a collection of graphs and informal
explanations. The documents are meant more for understanding at an
informal level.

We should improve the documentation across all levels, but that is for
after this release.


README.md
```diff
@@ -29,13 +29,24 @@ scenarios that can be problematic for other allocators:
...
+We recommend you read [docs/security](./docs/security/README.md) to find out about the current design, and 
+if you want to dive into the code (./docs/AddressSpace.md) provides a good overview of the allocation and deallocation paths.
 
 [![snmalloc CI](https://github.com/microsoft/snmalloc/actions/workflows/main.yml/badge.svg?branch=master)](https://github.com/microsoft/snmalloc/actions/workflows/main.yml)
 
+# Hardening
+
+There is a hardened version of snmalloc, it contains
+
+*  Randomisation of the allocations relative locations,
```
davidchisnall, line 43:
```suggestion
*  Randomisation of the allocations' relative locations,
```

README.md
```diff
@@ -29,13 +29,24 @@ scenarios that can be problematic for other allocators:
...
+*  Most meta-data is stored separately from allocations, and is proected with guard pages,
```
davidchisnall, line 44:
```suggestion
*  Most meta-data is stored separately from allocations, and is protected with guard pages,
```

docs/security/FreelistProtection.md
```diff
@@ -0,0 +1,107 @@
+# Protecting meta-data
+
+Corrupting an allocator's meta-data is a common pattern for increasing the power of a use-after-free or out-of-bounds write vulnerabilities.  If you can corrupt the allocator's meta-data, then you can take a control gadget in one part of a system, and use it to affect other parts of the system.  There are various approaches to protecting allocator meta-data, the most common are:
```
davidchisnall, line 3: Can you put each sentence on a separate line so that diffs are easy to read and it's easier to review please?

docs/security/FreelistProtection.md
```diff
@@ -0,0 +1,107 @@
...
+
+* make the allocator meta-data hard to find through randomisation
```
ihaller, line 5: Do you also want to call out separating the metadata pool from the allocated ranges? That is key both for CHERI and sandboxing, where the allocated range is considered dangerous for the allocator to interact with. It is part of the "hard to find" paradigm, but more specific and something that people don't normally think about with heap corruption attacks.

davidchisnall: Agreed on putting this in the notes.  With the latest version (on 64-bit platforms), address space never moves between use as metadata and use as allocations, which means that UAFs won't arbitrarily corrupt allocator state.  This is a bigger win with CHERI (which isn't supported in 0.6, but will be in 0.7) when you don't have revocation because it means that there's no path that allows code outside of the allocator to reach metadata.  

For sandboxing use cases this is less relevant because the pagemap still has pointers to the metadata, so anything that can reach the pagemap can find the entries and we need to protect them in some other way (in the process sandboxing code, by allocating them in memory that the untrusted code doesn't have MMU mappings for, in the CHERI sandboxing case we might do this via sealing, but if the allocator is part of your TCB anyway then we may just trust it).

docs/security/FreelistProtection.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Protecting a free queue.
+
+The idea is remarkably simple: a doubly linked list is far harder to corrupt than a single linked list, because you can check its invariant:
+```
+   x.next.prev == x
+```
+In every kind of free list in snmalloc, we now use something a kin to this:
+```
+  f(x.next).prev == g(x, f(x.next))
```
rmn30, line 31: This looks like you are `f(x.next)` as structure pointer but below it says that `f` is XOR with random values?!

nwf (commented):
Whoo, docs! :)

docs/security/FreelistProtection.md
```diff
@@ -0,0 +1,127 @@
...
+```
+  g(a, b) = (a XOR k1) * (b XOR k2)
+```
+where `k1` and `k2` are two randomly chosen 64 bit values.
+This gives a value that is hard to forge that encodes the back edge relationship.
+
+As we build the list, we add this value to the disused object, and when we consume the free list later, we check the value is correct.
+Importantly, the order of construction and consumption have to be the same, which means we can only use queues, and not stacks.
+
+The checks give us a way to detect that the list has not been mutated.
```
rmn30, line 48:
```suggestion
The checks give us a way to detect that the list has not been corrupted.
```
Hopefully it is mutated in useful ways...

achamayou (approved):

SchrodingerZhu (approved):
downstream sync will be scheduled. :)

docs/security/FreelistProtection.md
```diff
@@ -0,0 +1,127 @@
...
+    lea    rcx,[rdi+rsi*1]                     #+Calculate signed_prev location
+    add    rcx,0x8                             #+|  rcx = fs:[r8+rdi*1+0x8]
+    mov    rsi,QWORD PTR fs:[r8+0x2d8]         #+Calculate next signed_prev value
+    add    rsi,rax                             #+|
+    add    rdx,QWORD PTR fs:[r8+0x2e0]         #+|
+    imul   rdx,rsi                             #+|
+    mov    QWORD PTR [rcx],rdx                 #+Store signed_prev for next entry.
+    ret
+```
+The additions specific to handling the checks are marked with `+`.
```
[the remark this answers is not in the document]
rmn30: Ooops, this duplicates @nwf 's suggestion above.

docs/security/FreelistProtection.md
```diff
@@ -0,0 +1,127 @@
...
+Note: the free list header now contains the value that `prev` should contain, which leads to slightly worse x86 codegen.
+For instance the checks introduce `shl rdi,0x4`, which was previously fused with an `lea` instruction without the checks.
+
+## Conclusion
+
+This approach provides a strong defense against corruption of the free lists used in snmalloc.
+This means all inline meta-data has corruption detection.
+The check is remarkably simple for building double free detection, and has far lower memory overhead compared to using an allocation bitmap.
+
+[Next we show hot to randomise the layout of memory in snmalloc, and thus make it harder to guess relative address of a pair of allocations.](./Randomisation.md)
```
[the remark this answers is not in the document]
rmn30: Also a duplicate.

docs/security/FreelistProtection.md
```diff
@@ -0,0 +1,127 @@
...
+```x86asm
+<malloc(unsigned long)>:
+    lea    rax,[rdi-0x1]                       # Check for small size class
+    cmp    rax,0xdfff                          # | zero is considered a large size
+    ja     SLOW_SIZE                           # | to remove from fast path.
+    shr    rax,0x4                             # Lookup size class in table
+    lea    rcx,[size_table]                    # | 
+    movzx  edx,BYTE PTR [rax+rcx*1]            # | 
+    mov    rdi,rdx                             #+Caclulate index into free lists
+    shl    rdi,0x4                             #+|
```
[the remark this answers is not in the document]
mjp41: Actually, it is because lea can only represent shifts upto 3.

docs/security/data/mimalloc-bench.csv.csv
```diff
@@ -0,0 +1,2660 @@
+cfrac, sn-0.6.0, 06.28, 3328, 6.28, 0.00, 0, 435
```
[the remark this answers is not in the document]
mjp41: This is old data now, and I should drop the file.

docs/security/FreelistProtection.md
```diff
@@ -0,0 +1,123 @@
...
+
+Corrupting an allocator's meta-data is a common pattern for increasing the power of a use-after-free or out-of-bounds write vulnerabilities.
+If you can corrupt the allocator's meta-data, then you can take a control gadget in one part of a system, and use it to affect other parts of the system.
+There are various approaches to protecting allocator meta-data, the most common are:
+
+* make the allocator meta-data hard to find through randomisation
+* surround meta-data with guard pages
+* add some level of encryption/checksuming
+
+With the refactoring of the page table ([described earlier](./VariableSizedChunks.md)), we can put all the slab meta-data at hard to find places with randomisation and add guard pages around it.
```
mjp41, line 11: @ihaller @davidchisnall does this say enough about the meta-data protection?

```suggestion
With the refactoring of the page table ([described earlier](./VariableSizedChunks.md)), we can put all the slab meta-data in completely separate regions of memory to the allocations.
Within the meta-data regions, we add randomisation to make the data hard to find, and add large guard regions around the meta-data.
By using completely separate regions of memory for allocations and meta-data we ensure that no dangling allocation can refer to current meta-data.
This is particularly important for CHERI as it means a UAF can be used to corrupt allocator meta-data.

```

docs/security/FreelistProtection.md
mjp41: Okay, so should I add a bullet point like

```suggestion
* make the allocator meta-data hard to find through randomisation
* use completely separate ranges of memory for meta-data and allocations
```

docs/security/README.md
```diff
@@ -0,0 +1,39 @@
+# Hardening snmalloc
+
+The key focus of the 0.6.0 release of snmalloc is security.
+This was inspired by a few different things coming together.
+
+First, we had been discussing with the Microsoft Security Response various research on allocator hardening.
+Saar Amar had been categorising exploits and what features an allocator should have.
+As part of this, we both realised the existing structures of snmalloc made certain things hard to harden, but more interesting we had some ideas for stuff that could advance the state of the art.
+
+Secondly, we had been experimenting with adding support to snmalloc for CHERI.
```
rmn30, line 10:
```suggestion
Secondly, we had been experimenting with adding support to snmalloc for [CHERI](http://www.chericpu.org).
```

docs/security/VariableSizedChunks.md
```diff
@@ -0,0 +1,97 @@
...
+## Finding meta-data quickly
+
+Allocators must map allocations to associated meta-data.
+There are two common approaches for locating this associated meta-data:
+
+* At some specified aligned position relative to a current pointer
+* In a global map
+
+Most allocators use some combination of both.
+In the original, snmalloc design we had a concept of superslab, where the first part represented the meta-data for all the slabs contained in the superslab.
```
rmn30, line 34:
```suggestion
In the original snmalloc design we had a concept of superslab, where the first part represented the meta-data for all the slabs contained in the superslab.
```

rmn30 (approved):
Very minor typos.

PR #518, opened 2022-05-08
Debug win32

mjp41:
There is an infrequent crash on Win32.  This is some logging I was using to investigate it.  It also modifies the failing test to check for allocation failure.


mjp41:
Using Admin to force this in, so that other PRs will have the changes to Contention.  If there are concerns with this PR, add review comments here, and I'll raise a subsequent PR to address them.

PR #519, opened 2022-05-09
Out-of-memory can fail silently

mjp41:
If this test fails to allocate memory, that should not cause the test to
fail.  The 'abort' was added previously to confirm a infrequent failure
was caused by out-of-memory causing the test to assign to nullptr.

This was confirmed in a CI run, and now the test can be made to ignore
allocation failure.

### DIFF
--- a/src/test/perf/contention/contention.cc
+++ b/src/test/perf/contention/contention.cc
@@ -90,7 +90,7 @@ void test_tasks_f(size_t id)
     else
     {
       std::cout << "Failed to allocate " << size << " bytes" << std::endl;
-      abort();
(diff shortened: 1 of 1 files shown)


mjp41:
Here is the previous failed output:
```
2/50 Test  #2: perf-contention-check ..............Exit code 0xc000040[9](https://github.com/microsoft/snmalloc/runs/6343044521?check_suite_focus=true#step:5:9)
***Exception:   5.34 sec
Allocator is snmalloc
Sequential setup
Begin parallel test:
Failed to allocate 634 bytes
*****ABORT******
Failed to allocate 406 bytes
```

PR #520, opened 2022-05-10
Check size is correct if specified

mjp41:
Some secure allocators check that the C++ supplied size is correct
relative to the meta-data. This adds a check to the secure version of
snmalloc to do that.


nwf (approved):
LGTM; perhaps in the fullness of time we should have `dealloc_cheri_checks` also call `check_size` (there's a TODO therein about possibly enforcing this policy, too).

PR #521, opened 2022-05-10
debug_check_empty

mjp41:
Currently a failing debug_check_empty does not provide any information.
This change allows it to print the size of the one of the allocations
that has not been freed.


nwf (approved):
LGTM

PR #522, opened 2022-05-10
Fix markdown

panekj:


mjp41 (approved):
LGTM, thanks for fixing

panekj:
Oops, wrong file

panekj:
There is missing use of `[^mmu-perms]` in `StrictProvenance.md`

nwf (approved):
Thanks for the fix; the missing `mmu-perms` note looks to have always been detached (oops!).  It should probably go at the end of the sentence "Borrowing terminology from CHERI, we speak of the authority (to a subset of the address space) held by a pointer and will justify actions in terms of this authority.".  If you want to add that, feel free, otherwise I'll track it locally and roll it into a later PR.

PR #527, opened 2022-05-17
Add a pipe operation to ranges

mjp41:
The ranges are naturally put together with pipes. This
commit does some template magic to make the code more
readable.  There should be now functional changes with
this change.


src/snmalloc/backend_helpers/range_helpers.h
```diff
@@ -34,4 +34,39 @@ namespace snmalloc
       length -= align;
     }
   }
+
+  /**
+   * Forward definition to allow multiple template specialisations.
+   *
+   * This struct is used to recursively compose ranges.
```
mjp41, line 41:
```suggestion
   * This struct is used to recursively compose ranges.
```

src/snmalloc/backend_helpers/largebuddyrange.h
```diff
@@ -327,6 +328,17 @@ namespace snmalloc
     }
 
   public:
+    /**
+     * We use a nested Apply type to enable a Pipe operation.
+     */
+    template<typename ParentRange2>
```
davidchisnall, line 334: Please can we have this comment on all of the other ranges?

davidchisnall (approved):
Looks really nice!

src/snmalloc/backend_helpers/largebuddyrange.h
mjp41: I guess, I got bored about half way through.  I have a cuppa now, so can probably do the rest ;-)

mjp41: Done

rmn30 (approved):
Very clever! This does improve readability of the backend definition and the boilerplate added to the `Range` classes is fairly minimal. Maybe `Pipe` could be  `PipeRange` since it is effectively a `Range`?

nwf (approved):

PR #530, opened 2022-05-20
Refactor interface between backend and frontend

mjp41:
Primarily adding for CI feedback.  

I'm going to try and use this in some of scenarios, before I move it out of draft.

### DIFF
--- a/src/snmalloc/backend/backend.h
+++ b/src/snmalloc/backend/backend.h
@@ -22,9 +22,11 @@ namespace snmalloc
    */
   template<
     SNMALLOC_CONCEPT(ConceptPAL) PAL,
-    bool fixed_range,
     typename PageMapEntry,
-    typename Pagemap>
+    typename Pagemap,
+    typename LocalState,
+    typename GlobalMetaRange,
+    typename Stats>
   class BackendAllocator
   {
   public:
@@ -32,120 +34,6 @@ namespace snmalloc
     using SlabMetadata = FrontendSlabMetadata;
 
   public:
-#if defined(_WIN32) || defined(__CHERI_PURE_CAPABILITY__)
-    static constexpr bool CONSOLIDATE_PAL_ALLOCS = false;
-#else
-    static constexpr bool CONSOLIDATE_PAL_ALLOCS = true;
-#endif
-
-    // Set up source of memory
-    using Base = std::conditional_t<
-      fixed_range,
-      EmptyRange,
-      Pipe<
-        PalRange<Pal>,
-        PagemapRegisterRange<Pagemap, CONSOLIDATE_PAL_ALLOCS>>>;
-
-    static constexpr size_t MinBaseSizeBits()
-    {
-      if constexpr (pal_supports<AlignedAllocation, PAL>)
-      {
-        return bits::next_pow2_bits_const(PAL::minimum_alloc_size);
-      }
-      else
-      {
-        return MIN_CHUNK_BITS;
-      }
-    }
-
-    // Global range of memory
-    using GlobalR = Pipe<
-      Base,
-      LargeBuddyRange<24, bits::BITS - 1, Pagemap, MinBaseSizeBits()>,
-      LogRange<2>,
-      GlobalRange<>>;
-
-#ifdef SNMALLOC_META_PROTECTED
-    // Introduce two global ranges, so we don't mix Object and Meta
-    using CentralObjectRange = Pipe<
-      GlobalR,
-      LargeBuddyRange<24, bits::BITS - 1, Pagemap, MinBaseSizeBits()>,
-      LogRange<3>,
-      GlobalRange<>>;
-
-    using CentralMetaRange = Pipe<
-      GlobalR,
-      SubRange<PAL, 6>, // Use SubRange to introduce guard pages.
-      LargeBuddyRange<24, bits::BITS - 1, Pagemap, MinBaseSizeBits()>,
-      LogRange<4>,
-      GlobalRange<>>;
-
-    // Source for object allocations
-    using StatsObject =
-      Pipe<CentralObjectRange, CommitRange<PAL>, StatsRange<>>;
-
-    using ObjectRange =
-      Pipe<StatsObject, LargeBuddyRange<21, 21, Pagemap>, LogRange<5>>;
-
-    using StatsMeta = Pipe<CentralMetaRange, CommitRange<PAL>, StatsRange<>>;
-
-    using MetaRange = Pipe<
-      StatsMeta,
-      LargeBuddyRange<21 - 6, bits::BITS - 1, Pagemap>,
-      SmallBuddyRange<>>;
-
-    // Create global range that can service small meta-data requests.
-    // Don't want to add this to the CentralMetaRange to move Commit outside the
-    // lock on the common case.
-    using GlobalMetaRange = Pipe<StatsMeta, SmallBuddyRange<>, GlobalRange<>>;
-    using Stats = StatsCombiner<StatsObject, StatsMeta>;
-#else
-    // Source for object allocations and metadata
-    // No separation between the two
-    using Stats = Pipe<GlobalR, StatsRange<>>;
-    using ObjectRange = Pipe<
-      Stats,
-      CommitRange<PAL>,
-      LargeBuddyRange<21, 21, Pagemap>,
-      SmallBuddyRange<>>;
-    using GlobalMetaRange = Pipe<ObjectRange, GlobalRange<>>;
-#endif
-
-    struct LocalState
-    {
-      ObjectRange object_range;
-
-#ifdef SNMALLOC_META_PROTECTED
-      MetaRange meta_range;
-
-      MetaRange& get_meta_range()
-      {
-        return meta_range;
-      }
-#else
-      ObjectRange& get_meta_range()
-      {
-        return object_range;
-      }
-#endif
-    };
-
-  public:
-    template<bool fixed_range_ = fixed_range>
-    static std::enable_if_t<fixed_range_> init(void* heap_base, size_t heap_length)
-    {
-      static_assert(fixed_range_ == fixed_range, "Don't set SFINAE parameter!");
-
-      // Push memory into the global range.
-      range_to_pow_2_blocks<MIN_CHUNK_BITS>(
-        capptr::Chunk<void>(heap_base),
-        heap_length,
-        [&](capptr::Chunk<void> p, size_t sz, bool) {
-          GlobalR g;
-          g.dealloc_range(p, sz);
-        });
-    }
-
     /**
      * Provide a block of meta-data with size and align.
      *
--- a/src/snmalloc/backend/backend.h
+++ b/src/snmalloc/backend/backend.h
@@ -20,11 +20,13 @@ namespace snmalloc
    * This class implements the standard backend for handling allocations.
    * It abstracts page table management and address space management.
    */
-  template<SNMALLOC_CONCEPT(ConceptPAL) PAL, bool fixed_range>
+  template<
+    SNMALLOC_CONCEPT(ConceptPAL) PAL,
+    bool fixed_range,
+    typename PageMapEntry>
   class BackendAllocator
   {
   public:
-    class PageMapEntry;
     using Pal = PAL;
     using SlabMetadata = FrontendSlabMetadata;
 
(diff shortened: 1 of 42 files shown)


mjp41:
This turned out to be a fairly large change.  The first few commits are well structured, but after that it degenerates a bit.  I can break into two PRs if you think it would be useful.

src/snmalloc/backend/globalconfig.h
```diff
@@ -5,35 +5,65 @@
...
-  }
-#  endif
-
   /**
    * The default configuration for a global snmalloc.  This allocates memory
    * from the operating system and expects to manage memory anywhere in the
    * address space.
    */
-  class Globals final : public BackendAllocator<Pal, false>
+  class Globals final : public CommonConfig
```
davidchisnall, line 34: If I understand correctly, there is now a config class the exposes a backend?  It would be good to document the separation of concerns between these two things.

src/snmalloc/backend/globalconfig.h
```diff
@@ -5,35 +5,65 @@
...
 
   private:
-    using Backend = BackendAllocator<Pal, false>;
+    using ConcretePagemap =
+      FlatPagemap<MIN_CHUNK_BITS, PageMapEntry, Pal, false>;
+
+    using Pagemap = BasicPagemap<Pal, ConcretePagemap, PageMapEntry, false>;
+
+#  if defined(_WIN32) || defined(__CHERI_PURE_CAPABILITY__)
+    static constexpr bool CONSOLIDATE_PAL_ALLOCS = false;
```
davidchisnall, line 47: I think this is moved code, rather than new code, but it might be a good idea to make this a thing that either the PAL or AAL has the option of setting to false and initialising it as something like `pal_feature<ConsolidateAllocs> && aal_feature<ConsolidateAllocs>`

src/snmalloc/backend/globalconfig.h
```diff
@@ -70,11 +100,7 @@ namespace snmalloc
       key_global = FreeListKey(entropy.get_free_list_key());
 
       // Need to initialise pagemap.
-      Backend::init();
-
-#  ifdef USE_SNMALLOC_STATS
-      atexit(snmalloc::print_stats);
-#  endif
+      Pagemap::concretePagemap.init();
```
davidchisnall, line 103: Does the back end not need initialising anymore?  It might be nice to keep that path, even if it's an always-inline empty function, so that if the back end does need initialising in the future then we have a hook for doing it in place.

src/snmalloc/backend/meta_protected_range.h
```diff
@@ -0,0 +1,71 @@
...
+    typename Pagemap,
+    typename Base,
+    size_t MinSizeBits = MinBaseSizeBits<PAL>()>
+  struct MetaProtectedRangeLocalState
+  {
+  private:
+    // Global range of memory
+    using GlobalR = Pipe<
+      Base,
+      LargeBuddyRange<24, bits::BITS - 1, Pagemap, MinSizeBits>,
```
davidchisnall, line 18: Where does 24 come from here?  Can that be a symbolic constant?

src/snmalloc/backend/meta_protected_range.h
```diff
@@ -0,0 +1,71 @@
...
+      LogRange<2>,
+      GlobalRange<>>;
+
+    // Central source of object-range, does not pass back to GlobalR as
+    // that would allow flows from Objects to Meta-data, and thus UAF
+    // would be able to corrupt meta-data.
+    using CentralObjectRange = Pipe<
+      GlobalR,
+      LargeBuddyRange<24, bits::BITS - 1, Pagemap>,
```
davidchisnall, line 27: Same question.  Does this 24 need to match the one above, or are they independent.

src/snmalloc/backend/meta_protected_range.h
```diff
@@ -0,0 +1,71 @@
...
+      LogRange<3>,
```
davidchisnall, line 28: What is the number here?

src/snmalloc/backend/meta_protected_range.h
```diff
@@ -0,0 +1,71 @@
...
+      GlobalRange<>,
+      CommitRange<PAL>,
+      StatsRange<>>;
+
+    // Centralised source of meta-range
+    using CentralMetaRange = Pipe<
+      GlobalR,
+      SubRange<PAL, 6>, // Use SubRange to introduce guard pages.
```
davidchisnall, line 36: 6?

src/snmalloc/backend/meta_protected_range.h
```diff
@@ -0,0 +1,71 @@
...
+      LargeBuddyRange<24, bits::BITS - 1, Pagemap>,
+      LogRange<4>,
```
davidchisnall, line 38: 4?

src/snmalloc/backend/meta_protected_range.h
```diff
@@ -0,0 +1,71 @@
...
+      GlobalRange<>,
+      CommitRange<PAL>,
+      StatsRange<>>;
+
+    // Local caching of object range
+    using ObjectRange =
+      Pipe<CentralObjectRange, LargeBuddyRange<21, 21, Pagemap>, LogRange<5>>;
```
davidchisnall, line 45: So many numbers!

src/snmalloc/backend/standard_range.h
```diff
@@ -0,0 +1,43 @@
...
+{
+  template<
+    typename PAL,
+    typename Pagemap,
+    typename Base = EmptyRange,
+    size_t MinSizeBits = MinBaseSizeBits<PAL>()>
+  struct StandardLocalState
+  {
+    // Global range of memory, expose this so can be filled by init.
+    using GlobalR = Pipe<
```
davidchisnall, line 17: This looks as if it's the same as in the meta-protected one.  Can it be factored out into a common superclass?

src/snmalloc/backend/meta_protected_range.h
mjp41: I really wanted a string, but I couldn't make that work, so a number was a poor mans string.

src/snmalloc/backend/globalconfig.h
mjp41: Can we punt that to a subsequent PR?

src/snmalloc/backend/meta_protected_range.h
davidchisnall: I see.  It's fairly easy with C++20, so we can hopefully fix that soon.

src/snmalloc/backend/globalconfig.h
mjp41: The call was there before as the Backend was responsible for initialising the Backend. The Backend is a purely static class and has no state.  The LocalState is what might need initialising or things inside it, and that can be done by the Config.

src/snmalloc/backend/backend.h
```diff
@@ -24,11 +24,12 @@ namespace snmalloc
     SNMALLOC_CONCEPT(ConceptPAL) PAL,
     typename PageMapEntry,
     typename Pagemap,
-    typename LocalState,
-    typename GlobalMetaRange,
-    typename Stats>
+    typename LocalState>
```
nwf-msr, line 14: Perhaps squash this into the previous commit?

src/test/setup.h
```diff
@@ -64,7 +64,7 @@ void print_stack_trace()
 void _cdecl error(int signal)
 {
   snmalloc::UNUSED(signal);
-  snmalloc::Pal::message("*****ABORT******");
+  snmalloc::DefaultPal::message("*****ABORT******");
```
nwf-msr, line 67: Squash this commit into the "remove global Pal" commit?

src/snmalloc/backend/standard_range.h
mjp41: Not easily, I have just tried, but due to the way templates and inheritance work, I am writing considerably more code, and I don't think we are making it more readable.

nwf-msr (commented):
I think I'm missing the big picture of what's being changed and perhaps some rationale as well.  AFAICT the code itself seems fine, but I'm feeling a bit lost.

src/snmalloc/backend/backend.h
```diff
@@ -1,223 +1,27 @@
...
 {
   /**
    * This class implements the standard backend for handling allocations.
    * It abstracts page table management and address space management.
    */
-  template<SNMALLOC_CONCEPT(ConceptPAL) PAL, bool fixed_range>
-  class BackendAllocator : public CommonConfig
+  template<
+    SNMALLOC_CONCEPT(ConceptPAL) PAL,
+    typename PageMapEntry,
```
ihaller, line 12: Does this alias with the PageMapEntry class name? Would it make sense to have a slight variation of the name just to avoid confusion?

src/snmalloc/mem/backend_concept.h
```diff
@@ -110,8 +109,7 @@ namespace snmalloc
   template<typename Globals>
   concept ConceptBackendGlobals =
```
ihaller, line 110: Should the Backend inner class and its methods be captured here? Especially regarding the interfaces that are needed by the other layers to work.

mjp41: Yes, I have normally been very lazy with Concepts, and waited for @nwf-msr to get annoyed and add them, but I will try to be better ;-)

mjp41: Done, probably not as well as Wes, but he can take a pass in a subsequent PR.

src/snmalloc/backend/backend.h
mjp41: Made the class `DefaultPagemapEntry`, so it is clear when it is used to instantiate something `PagemapEntry`.

src/snmalloc/backend_helpers/commonconfig.h
```diff
@@ -111,6 +111,64 @@ namespace snmalloc
...
+
+      /**
+       * The private default constructor is usable only by the pagemap.
+       */
+      template<
+        size_t GRANULARITY_BITS,
+        typename T,
+        typename PAL,
+        bool has_bounds>
+      friend class FlatPagemap;
```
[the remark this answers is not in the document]
mjp41: I couldn't make the instantiation order work with it being a template parameter due to moving things around.

src/snmalloc/pal/pal.h
```diff
@@ -64,22 +68,13 @@ namespace snmalloc
...
 
-  using Pal =
-#if defined(SNMALLOC_MEMORY_PROVIDER)
-    PALPlainMixin<SNMALLOC_MEMORY_PROVIDER>;
-#elif defined(OPEN_ENCLAVE)
-    PALOpenEnclave;
-#else
-    DefaultPal;
-#endif
-
```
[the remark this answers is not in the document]
mjp41: I had some issues attempting to make this change.  I have not made it worse, so I am going to leave, and you are welcome to make changes in a subsequent PR.

src/snmalloc/backend/backend.h
```diff
@@ -5,7 +5,8 @@ namespace snmalloc
 {
   /**
    * This class implements the standard backend for handling allocations.
-   * It abstracts page table management and address space management.
+   * It is parameterised by its page table management (Pagemap) and
```
davidchisnall, line 8: Page table management is a bit confusing as a term for userspace things, which don't see architectural page tables.

mjp41:
```suggestion
   * It is parameterised by its Pagemap management and
```

src/snmalloc/backend/globalconfig.h
```diff
@@ -5,45 +5,110 @@
 #ifndef SNMALLOC_PROVIDE_OWN_CONFIG
 
 #  include "../backend/backend.h"
```
nwf-msr, line 7: Nit: while you're in the neighborhood, this can just be
```suggestion
#  include "backend.h"
```

src/snmalloc/backend/globalconfig.h
```diff
@@ -4,46 +4,111 @@
 // `snmalloc.h` or consume the global allocation APIs.
 #ifndef SNMALLOC_PROVIDE_OWN_CONFIG
 
-#  include "../backend/backend.h"
+#  include "backend.h"
+#  include "../backend_helpers/backend_helpers.h"
+#  include "meta_protected_range.h"
+#  include "standard_range.h"
```
mjp41, line 10:
```suggestion
#  include "../backend_helpers/backend_helpers.h"
#  include "backend.h"
#  include "meta_protected_range.h"
#  include "standard_range.h"
```

PR #532, opened 2022-05-25
Add Morello CI

nwf-msr:
For review and commentary.  If you want to test this and need access to the cluster, please let me know.  We should not merge this until R&CT 4968 is fully approved and the automation is actually set up in the right place on Azure.  (If that happens before CHERI support in `snmalloc`, we can just disable the Purecap builds herein.)

See https://github.com/nwf-msr/snmalloc/actions/runs/2385794002 for an example run.  The Purecap Release variant does not, at present, build, due to https://github.com/CTSRD-CHERI/llvm-project/issues/618, and the Purecap Debug variant does not build either, presumably just because we're missing a CHERI AAL and PAL.  (I haven't looked in detail.)

### DIFF
--- a/.github/workflows/morello.yml
+++ b/.github/workflows/morello.yml
@@ -1,0 +1,93 @@
+name: snmalloc CI for Morello
+
+# Controls when the workflow will run
+on:
+  # Triggers the workflow on push or pull request events but only for the master branch
+  push:
+    branches: [ main ]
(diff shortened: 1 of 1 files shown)


mjp41 (commented):
As far as I can tell this looks good.

davidchisnall (approved):

PR #533, opened 2022-05-25
Haiku build fix post refactoring.

devnexen:


src/snmalloc/pal/pal_haiku.h
```diff
@@ -37,6 +37,16 @@ namespace snmalloc
       SNMALLOC_ASSERT(is_aligned_block<page_size>(p, size));
       posix_madvise(p, size, POSIX_MADV_DONTNEED);
     }
+
+    /**
+     * Hopefully a temporary workaround until the kernel random feature
+     * is exposed properly in the userspace ?
+     */
+
```
mjp41, line 45: Extraneous new line.
```suggestion
```

mjp41 (approved):
LGTM

src/snmalloc/pal/pal_haiku.h
```diff
@@ -37,6 +37,15 @@ namespace snmalloc
       SNMALLOC_ASSERT(is_aligned_block<page_size>(p, size));
       posix_madvise(p, size, POSIX_MADV_DONTNEED);
     }
+
+    /**
+     * Hopefully a temporary workaround until the kernel random feature
+     * is exposed properly in the userspace ?
```
davidchisnall, line 43: Is there a Haiku bug that is tracking this?  If so, it would be great to link to it in the comment so that the next person to revisit this can easily see if it's still needed.

devnexen: there is none I just read the os source code and see some good promises regarding this but haiku is still in beta.

Issue #485, opened 2022-03-20
Large OS pages support


misakikasumi:
Hi,
I'm trying to make snmalloc use large OS pages in my application. My use case has been mentioned in https://github.com/microsoft/snmalloc/issues/484#issuecomment-1073236187. My application is targeting Windows. It is memory bottlenecked and I want to reduce page faults. On Linux, 4KB pages can be merged into 2MB large pages implicitly and automatically, but on Windows large page allocation is explicit.

[I have tried to modify the PAL to use large pages](https://github.com/microsoft/snmalloc/commit/662e209a9528250e19f02a71e3b9cb5b3c7a5172). The key modifications are:
```C++
static SNMALLOC_CONSTINIT_STATIC size_t minimum_alloc_size = 0x200000;    // 2MB
static constexpr size_t page_size = 0x200000;                             // 2MB

// in `notify_using`
void* r = VirtualAlloc(p, size, MEM_COMMIT | MEM_LARGE_PAGES, PAGE_READWRITE);
```

This PAL works well with snmalloc1, except the need of some small patches (`alignas(OS_PAGE_SIZE)` not possible, and need to check `rsize > OS_PAGE_SIZE` before `rsize - OS_PAGE_SIZE`). I don't know there are flaws or not but it can just run.

snmalloc2 cannot work with this PAL. It crashes at:
https://github.com/microsoft/snmalloc/blob/26324e8bfc28eda735a3335f83f3f1d8f21cc4a0/src/mem/corealloc.h#L690

I'm wondering whether I can use PAL to enable large os page support and expect snmalloc1 to work perfectly, and why this don't work with snmalloc2.

mjp41:
Would you be able to post the precise commit you used on `snmalloc2`?  I would have expected snmalloc2 to work better with Large Pages than `snmalloc1`.  A lot of the redesign was specifically for that, and getting this working is on my list.

misakikasumi:
Here is the PAL I used with snmalloc2

```diff
diff --git a/src/pal/pal_windows.h b/src/pal/pal_windows.h
index 6d64a25..98f9759 100644
--- a/src/pal/pal_windows.h
+++ b/src/pal/pal_windows.h
@@ -15,13 +15,6 @@
 #  include <windows.h>
 #  pragma comment(lib, "bcrypt.lib")
 #  include <bcrypt.h>
-// VirtualAlloc2 is exposed in RS5 headers.
-#  ifdef NTDDI_WIN10_RS5
[...]

PR #534, opened 2022-05-30
Alter PAGE_SIZE usage

mjp41:
Using PAGE_SIZE as the minimum size of the CHUNK means that if this is
configured to 2MiB, then there is a gap between
MAX_SMALL_SIZECLASS_SIZE, and MIN_CHUNK_SIZE, and thus
we can't represent certain sizes.

This removes that connection, and adds a static assert to check for the issue.

### DIFF
--- a/src/snmalloc/backend/meta_protected_range.h
+++ b/src/snmalloc/backend/meta_protected_range.h
@@ -40,6 +40,12 @@ namespace snmalloc
       LogRange<2>,
       GlobalRange<>>;
 
+    static constexpr size_t page_size_bits =
+      bits::next_pow2_bits_const(PAL::page_size);
+
+    static constexpr size_t max_page_chunk_size_bits =
+      bits::max(page_size_bits, MIN_CHUNK_BITS);
+
     // Central source of object-range, does not pass back to GlobalR as
     // that would allow flows from Objects to Meta-data, and thus UAF
     // would be able to corrupt meta-data.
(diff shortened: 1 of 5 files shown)


mjp41:
#485 I believe this will fix the crashes that @misakikasumi observed when trying to build Windows Huge page support.

mjp41:
@nwf-msr this primarily affects your PowerPC setup, so if you could make sure I haven't broken anything that would be great.

nwf-msr:
Ah, a little prodding and some belated understanding:
https://github.com/microsoft/snmalloc/blob/1b8aa6bc0d33e06c3f8d2ecf26ba3601da5c6967/src/snmalloc/backend_helpers/smallbuddyrange.h#L172-L180
needs to grab things in multiples of `OS_PAGE_SIZE` otherwise the `mprotect` inside
https://github.com/microsoft/snmalloc/blob/1b8aa6bc0d33e06c3f8d2ecf26ba3601da5c6967/src/snmalloc/pal/pal_posix.h#L227-L242 
will return `EINVAL` (as it is here; should we `SNMALLOC_ASSERT` on that returning 0?).  The assert isn't saving us because CommitRange passes `NoZero`:
https://github.com/microsoft/snmalloc/blob/1b8aa6bc0d33e06c3f8d2ecf26ba3601da5c6967/src/snmalloc/backend_helpers/commitrange.h#L26-L32

Maybe the PAL's assertion should actually be
```
is_aligned_block<OS::page_size>(p, size) || (!PalEnforceAccess && (zero_mem == NoZero)) 
```
or maybe just
```
is_aligned_block<OS::page_size>(p, size)
```
since `PalEnforceAccess` is a static, global configuration knob?

mjp41:
Moving to draft.  To fix this issue is going to conflict with #530.  To fix, we need 

* Explicit `OS_PAGE_SIZE` as the minimum size for the `LargeBuddyRange` in a few places so that we don't make sub page requests to the `CommitRange`.
* Add assert to `CommitRange`, so it fails more cleanly
* Probably a `static_assert` to check `minimium size <= request_size` on `LargeBuddyRange`.

nwf-msr (approved):
LGTM and passes on POWER

PR #535, opened 2022-05-30
RFC, NFC: refactor ranges to be nested templates

nwf-msr:
This way, we don't have to specify a Parent when we're just interested in
Pipe-ing things together.

We could have called these inner classes Apply and left the Pipe implementation
alone, but it's probably better to call them Type and adjust the Pipe code.

This commit is **absolutely** best reviewed ignoring whitespace changes!

This commit also conflicts, for trivial reasons, with #530.  I would be tempted to wait until that has landed and then commute this back across that if we want this.

### DIFF
--- a/src/snmalloc/backend/meta_protected_range.h
+++ b/src/snmalloc/backend/meta_protected_range.h
@@ -38,7 +38,7 @@ namespace snmalloc
         Pagemap,
         MinSizeBits>,
       LogRange<2>,
-      GlobalRange<>>;
+      GlobalRange>;
 
     static constexpr size_t page_size_bits =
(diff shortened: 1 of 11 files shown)


mjp41:
So the nested type was my initial plan for how to do this. I discussed with @davidchisnall and he suggested the default and inner using.

I implemented both and went with the using due to the smaller diff. 

I do find the occasional <> annoying, and I think the nested type is slightly clearer code. But I'm not sure if it is worth changing.

mjp41 (approved):
I am happy with this.  I don't have strong feelings about it either way.  This is more how I originally thought about types, so I'll approve.

PR #536, opened 2022-05-31
Add a TID to the PAL

mjp41:
Making this part of the PAL allows other platforms to replace with
something more suitable.

### DIFF
--- a/src/snmalloc/ds/flaglock.h
+++ b/src/snmalloc/ds/flaglock.h
@@ -15,6 +15,8 @@ namespace snmalloc
    */
   struct DebugFlagWord
   {
+    using ThreadIdentity = DefaultPal::ThreadIdentity;
+
     /**
      * @brief flag
      * The underlying atomic field.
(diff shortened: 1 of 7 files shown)


src/snmalloc/ds/flaglock.h
```diff
@@ -56,24 +57,20 @@ namespace snmalloc
     }
 
   private:
-    using ThreadIdentity = int const*;
-
+    using ThreadIdentity = size_t;
```
davidchisnall, line 60: Should this be a type provided by the PAL?  I can imagine some platforms might want to use a pointer to a thread control block, for example, rather than an integer.

davidchisnall (approved):

src/snmalloc/ds/flaglock.h
```diff
@@ -56,24 +59,19 @@ namespace snmalloc
...
      */
-    std::atomic<ThreadIdentity> owner = nullptr;
+    std::atomic<ThreadIdentity> owner = ThreadIdentity();
 
     /**
      * @brief get_thread_identity
      * @return The identity of current thread.
      */
-    inline ThreadIdentity get_thread_identity()
+    static size_t get_thread_identity()
```
nwf-msr, line 72: I think this should still be a `ThreadIdentity` even if they're always `size_t`-s right now.

PR #537, opened 2022-05-31
StrictProvenance backend support, take 2022.05

nwf-msr:
Here we go again.  This gets easier every time, at least.  Especially after all the recent backend refactoring, I would almost dare suggest it was _straightforward_.  This PR should be reviewable commit-by-commit.  It has a prefix of preparatory NFC commits that don't (well, aren't supposed to) change anything and then the rest that actually do something.

This series re-introduces `capptr::Arena` and friends and pushes that through the backend ranges.  The `LargeBuddyRange`-s always deal with `Arena`-bounded pointers in both directions; their consumers -- the `backend` `chunk_alloc`-ator and the `SmallBuddyRange` -- know how to tighten those down to `Chunk`-bounded pointers.  For chunks that are being handed out to the frontend as such, the backend stashes the `Arena`-bounded authority inside the `SlabMetadata` using a wrapper type (for non-`StrictProvenance` architectures, there is a similar wrapper type in the same place that provides the usual sleight of hand).  For chunks used by the `SmallBuddyRange`, it relies on a `ProvenanceCaptureRange` to hold the `Arena`-bounded authority within the `Pagemap` itself, rather like the `LargeBuddyRange`, while communicating using `Chunk`-bounded pointers with its child range(s).

This is missing the increased care around zeroing out internal pointers present in #510, so for that reason, if no other, is probably not ready to merge.

### DIFF
--- a/src/snmalloc/backend_helpers/defaultpagemapentry.h
+++ b/src/snmalloc/backend_helpers/defaultpagemapentry.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "../mem/mem.h"
+#include "cheri_slabmetadata_mixin.h"
 
 namespace snmalloc
 {
@@ -63,6 +64,9 @@ namespace snmalloc
     SNMALLOC_FAST_PATH DefaultPagemapEntryT() = default;
   };
 
-  using DefaultPagemapEntry = DefaultPagemapEntryT<FrontendSlabMetadata>;
+  using DefaultPagemapEntry = DefaultPagemapEntryT<std::conditional_t<
+    backend_strict_provenance,
+    StrictProvenanceSlabMetadataMixin<FrontendSlabMetadata>,
+    LaxProvenanceSlabMetadataMixin<FrontendSlabMetadata>>>;
 
 } // namespace snmalloc
--- a/src/snmalloc/backend_helpers/defaultpagemapentry.h
+++ b/src/snmalloc/backend_helpers/defaultpagemapentry.h
@@ -9,7 +9,7 @@ namespace snmalloc
    * The following class could be replaced by:
    *
    * ```
-   * using DefaultPagemapEntry = FrontendMetaEntry<SlabMetadata>;
+   * using DefaultPagemapEntry = FrontendMetaEntry<FrontendSlabMetadata>;
    * ```
    *
    * The full form here provides an example of how to extend the pagemap
@@ -17,7 +17,8 @@ namespace snmalloc
    * constructs meta entries, it only ever reads them or modifies them in
    * place.
    */
-  class DefaultPagemapEntry : public FrontendMetaEntry<FrontendSlabMetadata>
+  template<typename SlabMetadata>
+  class DefaultPagemapEntryT : public FrontendMetaEntry<SlabMetadata>
   {
     /**
      * The private initialising constructor is usable only by this back end.
@@ -43,22 +44,25 @@ namespace snmalloc
      * metadata in meta entries when they are first constructed.
      */
     SNMALLOC_FAST_PATH
-    DefaultPagemapEntry(FrontendSlabMetadata* meta, uintptr_t ras)
-    : FrontendMetaEntry<FrontendSlabMetadata>(meta, ras)
+    DefaultPagemapEntryT(SlabMetadata* meta, uintptr_t ras)
+    : FrontendMetaEntry<SlabMetadata>(meta, ras)
     {}
 
     /**
      * Copy assignment is used only by the pagemap.
      */
-    DefaultPagemapEntry& operator=(const DefaultPagemapEntry& other)
+    DefaultPagemapEntryT& operator=(const DefaultPagemapEntryT& other)
     {
-      FrontendMetaEntry<FrontendSlabMetadata>::operator=(other);
+      FrontendMetaEntry<SlabMetadata>::operator=(other);
       return *this;
     }
 
     /**
      * Default constructor.  This must be callable from the pagemap.
      */
-    SNMALLOC_FAST_PATH DefaultPagemapEntry() = default;
+    SNMALLOC_FAST_PATH DefaultPagemapEntryT() = default;
   };
+
+  using DefaultPagemapEntry = DefaultPagemapEntryT<FrontendSlabMetadata>;
+
 } // namespace snmalloc
--- a/src/test/func/cheri/cheri.cc
+++ b/src/test/func/cheri/cheri.cc
@@ -1,0 +1,153 @@
+#include <iostream>
+
+#if defined(SNMALLOC_PASS_THROUGH) || !defined(__CHERI_PURE_CAPABILITY__)
+// This test does not make sense in pass-through or w/o CHERI
+int main()
+{
+  return 0;
+}
+#else
+
+// #  define SNMALLOC_TRACING
+
+#  include <cheri/cherireg.h>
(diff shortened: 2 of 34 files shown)


nwf-msr:
FWIW, I also have a version of this sitting atop #535.

mjp41:
@nwf-msr so I think this is good.  But I want to understand the precise design and threat model

* We need to bound any allocation that we provide to the user.
* We need to be able to consolidate allocations that have been given to the user back into larger allocations.
* CHERI does not provide a consolidation mechanism, so we have to stash the original mmap permission and use that when we combine things.

I believe these are the core design constraints.  These force the expanding the SlabMetadata and stashing the original capability in it. 

Now there is an additional constraint that I am unsure if we want or need

* Meta data should have reduced bounds.

Now, if we don't require meta-data to have reduced bounds, then we do not actually need to modify SmallBuddy or introduce the ProvenanceCaptureRange.  The backend and meta data are all bounded to the original mmap range, and only the allocations going out to the user are bounded. 

So as a principle of least required priviledge, bounding seems good, but it is unclear to me if it actually provides any additional safety guarantees.  For instance, bounding the `FrontendMetadata` to its size, would mean that it doesn't give access to anything else.  However, it contains a field with a permission to a lot of memory.  This field is used infrequently, so could be "sealed", where as the metadata pointer is used on a fast path, so you might not want to seal it.

What is the principle you are driving for?  Only infrequently accessed pointers have a large capability?

In seems the smallest change to get CHERI working is less than this PR?

nwf-msr:
> @nwf-msr so I think this is good. But I want to understand the precise design and threat model
> 
> * We need to bound any allocation that we provide to the user.
> * We need to be able to consolidate allocations that have been given to the user back into larger allocations.
> * CHERI does not provide a consolidation mechanism, so we have to stash the original mmap permission and use that when we combine things.
> 
> I believe these are the core design constraints.

I think that's right.

> These force the expanding the SlabMetadata and stashing the original capability in it.

 Of the varieties of amplification we've tried, I think the "larger SlabMetadata" is the least bad we've come up with, but I'd not characterize that particular data structure decision as forced, exactly.  (I keep hoping for a less-redundant encoding, but that's probably at odds with keeping the number of cache-lines loaded per `free` down.  So it goes.)

> Now there is an additional constraint that I am unsure if we want or need
> 
> * Meta data should have reduced bounds.
> 
> Now, if we don't require meta-data to have reduced bounds, then we do not actually need to modify SmallBuddy or introduce the ProvenanceCaptureRange. The backend and meta data are all bounded to the original mmap range, and only the allocations going out to the user are bounded.

If metadata were confined to the (eventual) allocator compartment, that's quite plausible.  Right now, though, metadata includes not only the `SlabMetadata` but also the `Allocator` objects that the user comes to hold (https://github.com/microsoft/snmalloc/blob/9464556129ad38ed6cb19c472981576a93f529d4/src/snmalloc/mem/pool.h#L133).  So, I think, in fact, this PR doesn't quite go far enough at the moment, in that those are not bounded to their size.

However, since we envision the compartmentalized design, in which `Allocator` pointers held by the user are sealed, that might argue that we want to bound metadata not in `alloc_meta_data` or the small-buddy range itself but only in callers that actually ship objects out to the user.  We could have the small-buddy range understand two different kinds of `dealloc` based on the bounds of the object being given back.  That might more-tightly couple `SmallBuddyRange` and `ProvenanceCaptureRange`, but we might be able to keep amplification as a (potentially) more generic call to the parent range that only some offer?  Needs more thought, I think.

mjp41 (approved):
I am happy with this in its current state.  When this lands can we enable CHERI CI?

rmn30:
> It'd also be good to check that none of this runs afoul of @rmn30's work with CHERI+MTE; ideally it should just be a bit of spelling changes to use this for his work, but if it turns out I've made something harder or impossible by mistake I wouldn't be hugely surprised.

All looks good so far. I'll find out once I try a rebase / merge.

nwf-msr:
Rebased and added two things to the test harness (the last two commits): a new bit in `func/malloc` to check for `VMEM` and a new `func/cheri` test for some narrow, CHERI-specific behaviors (verifying that we zero large objects by `mmap` correctly, implying that we internally amplified correctly) and that the pooled, metadata `CoreAlloc` objects are indeed bounded and lack `VMEM` authority.

src/snmalloc/backend_helpers/defaultpagemapentry.h
```diff
@@ -43,22 +44,96 @@ namespace snmalloc
...
+      typename A4>
+    friend class BackendAllocator;
+
+    capptr::Arena<void> arena;
+
+    /* Set the arena pointer */
+    void arena_set(capptr::Arena<void> a)
+    {
+      arena = a;
+    }
```
davidchisnall, line 94: Why does this exist, given that it updates a field that has the same visibility as the set method?

Edit: I see that it's necessary to match the interface shared with the Lax version.  Can we have a concept that encapsulates this and some comments explaining why the method exists here?

src/test/func/cheri/cheri.cc
```diff
@@ -0,0 +1,153 @@
...
+    size_t pagesize[8];
+    int err = getpagesizes(pagesize, sizeof(pagesize) / sizeof(pagesize[0]));
+    SNMALLOC_CHECK(err > 0);
+    SNMALLOC_CHECK(pagesize[0] == OS_PAGE_SIZE);
+  }
+#  endif
+
+  auto alloc = get_scoped_allocator();
+
+  message<128>("Grab small object");
```
davidchisnall, line 51: You almost certainly don't need the template parameter on message.

davidchisnall:
It would be easier to review if the change from the `CapPtr` constructor to `unsafe_from` could be a separate NFC PR.  There's a lot of code churn in this that hides the important bits.

nwf-msr:
> It would be easier to review if the change from the `CapPtr` constructor to `unsafe_from` could be a separate NFC PR. There's a lot of code churn in this that hides the important bits.

I'm happy to split out any subset of the commits, sure, and I should perhaps have noted up front that this PR is structured to be reviewed commit-by-commit rather than all at once.

src/snmalloc/backend_helpers/defaultpagemapentry.h
rmn30: I suppose the arena field could be private?

rmn30 (approved):

src/snmalloc/backend_helpers/defaultpagemapentry.h
nwf-msr: It's a `class`, so it's all private, and the `Backend` is `friend`-ed to allow access.  ("Ah, C++.")

I've moved this out to `src/snmalloc/backend_helpers/cheri_slabmetadata_mixin.h` and added a `concept`.  See https://github.com/microsoft/snmalloc/pull/537/commits/bdb64a25332a1cea54b9eb2fcabc1891bd3c2c08 (which replaces https://github.com/microsoft/snmalloc/commit/72b47b207def5bc12cd39a2b796c28044797ae10).

nwf-msr:
Bad timing on my part; sorry @rmn30!

The changes since what you reviewed are mostly confined to https://github.com/microsoft/snmalloc/pull/537/commits/bdb64a25332a1cea54b9eb2fcabc1891bd3c2c08 but, in full, are https://github.com/microsoft/snmalloc/compare/af2adf3..ed88792

PR #540, opened 2022-06-07
Separate locking from static range

mjp41:
Pull a part the locking from the static range. This enables
locking to be added to a range directly, and does not require
it to be made static.  This is useful in cases where the source
of memory is shared between threads, but not static.  I.e. there
are multiple instances of the same type.

### DIFF
--- a/src/snmalloc/backend_helpers/lockrange.h
+++ b/src/snmalloc/backend_helpers/lockrange.h
@@ -1,0 +1,46 @@
+#pragma once
+
+#include "../ds/ds.h"
+#include "empty_range.h"
+
+namespace snmalloc
+{
+  /**
+   * Protect the ParentRange with a spin lock.
+   */
+  struct LockRange
+  {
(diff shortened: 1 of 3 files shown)


mjp41:
This isn't actually required by the process_sandbox usage.  I think it is still nice to pull apart, so I have left the PR up.  But it can be dropped.

src/snmalloc/backend_helpers/lockrange.h
```diff
@@ -0,0 +1,44 @@
+#pragma once
+
+#include "../ds/ds.h"
+#include "empty_range.h"
+
+namespace snmalloc
+{
+  /**
+   * Protect the ParentRange with a spin lock.
```
nwf-msr, line 9: It might be worth a comment that this is intended to have a `StaticRange` as a transitive child, otherwise we'll end up with a whole lot of uncontended locks.

nwf-msr (approved):

PR #541, opened 2022-06-09
memcpy vs StrictProvenance

nwf-msr:
More Lincoln Logs-style work, mostly: assemble existing pieces differently for a `StrictProvenance` / CHERI-aware `memcpy`.  As documented, there are two constraints in tension:

- We must use capability load and store instructions wherever possible, so that our `memcpy` obliviously copies capabilities (and preserves tags) during its operation.
- We cannot use capability load or store instructions at misaligned addresses, since those (are permitted to) trap.

Therefore, we have three cases to consider:
1. copies too small to move a capability: use a jump table of _data word_ copies and be done with it
2. "equally (mis)aligned" source and targets: these may have a (naturally aligned) capability inside the source that would also land at a naturally aligned destination address: perform _data word_ copies to get us up to `alignof(void*)`, perform _capability_ copies, and then go back to doing _data word_ copies to get us up to the end.
3. differently aligned source and targets: since at least one of these must be misaligned relative to capabilities, use exclusively _data word_ copies, which will result in the destination having all capability tags clear.

The only part of this that isn't (relatively) straightforward assembly of existing pieces is the change to `copy_one` to not use `__builtin_memcpy_inline` on `StrictProvenance` architectures to work around (or work with?) https://github.com/CTSRD-CHERI/llvm-project/issues/623 .

### DIFF
--- a/src/snmalloc/global/memcpy.h
+++ b/src/snmalloc/global/memcpy.h
@@ -159,8 +159,7 @@ namespace snmalloc
       std::max(sizeof(uint64_t), sizeof(void*));
 
     /**
-     * Hook for architecture-specific optimisations.  Does nothing in the
-     * default case.
+     * Hook for architecture-specific optimisations.
      */
     static SNMALLOC_FAST_PATH_INLINE void
     copy(void* dst, const void* src, size_t len)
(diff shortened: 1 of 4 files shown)


src/snmalloc/global/memcpy.h
```diff
@@ -12,6 +12,17 @@ namespace snmalloc
...
 #if __has_builtin(__builtin_memcpy_inline)
+    if constexpr (aal_supports<StrictProvenance>)
+    {
+      static_assert(Size <= sizeof(void*));
+      if constexpr (Size == sizeof(void*))
+      {
+        auto dp = static_cast<void**>(dst);
+        auto sp = static_cast<void* const*>(src);
+        *dp = *sp;
+        return;
```
davidchisnall, line 23: I don't think that this is correct, for two reasons:

 - It will not to capability-sized copies for larger-than-a-capability things.
 - It assumes capability alignment, which may not be the case.

src/snmalloc/global/memcpy.h
```diff
@@ -179,6 +189,60 @@ namespace snmalloc
...
+  {
+    static SNMALLOC_FAST_PATH_INLINE void
+    copy(void* dst, const void* src, size_t len)
+    {
+      static_assert(bits::is_pow2(sizeof(void*)));
+
+      // If this is a small size, use a jump table
+      if (len < sizeof(void*))
+      {
+        small_copies<sizeof(void*), sizeof(long)>(dst, src, len);
```
davidchisnall, line 209:
```suggestion
        small_copies<sizeof(void*)-1, sizeof(long)>(dst, src, len);
```
I missed the `<` above and so was confused about why this works.

src/snmalloc/global/memcpy.h
```diff
@@ -179,6 +189,60 @@ namespace snmalloc
...
+        len = len & (sizeof(void*) - 1);
+
+        // Use a jump table at the end, too.  If we did the copy_end
+        // overlapping store trick, we'd risk damaging the capability in
+        // the cell behind us.
+        small_copies<sizeof(void*), sizeof(long)>(dst, src, len);
+      }
+      // Otherwise, we cannot use pointer-width operations because one of
+      // the load or store is going to be misaligned.  So, same dance, but
+      // with integer registers only.
```
davidchisnall, line 237: Why not do a vectorised `memcpy` here?  On Morello, doing a Neon 128-bit memcpy is probably faster than doing a 64-bit one.

src/snmalloc/global/memcpy.h
```diff
@@ -179,6 +189,60 @@ namespace snmalloc
...
+      // If this is a small size, use a jump table
+      if (len < sizeof(void*))
+      {
+        small_copies<sizeof(void*), sizeof(long)>(dst, src, len);
+      }
+      // Equally-misaligned segments could be holding pointers internally,
+      // so perform unaligned operations at the top and bottom of the range.
+      else if (
+        (address_cast(src) & (sizeof(void*) - 1)) ==
+        (address_cast(dst) & (sizeof(void*) - 1)))
```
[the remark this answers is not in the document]
nwf-msr: I didn't see a `pointer_misalignment` helper floating around, but I can add one.

nwf-msr:
Take two.  This leaves the existing primitives alone and instead just directly loops in the `GenericStrictProvenance` architecture support.

Running `perf-memcpy-fast` on my Morello box yields:
snmalloc-memcpy-cheri

nwf-msr:
With a little bit of cleverness from @mjp41, this version looks to do even better.  While here, I've expanded the `test/func/cheri` to probe all possible equal and unequal relative misalignments.

snmalloc-memcpy-cheri

nwf-msr:
OK, I believe I have fixed my arithmetic error(s), significantly increased test coverage in `func/cheri`, and have appeased MSVC by chanting a bit of `static_cast` between signed and unsigned types.  This version has performance much like the earlier draft, I'm afraid:
snmalloc-memcpy-cheri

src/snmalloc/global/memcpy.h
```diff
@@ -179,6 +178,142 @@ namespace snmalloc
...
+        (static_cast<size_t>(-static_cast<ptrdiff_t>(src_misalign)) &
+         (alignof(void*) - 1));
+
+      /*
+       * If this is a small size, use a jump table
+       */
+      if (len < sizeof(void*))
+      {
+        small_copies<sizeof(void*) - 1, sizeof(long)>(dst, src, len);
+      }
```
mjp41, line 223: Could you do a larger jump table here?
```suggestion
      if (len < min_contains_aligned_ptr_size)
      {
        small_copies<sizeof(void*) - 1, sizeof(long)>(dst, src, len);
      }
```

src/snmalloc/global/memcpy.h
```diff
@@ -179,6 +178,142 @@ namespace snmalloc
...
+      /*
+       * As a function of misalignment relative to pointers, how big do we need
+       * to be such that the span could contain an aligned pointer?  We'd need
+       * to be big enough to contain the pointer and would need an additional
+       * however many bytes it would take to get us up to alignment.  That is,
+       * (sizeof(void*) - src_misalign) except in the case that src_misalign is
+       * 0, when the answer is 0, which we can get with some bit-twiddling.
+       */
+      size_t min_contains_aligned_ptr_size = sizeof(void*) +
+        (static_cast<size_t>(-static_cast<ptrdiff_t>(src_misalign)) &
```
mjp41, line 214:
```suggestion
        (static_cast<size_t>(-static_cast<ptrdiff_t>(address_cast(src))) &
```

nwf-msr:
And here's (7bcaf48) with @mjp41's feedback.
snmalloc-memcpy-cheri

davidchisnall (approved):
LGTM.  Do you have a comparison on Morello of the two CHERI/non-CHERI memcpy versions?

nwf-msr:
Here's `sn` and FreeBSD's `je` in their `cheri` purecap and `aarch64` builds:
snmalloc-memcpy

nwf-msr:
Ahem.  Sorry, now with the correct labels.  Here's `snmalloc`'s `memcpy` vs. CheriBSD's `memcpy` in both CHERI `purecap` and `aarch64` builds running on Morello:
snmalloc-memcpy

mjp41 (approved):
LGTM

PR #542, opened 2022-06-09
NFC: cmake: add SNMALLOC_LINK_ICF, default on

nwf-msr:
ICF currently breaks building on Morello, so allow cmake to notch it out.


davidchisnall (approved):
Thanks!  I've had to patch the ICF thing out for testing some combinations of compiler / linker as well, so that's going to be very useful.

PR #543, opened 2022-06-10
Add buffer append method for {,u}intptr_t

nwf-msr:
Otherwise, on platforms for which {,u}intptr_t aren't just typedef-s of
other scalar types, it's ambiguous which way an implicit cast should go.

### DIFF
--- a/src/snmalloc/ds_core/helpers.h
+++ b/src/snmalloc/ds_core/helpers.h
@@ -279,6 +279,30 @@ namespace snmalloc
       }
     }
 
+    /*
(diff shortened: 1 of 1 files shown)


nwf-msr:
With this and the `SNMALLOC_LINK_ICF=OFF` hack, all four of {debug, release} x {hybrid, purecap} build in Morello CI.

davidchisnall (commented):
I have a similar fix locally to this that casts to `void*` in both cases.  This is probably closer to the right thing because then we can print the full capability in the `void*` case.

mjp41 (approved):
LGTM

PR #545, opened 2022-06-16
Support older linux systems

### LINKED ISSUE #544: Missing defines on systems with kernel < 4.X
## Description
We are running into trouble with centos 7 based builds of snmalloc. ( [Dockerimage](https://hub.docker.com/layers/centos/library/centos/centos7.9.2009/images/sha256-8faead07bd1d5fdd17414a1759eae004da5daa9575a846f4730f44dec0f47843) )

We do have to support such an old system for the time being (don't ask), you might not want to support it, which i totally understand.
The system is based on gcc 9, kernel 3.10.0 and libraries, which is all horribly outdated.

Our build system gives the following error:

```
...
 /cargo/registry/src/github.com-1ecc6299db9ec823/snmalloc-sys-0.3.1/snmalloc/src/snmalloc/override/../global/../backend/../backend_helpers/../mem/../ds/../pal/pal_linux.h: In static member function 'static void snmalloc::PALLinux::notify_not_using(void*, size_t)':
  /cargo/registry/src/github.com-1ecc6299db9ec823/snmalloc-sys-0.3.1/snmalloc/src/snmalloc/override/../global/../backend/../backend_helpers/../mem/../ds/../pal/pal_linux.h:111:24: error: 'MADV_FREE' was not declared in this scope; did you mean 'MADV_REMOVE'?
    111 |       madvise(p, size, MADV_FREE);
        |                        ^~~~~~~~~
        |                        MADV_REMOVE
  /cargo/registry/src/github.com-1ecc6299db9ec823/snmalloc-sys-0.3.1/snmalloc/src/snmalloc/override/../global/../backend/../backend_helpers/../mem/../ds/../pal/pal_linux.h: In static member function 'static uint64_t snmalloc::PALLinux::get_entropy64()':
  /cargo/registry/src/github.com-1ecc6299db9ec823/snmalloc-sys-0.3.1/snmalloc/src/snmalloc/override/../global/../backend/../backend_helpers/../mem/../ds/../pal/pal_linux.h:180:57: error: 'GRND_NONBLOCK' was not declared in this scope; did you mean 'O_NONBLOCK'?
    180 |           ret = syscall(SYS_getrandom, current, length, GRND_NONBLOCK);
        |                                                         ^~~~~~~~~~~~~
        |                                                         O_NONBLOCK
  gmake[3]: *** [CMakeFiles/snmallocshim-rust.dir/src/snmalloc/override/rust.cc.o] Error 1
```

## Possible Solutions

On those old systems we do have `GRND_NONBLOCK` etc. defined in `/usr/include/linux/random.h`, i do think this include is missing (maybe guarded by a check if `GRND_NONBLOCK` is defined).

`MADV_FREE` is only available since linux 4.5. Should we fallback to `MADV_REMOVE` for systems where `MADV_FREE` is not defined?

I will happily provide those fixes if you deem them good enough or maybe you have another idea?

mfelsche:
The particular system to support here is centos 7 with 

* kernel-headers for kernel 3.10.0
* glibc 2.17
Issued with that platform:

* `MADV_FREE` undefined
* `GRND_NONBLOCK` not defined as there is no `sys/random.h`

Fixes #544

### DIFF
--- a/src/snmalloc/pal/pal_linux.h
+++ b/src/snmalloc/pal/pal_linux.h
@@ -40,6 +40,19 @@ namespace snmalloc
      */
     static constexpr int default_mmap_flags = MAP_NORESERVE;
 
+    /**
+     * MADV_FREE is only available since Linux 4.5.
+     *
+     * Fallback to MADV_DONTNEED on older kernels
+     */
+    static constexpr int madvise_free_flags =
+#  ifdef SNMALLOC_HAS_LINUX_RANDOM_H
+      MADV_FREE
+#  else
+      MADV_DONTNEED
+#  endif
+      ;
+
     static void* reserve(size_t size) noexcept
     {
       void* p = PALPOSIX<PALLinux>::reserve(size);
@@ -113,15 +126,7 @@ namespace snmalloc
         memset(p, 0x5a, size);
 
       madvise(p, size, MADV_DONTDUMP);
-      madvise(
-        p,
-        size,
-#  if defined(MADV_FREE)
-        MADV_FREE
-#  else
-        MADV_DONTNEED
-#  endif
-      );
+      madvise(p, size, madvise_free_flags);
 
       if constexpr (PalEnforceAccess)
       {
--- a/src/snmalloc/pal/pal_linux.h
+++ b/src/snmalloc/pal/pal_linux.h
@@ -8,7 +8,9 @@
 #  include <sys/mman.h>
 #  include <sys/prctl.h>
 #  include <syscall.h>
-#  if !defined(GRND_NONBLOCK) && __has_include(<linux/random.h>)
+// __has_include does not reliably determine if we actually have linux/random.h
+// available
+#  if defined(SNMALLOC_HAS_LINUX_RANDOM_H)
 #    include <linux/random.h>
 #  endif
 
--- a/src/snmalloc/pal/pal_linux.h
+++ b/src/snmalloc/pal/pal_linux.h
@@ -8,6 +8,9 @@
 #  include <sys/mman.h>
 #  include <sys/prctl.h>
 #  include <syscall.h>
+#  if !defined(GRND_NONBLOCK) && __has_include(<linux/random.h>)
+#    include <linux/random.h>
+#  endif
 
 extern "C" int puts(const char* str);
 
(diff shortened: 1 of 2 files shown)


mfelsche:
Example docker image to reproduce the issue on (just in case) (The underlying kernel doesn't actually matter here):

```
FROM centos:7.9.2009

# install extra software repos
RUN yum install --assumeyes \
    centos-release-scl `# provides devtoolset-9*` \
    epel-release       `# provides deps like cmake3 and clang`

# get gcc9 via devtoolset-9, since gcc4 available from centos7 repos does not
# https://access.redhat.com/documentation/en-us/red_hat_developer_toolset/9/html/9.0_release_notes/dts9.0_release
RUN yum install --assumeyes \
    devtoolset-9-gcc \
    devtoolset-9-gcc-c++ \
    make \
    cmake3 \
    git \
    devtoolset-9-libatomic-devel
```

src/snmalloc/pal/pal_linux.h
```diff
@@ -8,6 +8,11 @@
 #  include <sys/mman.h>
 #  include <sys/prctl.h>
 #  include <syscall.h>
+// __has_include does not reliably determine if we actually have linux/random.h
+// available
+#  if defined(SNMALLOC_HAS_LINUX_RANDOM_H)
```
nwf-msr, line 13: What's going on with `__has_include`?

It might also be worth a comment here to point out that below we check for `SYS_getrandom`, which won't be defined if this file isn't included?

src/snmalloc/pal/pal_linux.h
```diff
@@ -108,7 +113,15 @@ namespace snmalloc
...
       madvise(p, size, MADV_DONTDUMP);
-      madvise(p, size, MADV_FREE);
+      madvise(
+        p,
+        size,
+#  if defined(MADV_FREE)
+        MADV_FREE
+#  else
+        MADV_DONTNEED
+#  endif
```
nwf-msr, line 123: I might have a very slight stylistic preference for this being made a `madvise_free_flags` `constexpr` or something rather than C preprocessor pragmas "inside" a function call.

src/snmalloc/pal/pal_linux.h
mfelsche: I honestly cannot tell you what is going on with `__has_include` other than with gcc 9.3 is wasn't picking up the following:

```c
#if __has_include(<linux/random.h>)
#   include <linux/random.h>
#endif
```

I honestly have no idea why.

mfelsche: `SYS_getrandom` seems to come from `<syscall.h>`

mfelsche: But overall if both `sys/random.h` and `linux/random.h` aren't available, the build will fail.

mjp41:
@mfelsche do you need the Rust package updating for this too?

mfelsche:
Thanks for your quick reviews and merging! Yeah, we do need the rust package be updated too, actually!

PR #546, opened 2022-06-17
Fix wrong ifdef in pal_linux.h

mfelsche:
Horribly sorry, i fatfingered the wrong ifdef around the missing `MADV_FREE` flag on older linux systems (kernel-headers < 4.5).


mjp41 (approved):
LGTM.

PR #548, opened 2022-08-05
annotate `alloc` on unixes.

devnexen:


src/snmalloc/mem/localalloc.h
```diff
@@ -1,7 +1,9 @@
 #pragma once
 
-#ifdef _MSC_VER
-#  define ALLOCATOR __declspec(allocator)
+#if defined(_MSC_VER)
+#  define ALLOCATOR __declspec(allocator) __declspec(restrict)
+#elif defined(__GNUC__)
```
nwf, line 5: How about
```suggestion
#elif __has_attribute(malloc)
```
since I think both clang and gcc support that (though if CI says that we're supporting versions too old, well, so be it, I guess).

src/snmalloc/mem/localalloc.h
```diff
@@ -1,7 +1,9 @@
...
+#  define ALLOCATOR __attribute__((__malloc__))
```
nwf, line 6: Shouldn't that be
```suggestion
#  define ALLOCATOR __attribute__((malloc))
```
and probably we want to do something like
```suggestion
#  define ALLOCATOR __attribute__((malloc, malloc(dealloc)))
```
to associate the deallocator, but perhaps that means it has to be 
```
#  define ALLOCATOR(dealloc) __attribute__((malloc, malloc(dealloc)))
```

devnexen: > #  define ALLOCATOR(dealloc) __attribute__((malloc, malloc(dealloc)))

I thought of that but clang does not support this form.

src/snmalloc/mem/localalloc.h
devnexen: well it seems only clang and gcc support this anyway https://www.intel.com/content/www/us/en/develop/documentation/cpp-compiler-developer-guide-and-reference/top/compiler-reference/attributes.html seems i changes nothing in this case but I can always try.

mjp41 (approved):
LGTM, @nwf-msr are you happy with the updates?

src/snmalloc/mem/localalloc.h
davidchisnall: This form is somewhat useful for static analysis but it likely to lead to miscompiles as currently proposed to be integrated in LLVM.  The problem is that the compiler doesn't track the hierarchy of allocators that are in use.  An allocation path might have a call to alloc, whereas the deallocation path might end up being inlined to the point where it knows to call `munmap` or similar.  I think that isn't a problem for the most recent snmalloc (we never hand actually unmap memory) but it might be a problem if we add the same attributes for the ranges (if you do a large thread-local allocation and free then we'll still hit the allocator path but may push it back into the thread's large chunk cache on deallocation and so enough might be inlined to confuse the optimiser into thinking that you've allocated with one allocator and freed with another).

nwf-msr:
> LGTM, @nwf-msr are you happy with the updates?

LGTM

devnexen:
ping :)

mjp41:
> ping :)

Sorry, I forgot about this one.

PR #550, opened 2022-08-10
Add stricter double free checking on large allocs

mjp41:
Although, double frees are generally caught on large allocs.  This is
done very late in the process after many operations more operations have
occurred.

This change brings that check much earlier in the process.


nwf-msr (approved):
LGTM

PR #552, opened 2022-08-12
In Checked mode use nullptr for malloc(0)

mjp41:
The spec allows malloc(0) to return an actual allocation, or nullptr.
Some code has compatibility issues with assuming if malloc returns nullptr
then the memory is full.  Hence, we normally use malloc(0) to return an
allocation.

In the checked client it seems better to return nullptr, and I believe
we have more opportunity for clients to find errors with it.


davidchisnall:
This would mean that we couldn't use the checked mode in FreeBSD libc, because some of the system utilities have this bug.  The problem is that `realloc` has no mechanism to report failure other than returning `NULL`, but returning `NULL` is success for `realloc` with size 0.

PR #556, opened 2022-09-10
Make smallbuddy handle larger requests correctly

mjp41:
Verona was using a pool allocator to requests large pieces of meta-data. This was broken with the current design, so the smallbuddy can now pass requests up if they are too large.

### DIFF
--- a/src/snmalloc/backend_helpers/smallbuddyrange.h
+++ b/src/snmalloc/backend_helpers/smallbuddyrange.h
@@ -174,13 +174,22 @@ namespace snmalloc
           base,
           length,
           [this](CapPtr<void, ChunkBounds> base, size_t align, bool) {
-            CapPtr<void, ChunkBounds> overflow =
-              buddy_small
-                .add_block(
-                  base.template as_reinterpret<FreeChunk<ChunkBounds>>(), align)
-                .template as_reinterpret<void>();
-            if (overflow != nullptr)
-              parent.dealloc_range(overflow, bits::one_at_bit(MIN_CHUNK_BITS));
+            if (align < MIN_CHUNK_SIZE)
+            {
+              CapPtr<void, ChunkBounds> overflow =
+                buddy_small
+                  .add_block(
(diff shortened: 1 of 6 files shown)


src/snmalloc/backend_helpers/smallbuddyrange.h
```diff
@@ -204,7 +213,8 @@ namespace snmalloc
 
       CapPtr<void, ChunkBounds> alloc_range(size_t size)
       {
-        SNMALLOC_ASSERT(size < MIN_CHUNK_SIZE);
+        if (size >= MIN_CHUNK_SIZE)
+          return parent.alloc_range(size);
```
nwf, line 217: I recall us going back and forth on this previously, with the idea that the caller should skip the `SmallBuddy` ranges if they needed something larger.  This fallback requires that the return type of the `SmallBuddy` match that of its parent, which might not, in general, be true: while a `LargeBuddy` "must" operate on `CapPtr`s intended to be large enough to include whole `PageMap` `MetaEntry`s worth of space, as it uses those for its metadata, the `SmallBuddy`, which uses in-band metadata, could have more tightly bounded types.

I suppose we're not actually taking advantage of that right now, tho'... but is there reason not to ask for the Range's `parent` in the caller instead?

mjp41: So ultimately we need a branch somewhere in `alloc_meta_data`.  This would need to determine how large the request is and then which range to send it too.  The code in `SmallBuddyRange::alloc_with_leftover`, returns the slop at the end of the allocation.  This could be useful for large meta data allocations as well as small, so might want to be duplicated, though not essential.

mjp41: Following Teams call, I have added a test to ensure future changes don't break the Verona use case.

We will leave this design as is, but will revisit after @nwf-msr investigates a fine granularity of permissions.

nwf-msr (approved):
LGTM

PR #557, opened 2022-09-13
Introduce AAL hook for CHERI representable length rounding

nwf-msr:
While here, add a test that our sizeclasses are fixed points of the rounding function.

This matters now that some bigger things are to be sourced from the metadata Range; see #556.


mjp41 (approved):
LGTM

PR #558, opened 2022-09-16
Pagemap Rounding

mjp41:
This PR ensures that if a pagemap is used in a fixed region that has very large granularity it does require undue additional alignment.

### DIFF
--- a/src/snmalloc/backend_helpers/pagemap.h
+++ b/src/snmalloc/backend_helpers/pagemap.h
@@ -117,31 +117,35 @@ namespace snmalloc
 #endif
       SNMALLOC_ASSERT(s != 0);
       // TODO take account of pagemap size in the calculation of how big it
-      // needs to be.
+      // needs to be.  The following code creates a pagemap that covers the
+      // pagemap as well as the left over. This is not ideal, and we should
+      // really calculate the division with 
+      //
+      //  bits::one_at_bit(GRANULARITY_BITS) + sizeof(T)
+      //
+      // There are awkward corner cases for the alignment of the start and
+      // the end that are hard to calculate. So this is not currently done. 
 
-      // Align the start and end.  We won't store for the very ends as they
-      // are not aligned to a chunk boundary.
-      auto heap_base = pointer_align_up(b, bits::one_at_bit(GRANULARITY_BITS));
-      auto end = pointer_align_down(
+      // Calculate range in pagemap that is associated to this space.
+      // Over calculate to cover any unaligned parts at either end.
+      auto b_align = pointer_align_down(b, bits::one_at_bit(GRANULARITY_BITS));
+      auto end = pointer_align_up(
         pointer_offset(b, s), bits::one_at_bit(GRANULARITY_BITS));
-      size = pointer_diff(heap_base, end);
 
-      // Put pagemap at start of range.
-      // TODO CHERI capability bound here!
+      // Setup the pagemap.
+      base = address_cast(b_align);
+      size = pointer_diff(b_align, end);
       body = static_cast<T*>(b);
       body_opt = body;
-      // Advance by size of pagemap.
-      // Note that base needs to be aligned to GRANULARITY for the rest of the
-      // code to work
-      // TODO CHERI capability bound here!
-      heap_base = pointer_align_up(
-        pointer_offset(b, (size >> SHIFT) * sizeof(T)),
-        bits::one_at_bit(GRANULARITY_BITS));
-      base = address_cast(heap_base);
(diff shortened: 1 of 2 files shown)


src/snmalloc/backend_helpers/pagemap.h
```diff
@@ -117,31 +117,35 @@ namespace snmalloc
...
+      // There are awkward corner cases for the alignment of the start and
+      // the end that are hard to calculate. So this is not currently done.
 
-      // Align the start and end.  We won't store for the very ends as they
-      // are not aligned to a chunk boundary.
-      auto heap_base = pointer_align_up(b, bits::one_at_bit(GRANULARITY_BITS));
-      auto end = pointer_align_down(
+      // Calculate range in pagemap that is associated to this space.
+      // Over calculate to cover any unaligned parts at either end.
+      auto b_align = pointer_align_down(b, bits::one_at_bit(GRANULARITY_BITS));
```
nwf-msr, line 131: While this might take `b_align` below `b`, and so logically start our `PageMap` out of bounds, that's OK because we never access the `MetaEntry`-s associated with the `PageMap` itself?

src/snmalloc/backend_helpers/pagemap.h
```diff
@@ -117,31 +117,35 @@ namespace snmalloc
...
-      heap_base = pointer_align_up(
-        pointer_offset(b, (size >> SHIFT) * sizeof(T)),
-        bits::one_at_bit(GRANULARITY_BITS));
-      base = address_cast(heap_base);
-      SNMALLOC_ASSERT(
-        base == bits::align_up(base, bits::one_at_bit(GRANULARITY_BITS)));
+      auto heap_base = pointer_offset(b, pagemap_size);
 
-      return {heap_base, pointer_diff(heap_base, end)};
+      return {heap_base, s - pagemap_size};
```
nwf-msr, line 148: `s - pagemap_size` might be negative in pathological cases.  Consider `sizeof(T) == 1 << SHIFT` and `s = sizeof(T)` and `b & (sizeof(T) - 1) != 0`: we will have rounded `b_align` down (to the nearest `SHIFT == GRANULARITY_BITS` boundary) and `end` up so that `size = end - b_align` is `2 * (1 << SHIFT)`.

A `static_assert` that `sizeof(T) < (1 << SHIFT)` should be enough to rule that pathology out?

src/test/func/pagemap/pagemap.cc
```diff
@@ -99,6 +101,20 @@ void test_pagemap(bool bounded)
 
   // Check pattern is correctly stored
   std::cout << std::endl;
+
+  if (bounded)
+  {
+    // Check we have not corrupted the heap.
+    for (address_t ptr = low; ptr < high; ptr++)
+    {
+      auto* p = (char*)ptr;
```
nwf-msr, line 110: CHERI CI strikes again here, too; perhaps `low` wants to be a `uintptr_t` rather than an `address_t`?

nwf-msr (changes requested):

src/snmalloc/backend_helpers/pagemap.h
ihaller: Is b_align only used in address and size computations? If yes, then would it be clearer and safer to get the address of b and align the address down? That way no wild pointer exists.

src/snmalloc/backend_helpers/pagemap.h
```diff
@@ -117,31 +117,43 @@ namespace snmalloc
...
+      // the end that are hard to calculate. So this is not currently done.
 
-      // Align the start and end.  We won't store for the very ends as they
-      // are not aligned to a chunk boundary.
-      auto heap_base = pointer_align_up(b, bits::one_at_bit(GRANULARITY_BITS));
-      auto end = pointer_align_down(
+      // Calculate range in pagemap that is associated to this space.
+      // Over calculate to cover any unaligned parts at either end.
+      auto b_align = pointer_align_down(b, bits::one_at_bit(GRANULARITY_BITS));
+      auto end = pointer_align_up(
```
ihaller, line 132: Similarly end only seems to be used as an address to we can avoid the potentially out of bounds pointer floating around.

src/snmalloc/backend_helpers/pagemap.h
mjp41: So I have tried to make this capability less code as it is just addresses like you suggest:
https://github.com/microsoft/snmalloc/pull/558/commits/9542fbe931e99ec934978529142788dd0cd61c02

ihaller:
I am very happy with this change now.

PR #561, opened 2022-09-20
Add a multiple region example.

mjp41:


mjp41:
@aghosn does this example illustrate what you are after?

src/test/func/multi_region/multi_region.cc
```diff
@@ -0,0 +1,187 @@
...
+};
+
+int main()
+{
+#ifndef SNMALLOC_PASS_THROUGH // Depends on snmalloc specific features
+  setup();
+
+  // 28 is large enough to produce a nested allocator.
+  // It is also large enough for the example to run in.
+  // For 1MiB superslabs, SUPERSLAB_BITS + 4 is not big enough for the example.
```
nwf-msr, line 139: `SUPERSLAB_BITS`!  Now there's something I haven't seen in a long time.  (Probably the analogous comment in `fixed_region.cc` should also be scheduled for deletion.)

PR #562, opened 2022-09-20
Add indirect range

mjp41:
The ranges currently either completely own the parent range, or it is a static and thus completely shared.

The indirect range allows sharing without it being global.

### DIFF
--- a/src/snmalloc/backend_helpers/indirectrange.h
+++ b/src/snmalloc/backend_helpers/indirectrange.h
@@ -1,0 +1,53 @@
+#pragma once
+
+#include "../ds/ds.h"
+#include "empty_range.h"
+
+namespace snmalloc
+{
+  /**
+   * Stores a references to the parent range so that it can be shared
+   * without `static` scope.
+   * 
+   * This could be used to allow multiple allocators on a single region of memory.
+   */
+  struct IndirectRange
+  {
+    template<typename ParentRange = EmptyRange<>>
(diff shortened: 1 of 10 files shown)


src/snmalloc/backend_helpers/indirectrange.h
```diff
@@ -0,0 +1,53 @@
...
+    class Type : public RefParent<ParentRange>
+    {
+      using RefParent<ParentRange>::parent;
+
+    public:
+      static constexpr bool Aligned = ParentRange::Aligned;
+
+      static_assert(
+        ParentRange::ConcurrencySafe,
+        "StaticRange requires a concurrency safe parent.");
```
nwf-msr, line 26:
```suggestion
        "IndirectRange requires a concurrency safe parent.");
```

src/snmalloc/backend_helpers/range_helpers.h
```diff
@@ -130,4 +130,35 @@ namespace snmalloc
     }
   };
 
+  /**
+   * Helper class for allowing a range to be navigated to find an
+   * ancestor of a specific type. The parent is a pointer to a range
```
nwf-msr, line 135:
```suggestion
   * ancestor of a specific type. The parent is a pointer to a range;
```

nwf-msr (approved):
It's baa-aack.  LGTM modulo some nits.

PR #569, opened 2022-10-31
Towards heap walk

mjp41:
This includes a two changes that are on the way to implementing heap walk.  The changes performed are:

* Track the `laden` ranges of memory: the full and almost full slabs and large allocations.
* Ensure each slab metadata knows what range of memory it is describing.  This was not previously the case for a fully used slab where there is no free list.

### DIFF
--- a/src/snmalloc/aal/address.h
+++ b/src/snmalloc/aal/address.h
@@ -291,8 +291,9 @@ namespace snmalloc
   }
 
   /**
-   * Convert an address_t to a pointer.  The returned pointer should never be followed.
-   * On CHERI following this pointer will result in a capability violation.
+   * Convert an address_t to a pointer.  The returned pointer should never be
+   * followed. On CHERI following this pointer will result in a capability
+   * violation.
    */
   template<typename T>
   SNMALLOC_FAST_PATH_INLINE T* useless_ptr_from_addr(address_t p)
--- a/src/snmalloc/aal/address.h
+++ b/src/snmalloc/aal/address.h
@@ -290,4 +290,13 @@ namespace snmalloc
     return static_cast<size_t>(a - pointer_align_down<alignment>(a));
   }
 
+  /**
+   * Convert an address_t to a pointer.  The returned pointer should never be followed.
+   * On CHERI following this pointer will result in a capability violation.
+   */
+  template<typename T>
+  SNMALLOC_FAST_PATH_INLINE T* useless_ptr_from_addr(address_t p)
+  {
+    return reinterpret_cast<T*>(p);
+  }
 } // namespace snmalloc
--- a/src/snmalloc/aal/address.h
+++ b/src/snmalloc/aal/address.h
@@ -298,6 +298,6 @@ namespace snmalloc
   template<typename T>
   SNMALLOC_FAST_PATH_INLINE T* useless_ptr_from_addr(address_t p)
   {
-    return reinterpret_cast<T*>(p);
+    return reinterpret_cast<T*>(static_cast<uintptr_t>(p));
   }
 } // namespace snmalloc
--- a/src/snmalloc/mem/corealloc.h
+++ b/src/snmalloc/mem/corealloc.h
@@ -811,7 +811,7 @@ namespace snmalloc
 
       // Set meta slab to empty.
       meta->initialise(
-        sizeclass, address_cast(slab), entropy.get_free_list_key());
+        sizeclass, slab.unsafe_ptr(), entropy.get_free_list_key());
 
       // Build a free list for the slab
       alloc_new_list(slab, meta, rsize, slab_size, entropy);
--- a/src/snmalloc/mem/corealloc.h
+++ b/src/snmalloc/mem/corealloc.h
@@ -885,14 +885,13 @@ namespace snmalloc
         dealloc_local_slabs<true>(sizeclass);
       }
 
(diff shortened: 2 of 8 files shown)


SchrodingerZhu:
Will we finally reach precise heap allocation statistics?

mjp41:
> Will we finally reach precise heap allocation statistics?

This is not aimed at better statistics.  We could add that, but it is not on my current plan.  My rough thought would be to track per sizeclass

sizeclass | total allocations | curr slabs | max slabs
-------------------------------------------------------


The total allocations would be bumped everytime we grab a new free list, and curr slabs and max slabs would be increased on `alloc_chunk` and decreased on `dealloc_chunk`. This could then be done just with interlocked instructions on global state as it would be sufficiently infrequent to not cost too much.  It wouldn't track fragmentation though, which might be a useful thing to track.

mjp41:
This last commit https://github.com/microsoft/snmalloc/pull/569/commits/9042b405859858f0ef48ff7f21f1db5597013854 is my attempt to get the fix passed the CHERI compiler. We won't really be using this value, except to look up in the pagemap, but changing all the types in FreeObject to have a union with `address_t` seems painful.  @mwf-msr do you have any thoughts on what the correct approach is?

src/snmalloc/mem/corealloc.h
```diff
@@ -892,28 +921,45 @@ namespace snmalloc
...
     {
-      auto test = [&result](auto& queue, smallsizeclass_t size_class) {
-        queue.iterate([&result, size_class](auto slab_metadata) {
+      auto& key = entropy.get_free_list_key();
+
+      auto error = [&result, &key](auto slab_metadata) {
+        auto slab_interior = slab_metadata->get_slab_interior(key);
+        const PagemapEntry& entry =
+          Config::Backend::get_metaentry(slab_interior);
+        auto size_class = entry.get_sizeclass();
```
mjp41, line 930:
```suggestion
        SNMALLOC_ASSERT(slab_metadata == entry.get_metadata());
        auto size_class = entry.get_sizeclass();
```

src/snmalloc/mem/freelist.h
```diff
@@ -697,6 +687,16 @@ namespace snmalloc
...
           }
+
+          // Head is not live when a building is initialised.
+          // We use this slot to store a pointer into the slab for the
+          // allocations. This then establishes the invariant that head is
+          // always (a possibly encoded) pointer into the slab, and thus
+          // the Freelist builder always knows which block it is referring too.
+          head[i] = Object::code_next(
+            address_cast(&head[i]),
+            reinterpret_cast<Object::T<BQueue>*>(slab),
```
nwf, line 698: This might also be pronounced
```suggestion
            unsafe_from_uintptr<Object::T<BQueue>>(slab),
```
(Or maybe a `static_cast<uintptr_t>(slab)` is necessary.)  Whether that is more or less terrifying than `reinterpret_cast`, I will leave to you to decide.  I wonder if we want to wrap this up in something that makes it more obvious that it's just an address getting made into a provenance-free (and so, on CHERI, untagged) pointer?  `template<typename T> T* useless_ptr_from_addr(address_t)`?

src/snmalloc/mem/freelist.h
```diff
@@ -688,7 +678,7 @@ namespace snmalloc
       /**
        * Set the builder to a not building state.
        */
-      constexpr void init()
+      constexpr void init(void* slab, const FreeListKey& key)
```
nwf-msr, line 681: Given `useless_ptr_from_addr`, can this be
```suggestion
      constexpr void init(address_t slab, const FreeListKey& key)
```

If so, I think `extract_segment` and `initialise` and `initialise_large` (in `metadata.h`) can also take `address_t`.  I think the update propagation stops at their callers.

src/snmalloc/mem/freelist.h
mjp41: I tried this, but CHERI purecap build complained that I was creating a provenance free pointer.  Which is what I wanted to do :(

src/test/func/statistics/stats.cc
```diff
@@ -35,7 +48,46 @@ int main()
...
+void debug_check_empty_2()
+{
+  snmalloc::Alloc& a = snmalloc::ThreadAlloc::get();
+  bool result;
+  std::vector<void*> allocs;
+  size_t count = 2048;
+
+  for (size_t i = 0; i < count; i++)
+  {
+    auto r = a.alloc(128);
```
nwf-msr, line 66: Should this be
```suggestion
    auto r = a.alloc(size);
```

src/snmalloc/aal/address.h
```diff
@@ -290,4 +290,14 @@ namespace snmalloc
...
 
+  /**
+   * Convert an address_t to a pointer.  The returned pointer should never be
+   * followed. On CHERI following this pointer will result in a capability
+   * violation.
+   */
+  template<typename T>
+  SNMALLOC_FAST_PATH_INLINE T* useless_ptr_from_addr(address_t p)
+  {
+    return reinterpret_cast<T*>(p);
```
nwf-msr, line 301: Placate the CHERI compiler's twitchy provenance warning with
```suggestion
    return reinterpret_cast<T*>(static_cast<uintptr_t>(p));
```

nwf-msr (approved):
LGTM!

PR #570, opened 2022-11-02
Various changes to improve startup times for lots of threads

mjp41:
While profiling starting a lot of threads, I found a few bits that could be improved.  This PR attempts to reduce the contention on startup.  By

* Reducing work under locks
   - ~~Removing a lock from the pool implementation~~
   - Optimising the buddy allocator to only look as high as it has blocks of memory
   - Making the global meta range not perform madvise under a lock.  
* Implementing a proper MCS queue lock

### DIFF
--- a/src/snmalloc/backend/backend.h
+++ b/src/snmalloc/backend/backend.h
@@ -57,6 +57,9 @@ namespace snmalloc
           GlobalMetaRange::ConcurrencySafe,
           "Global meta data range needs to be concurrency safe.");
         GlobalMetaRange global_state;
+        // This assert is not necessary for correctness, but if
+        // this is not true, we will be wasting a lot of space.
+        SNMALLOC_ASSERT(size >= (MIN_CHUNK_SIZE / 4));
         auto rsize = bits::max(bits::next_pow2(size), MIN_CHUNK_SIZE);
         p = global_state.alloc_range(rsize);
       }
--- a/src/snmalloc/backend/backend.h
+++ b/src/snmalloc/backend/backend.h
@@ -57,7 +57,8 @@ namespace snmalloc
           GlobalMetaRange::ConcurrencySafe,
           "Global meta data range needs to be concurrency safe.");
         GlobalMetaRange global_state;
-        p = global_state.alloc_range(bits::next_pow2(size));
+        auto rsize = bits::max(bits::next_pow2(size), MIN_CHUNK_SIZE);
+        p = global_state.alloc_range(rsize);
       }
 
       if (p == nullptr)
--- a/src/snmalloc/backend/backend.h
+++ b/src/snmalloc/backend/backend.h
@@ -57,9 +57,14 @@ namespace snmalloc
           GlobalMetaRange::ConcurrencySafe,
           "Global meta data range needs to be concurrency safe.");
         GlobalMetaRange global_state;
+        // TODO - this cannot be uncommented currently due to some use cases.
+        //   We should allow the Pool to use the allocator or the backend for 
+        //   other scenarios.  E.g. the Pool is used by Verona-rt where an 
+        //   allocator has already been created. See #603
+        //   
         // This assert is not necessary for correctness, but if
         // this is not true, we will be wasting a lot of space.
-        SNMALLOC_ASSERT(size >= (MIN_CHUNK_SIZE / 16));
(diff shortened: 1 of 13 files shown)


src/snmalloc/backend_helpers/buddy.h
```diff
@@ -108,7 +112,11 @@ namespace snmalloc
 
       auto bigger = remove_block(size * 2);
       if (bigger == Rep::null)
+      {
+        SNMALLOC_ASSERT(idx == highest_used);
```
nwf-msr, line 116: It seems like we might hit this in OOM scenarios if we've used both halves of the biggest index we've ever used and can't get more?

src/snmalloc/backend/backend.h
```diff
@@ -57,7 +57,8 @@ namespace snmalloc
           GlobalMetaRange::ConcurrencySafe,
           "Global meta data range needs to be concurrency safe.");
         GlobalMetaRange global_state;
-        p = global_state.alloc_range(bits::next_pow2(size));
+        auto rsize = bits::max(bits::next_pow2(size), MIN_CHUNK_SIZE);
+        p = global_state.alloc_range(rsize);
```
nwf-msr, line 61: Don't we use `alloc_meta_data` for very tiny `Metaslab` objects?  Do they never hit this `else` branch?

nwf-msr (changes requested):

src/snmalloc/backend/backend.h
mjp41: Those cases will normally have a LocalState.  The only snmalloc use for this is allocating the allocator.

src/snmalloc/backend_helpers/buddy.h
mjp41: Thanks.  I reworked this a bit, and added some debug invariants.  It is now passing tests, and is easier to follow.

mjp41:
I have dropped the changes to the Pool implementation as they seemed to be hitting a miscompilation on CHERI.

src/snmalloc/backend_helpers/buddy.h
```diff
@@ -16,6 +16,7 @@ namespace snmalloc
   class Buddy
   {
     std::array<RBTree<Rep>, MAX_SIZE_BITS - MIN_SIZE_BITS> trees;
+    size_t empty_above = 0;
```
nwf, line 19: A comment probably couldn't go amiss, especially noting that this is above in the sense of `>=`, not just `>`, yes?

src/snmalloc/backend_helpers/buddy.h
```diff
@@ -70,8 +82,11 @@ namespace snmalloc
           size *= 2;
           addr = Rep::align_down(addr, size);
           if (size == bits::one_at_bit(MAX_SIZE_BITS))
+          {
+            invariant();
```
nwf, line 86: Sorry if this should be obvious, but... Does this actually test anything?  `empty_above` should be `idx + 1` and `trees.size()` should be `MAX_SIZE_BITS - MIN_SIZE_BITS`, which should be equal to `idx + 1`?

src/snmalloc/backend/backend.h
nwf: Maybe we can have some kind of assert that we don't hit this path for something, say, smaller than `MIN_CHUNK_SIZE / 4` or something reasonable in that neighborhood?

nwf (changes requested):

src/snmalloc/backend_helpers/buddy.h
mjp41: I wanted to ensure the invariant held on all paths out of the `add_block`, this was a sufficient set to achieve that.  So I placed it before each non-tail call return without thinking about what the invariant was.

I think you are right it trivially holds here.  However, I would like to leave it as the shape is nice.  But I'll add a comment.

src/snmalloc/backend_helpers/buddy.h
mjp41: The invariant implies `>=`, I'll add a comment.

src/snmalloc/backend/backend.h
mjp41: Makes sense.

mjp41:
I am closing this in favour of a few smaller PR with defined changes.  It is not clear the MCS queue lock or the changes to the metadata in this PR make sense.

PR #571, opened 2022-11-03
Change Pool to not use a lock

mjp41:
Add elements to the list without a lock.  This temporarily creates a cycle in the list, which means iterate could potentially return the same element twice.


nwf:
The Morello LLVM crash for purecap debug is tracked at https://git.morello-project.org/morello/llvm-project/-/issues/61 .

PR #574, opened 2022-12-04
fatal error making less C for C++20.

devnexen:


mjp41 (approved):
Nice.  This looks great.

davidchisnall:
The version check added in this PR is incorrect and is causing breakage with clang 15.  The `source_location` header is not present until 16.

It's also incredibly fragile to assume that compiler version implies C++ standard library version.  C++ has a standard mechanism for checking these features.  We should do:

```c++
#ifdef __cpp_lib_source_location
```

We might also want to check that `__cpp_lib_source_location >= 201907L`, but we can probably work with an older pre-standard version for this limited use.

PR #577, opened 2022-12-12
Catch CI up with github

nwf-msr:
Fallout from https://github.blog/changelog/2022-11-09-github-actions-ubuntu-latest-workflows-will-use-ubuntu-22-04/


mjp41 (approved):
LGTM.  If it not to much trouble, could you add a comment about why we aren't using latest. I guess we could evolve some tests to a later clang?

nwf-msr:
No trouble at all.  If you want to pick a `clang` version to switch to, or to add in addition to the `ubuntu-20.04`/`clang-10` mix, I'm happy to add those too.

nwf-msr:
Don't merge; testing newer clangs.  If it works, I'll expand some commentary and then we can merge.

nwf-msr:
OK, maybe ready now. :)

mjp41 (approved):
LGTM, thanks

PR #578, opened 2022-12-12
CHERI: Use another pagemap for arena pointers

nwf-msr:
Holding the arena pointers in slab metadata invites tricky problems in the presence of misbehaving clients and the current mix-in approach means that the metadata isn't a standard layout type.  Here's a first stab at using another `Pagemap` for the arena pointers in modern `snmalloc`.  We'd tried this approach before, but it feels much less bolted-on now, thanks to the "build-yourself-a-malloc kit" design that `snmalloc` has grown into.

This is, unfortunately, very slightly (0.2%) slower when used to run SPEC CPU2006 `471.omnetpp` (more or less a malloc performance benchmark) on the Morello cluster nodes, at the moment.  There may be a straightforward optimization I've missed; can investigate.

For review, probably best to proceed commit-by-commit rather than all at once, and you can ignore the #577 at the bottom.  As is not uncommon for my PRs, we may decide to pull some things off the bottom of this one to merge while the rest bakes a bit more.


mjp41:
Awesome, LGTM.  I think we can simplify FrontendMetadata and BackendMetadata, and push all needs for custom meta-data to use a different pagemap.   That works for this use case, and Monza.  I guess we should think about
* memory tagging
* other per-allocation tagging (used bit potentially for prompt UAF detection, or an alternative revocation mechanism.)

Those are the use cases where I have considered a custom meta-data.

src/snmalloc/backend_helpers/lockrange.h
```diff
@@ -7,6 +7,9 @@ namespace snmalloc
 {
   /**
    * Protect the ParentRange with a spin lock.
+   *
+   * Accesses via the ancestor() mechanism will bypass the lock and so
+   * should be used only where the resulting data races are acceptable.
```
mjp41, line 12: I wonder if there is a way to add some API that requires something to be explicitly set to pass a lock range with an ancestor call.

src/snmalloc/backend_helpers/pagemap.h
```diff
@@ -32,6 +32,24 @@ namespace snmalloc
...
+      "BasicPagemap's PagemapEntry type is not a MetaEntryBase");
+
+    /**
+     * This registration allocation hook serves to prevent snmalloc's backend
+     * ranges from consolidating across adjacent OS allocations on platforms
+     * (e.g., Windows or StrictProvenance) where that's required.  It is useful
+     * only for Pagemaps whose Entry type is a subclass of MetaEntryBase; if
+     * you need similar mechanism for different metadata structures, you'll
+     * need a different hook class.
+     */
```
nwf-msr, line 46: Ack; that commentary is stale from an earlier design.  It's the right gist, but refers to things that no longer exist.

src/snmalloc/backend/backend.h
```diff
@@ -23,9 +23,6 @@ namespace snmalloc
     using Pal = PAL;
     using SlabMetadata = typename PagemapEntry::SlabMetadata;
 
-#ifdef __cpp_concepts
-    static_assert(IsSlabMeta_Arena<SlabMetadata>);
```
rmn30, line 26: Is this concept now redundant?

nwf-msr: Good catch.  Should be; I'll remove it as part of an amended "deorbit" commit.

PR #580, opened 2022-12-15
Add BUILD_TESTING cmake option

nwf-msr:
As with CTest, but without the full machinery thereof.  This facilitates package builders to use the usual build targets (all, install) without needing to build the test programs if they're just going to get dropped on the floor.


nwf-msr:
Hrm.  I almost want to flip this to `OFF` by default, since I am struggling to explain to the BSD ports tree "yes, this build system defaults to the tests on, but I want you to default to the tests off, but turn them back on if the user explicitly asks".  :(

mjp41 (approved):

PR #581, opened 2022-12-22
adding few more c++17 operators override.

devnexen:


src/snmalloc/override/new.cc
```diff
@@ -71,3 +71,45 @@ void operator delete[](void* p, std::nothrow_t&)
 {
   ThreadAlloc::get().dealloc(p);
 }
+
+void* operator new(size_t size, std::align_val_t)
```
nwf-msr, line 75: Discarding the alignment requirement seems dubious?  While it's generally true that `snmalloc` prefer to align things much more strongly than generally required, it's still possible to ask for yet stronger alignment.  I think the correct thing to do is to use https://github.com/microsoft/snmalloc/blob/d0df2d028a45131e5b85dce873c0bcc572f6a57d/src/mem/sizeclasstable.h#L131 as done with https://github.com/microsoft/snmalloc/blob/d0df2d028a45131e5b85dce873c0bcc572f6a57d/src/override/malloc.cc#L158-L174
I do not know to what extent those checks are also required by C++ .

src/snmalloc/override/new.cc
```diff
@@ -71,3 +71,45 @@ void operator delete[](void* p, std::nothrow_t&)
...
+{
+  return ThreadAlloc::get().alloc(size);
+}
+
+void* operator new[](size_t size, std::align_val_t, std::nothrow_t&)
+{
+  return ThreadAlloc::get().alloc(size);
+}
+
+void operator delete(void* p, std::align_val_t)EXCEPTSPEC
```
nwf-msr, line 95:
```suggestion
void operator delete(void* p, std::align_val_t) EXCEPTSPEC
```

I'm surprised `clang-format` was OK with that without the whitespace.

src/snmalloc/override/new.cc
```diff
@@ -71,3 +71,45 @@ void operator delete[](void* p, std::nothrow_t&)
...
+{
+  ThreadAlloc::get().dealloc(p);
+}
+
+void operator delete[](void* p, std::align_val_t) EXCEPTSPEC
+{
+  ThreadAlloc::get().dealloc(p);
+}
+
+void operator delete(void* p, size_t size, std::align_val_t)EXCEPTSPEC
```
nwf-msr, line 105:
```suggestion
void operator delete(void* p, size_t size, std::align_val_t) EXCEPTSPEC
```

nwf-msr (dismissed):

src/snmalloc/override/new.cc
nwf-msr: ... oh wow, it in fact requires that it not be there.  That... that sure seems like a bug, but I guess it's best to do what it wants.  Sorry!

mjp41: @nwf-msr sorry I am confused now.  I think @devnexen has now added the space, but are you saying the space shouldn't be there?

mjp41:
@devnexen would be able to run clang-format to unblock the CI?

src/snmalloc/override/new.cc
devnexen: by the way is clang-format9 still recommended ? Thankfully I still have it somewhere.

mjp41: It is, but I might move this and Verona forward to a newer version.  It is getting a bit long in the tooth now.

mjp41 (approved):

Issue #584, opened 2023-01-26
`constexpr size_to_sizeclass` has a significant performance impact


pentelbart:
I haven't tested this on other platforms, only on CHERI RISC-V

The change to `constexpr` (#563) for `size_to_sizeclass` required `sizeclass_lookup` to lose `static` ([here](https://github.com/microsoft/snmalloc/blob/38d4483b2797d1c00eff9326f24b5cd609f52b4f/src/snmalloc/mem/sizeclasstable.h#L414)). My compiler emits this as a memcpy of the whole `SizeClassLookup` struct which runs every time the function is called, making it really costly for small allocations (~13k more instructions) . Changing it back eliminates the overhead, though I'm not clear enough on the surrounding context to know if this is the right move.

mjp41:
On x86 with Clang-10, the fast path for alloc is 
```asm
0000000000001bc0 <operator new(unsigned long)>:
    1bc0:       48 8d 47 ff             lea    -0x1(%rdi),%rax
    1bc4:       48 3d ff df 00 00       cmp    $0xdfff,%rax
    1bca:       77 3a                   ja     1c06 <operator new(unsigned long)+0x46>
    1bcc:       48 c1 e8 04             shr    $0x4,%rax
    1bd0:       48 8d 0d 41 96 00 00    lea    0x9641(%rip),%rcx        # b218 <snmalloc::sizeclass_metadata+0x1
208>
    1bd7:       0f b6 14 08             movzbl (%rax,%rcx,1),%edx
    1bdb:       48 8b 35 06 f4 00 00    mov    0xf406(%rip),%rsi        # 10fe8 <.got+0x18>
    1be2:       64 48 8b 04 25 00 00    mov    %fs:0x0,%rax
    1be9:       00 00
    1beb:       48 01 f0                add    %rsi,%rax
    1bee:       48 8d 0c d0             lea    (%rax,%rdx,8),%rcx
    1bf2:       64 48 8b 04 d6          mov    %fs:(%rsi,%rdx,8),%rax
    1bf7:       48 85 c0                test   %rax,%rax
    1bfa:       74 22                   je     1c1e <operator new(unsigned long)+0x5e>
    1bfc:       48 8b 10                mov    (%rax),%rdx
    1bff:       0f 18 0a                prefetcht0 (%rdx)
    1c02:       48 89 11                mov    %rdx,(%rcx)
    1c05:       c3                      retq
...
```

[...]

PR #585, opened 2023-01-26
Remove class nesting

mjp41:
Nesting the class inside a function generates bad code in Debug (#584). Unnesting seems to fix this.

### DIFF
--- a/src/snmalloc/mem/sizeclasstable.h
+++ b/src/snmalloc/mem/sizeclasstable.h
@@ -411,40 +411,40 @@ namespace snmalloc
     return (s - 1) >> MIN_ALLOC_BITS;
   }
 
-  constexpr smallsizeclass_t size_to_sizeclass(size_t size)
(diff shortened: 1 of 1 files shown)


mjp41:
So this almost halves the time taken in `./perf-singlethread-fast` in Debug.  

Before PR:
```
Count:  32768, Size:     16, ZeroMem: 1, Write: 1:    710732132 ns
```
After this PR:
```
Count:  32768, Size:     16, ZeroMem: 1, Write: 1:    402341898 ns
```

There also seems to be an improvement on time for Release.  I am going to investigate this more deeply.

mjp41:
The issues seems to be dependent on the version of Clang.  On Clang-10, I had slow downs, but not on Clang-12.

mjp41:
I propose taking this refactoring.

nwf:
Hrm.  It LGTM but the cluster CI seems to be acting up.  I'll look when I get a minute or when I get back.

PR #588, opened 2023-02-08
Attempt to run *BSD CI in VMs.

davidchisnall:


davidchisnall:
@devnexen, I've disabled the OpenBSD job for now because a load of tests fail.  If there's an update to the OpenBSD PAL that fixes them, that would be great.  I've disabled NetBSD because it doesn't run in xhyve and the qemu host is painfully slow: I have up watching it after 25 minutes.  [Here is the run with both enabled](https://github.com/microsoft/snmalloc/actions/runs/4122886320/jobs/7120219493).  It might eventually produce output for NetBSD, we can maybe do compile-only testing (or maybe exclude the perf tests?) if that doesn't take too long.  We may also get NetBSD-on-xhyve support soon (apparently 9.x boots on bhyve with UEFI mode, so should work, in theory).

mjp41 (approved):
LGTM

devnexen:
> @devnexen, I've disabled the OpenBSD job for now because a load of tests fail. If there's an update to the OpenBSD PAL that fixes them, that would be great. I've disabled NetBSD because it doesn't run in xhyve and the qemu host is painfully slow: I have up watching it after 25 minutes. [Here is the run with both enabled](https://github.com/microsoft/snmalloc/actions/runs/4122886320/jobs/7120219493). It might eventually produce output for NetBSD, we can maybe do compile-only testing (or maybe exclude the perf tests?) if that doesn't take too long. We may also get NetBSD-on-xhyve support soon (apparently 9.x boots on bhyve with UEFI mode, so should work, in theory).

Thanks for trying it, I ll have a look when I get the chance

PR #589, opened 2023-02-10
Try vm-actions instead of cross-platform-actions

davidchisnall:
This uses VirtualBox instead of xhyve.  It might be slower, but should be more reliable.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -288,7 +288,7 @@ if(NOT SNMALLOC_HEADER_ONLY_LIBRARY)
 
   if(NOT (DEFINED SNMALLOC_LINKER_FLAVOUR) OR ("${SNMALLOC_LINKER_FLAVOUR}" MATCHES "^$"))
     # Linker not specified externally; probe to see if we can make lld work
-    set(CMAKE_REQUIRED_LINK_OPTIONS -fuse-ld=lld)
+    set(CMAKE_REQUIRED_LINK_OPTIONS -fuse-ld=lld -Wl,--icf=all)
     check_cxx_source_compiles("int main() { return 1; }" LLD_WORKS)
     if (LLD_WORKS)
       message(STATUS "Using LLD to link snmalloc shims")
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -110,50 +110,100 @@ jobs:
 
   # GitHub doesn't natively support *BSD, but we can run them in VMs on Mac /
   # Linux runners
-  bsd:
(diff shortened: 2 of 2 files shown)


davidchisnall:
This is running the tests on {Net,Free,Open}BSD.  The tests currently fail on Net- and OpenBSD, though for different reasons.  I've wired up FreeBSD to the all-checks action, but not the other two, so the failing tests don't break merging but it's easy to see if a commit fixes them (once they're fixed, hopefully we can keep them from breaking again).

davidchisnall:
@devnexen, it looks as if some of the tests fail on NetBSD as well.

CMakeLists.txt
```diff
@@ -288,7 +288,7 @@ if(NOT SNMALLOC_HEADER_ONLY_LIBRARY)
 
   if(NOT (DEFINED SNMALLOC_LINKER_FLAVOUR) OR ("${SNMALLOC_LINKER_FLAVOUR}" MATCHES "^$"))
     # Linker not specified externally; probe to see if we can make lld work
-    set(CMAKE_REQUIRED_LINK_OPTIONS -fuse-ld=lld)
+    set(CMAKE_REQUIRED_LINK_OPTIONS -fuse-ld=lld -Wl,--icf=all)
```
mjp41, line 291: Why did you remove this?  It makes the library smaller as there are a lot of names for the same function in the API surface (e.g. lots of new overloads and malloc can all be folded into a single thing).

mjp41: FYI: this change increases the size of `libsnmallocshim.so` by about 10%.

devnexen:
> @devnexen, it looks as if some of the tests fail on NetBSD as well.

it comes down to the access enforcing thing. Once disabled all pass.

CMakeLists.txt
davidchisnall: I didn't remove it, I added it to the check that enables it.  Without this, we now have CI that triggers the reason that I have to carry a local patch for snmalloc to disable this: If you compile with GCC, the `-fuse-ld=lld` test fails (GCC consumes but silently ignores the flag?  Or just warns and doesn't error?), but then GCC passes this flag to ld.bfd and it fails to link.

davidchisnall:
@devnexen, please feel free to push to this branch or raise a PR against it.

CMakeLists.txt
mjp41: Thanks. I miss read the diff.

mjp41:
Can we merge this to unblock the CI?

PR #590, opened 2023-02-11
Update Morello CI

nwf-msr:
Motivated by not pinning us to the 2022.05 release any more; other changes while in the vicinity.


nwf-msr:
I don't think the CI failures are my fault; can I go despite them or should I wait?

mjp41 (approved):
LGTM. I think we should move FreeBSD CI out of all checks until it is more reliable again.

PR #593, opened 2023-02-19
custom memmove implementation proposal.

devnexen:
mostly like memcpy with optional bound checking but capable of handling overlapping cases thus using
reverse copy instead.

### DIFF
--- a/src/snmalloc/global/memcpy.h
+++ b/src/snmalloc/global/memcpy.h
@@ -42,6 +42,16 @@ namespace snmalloc
     }
   }
 
+  template<size_t Size, size_t PrefetchOffset = 0>
+  SNMALLOC_FAST_PATH_INLINE void
+  block_reverse_copy(void* dst, const void* src, size_t len)
+  {
+    for (size_t i = (len - 1); int64_t(i + Size) >= 0; i -= Size)
+    {
+      copy_one<Size>(pointer_offset(dst, i), pointer_offset(src, i));
+    }
+  }
+
   /**
    * Perform an overlapping copy of the end.  This will copy one (potentially
    * unaligned) `T` from the end of the source to the end of the destination.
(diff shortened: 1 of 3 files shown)


src/snmalloc/global/memcpy.h
```diff
@@ -459,4 +469,42 @@ namespace snmalloc
...
+
+    // Check the bounds of the arguments.
+    if (SNMALLOC_UNLIKELY(!check_bounds<(Checked && ReadsChecked)>(src, len)))
+      return report_fatal_bounds_error(
+        src, len, "memmove with source out of bounds of heap allocation");
+    if (SNMALLOC_UNLIKELY(!check_bounds<Checked>(dst, len)))
+      return report_fatal_bounds_error(
+        dst, len, "memmove with destination out of bounds of heap allocation");
+
+    if (dst > src)
```
mjp41, line 497: Would 
```suggestion
    if ((dst - src) < len)
```
be a better check?  Currently it doesn't check that there is actually an overlap?  @davidchisnall

mjp41 (commented):
Thanks for doing this.  I think the test coverage needs a bit of work.  It would be nice to vary the sizes and directions of overlap to make sure the code functions correctly.

src/test/func/memcpy/func-memcpy.cc
```diff
@@ -144,6 +160,37 @@ void check_bounds(size_t size, size_t out_of_bounds)
...
+  START_TEST("memmove overlaps, size {}", size);
+  auto* s = static_cast<unsigned int*>(my_malloc(size * sizeof(unsigned int)));
+  auto sz = size / 2;
+  for (size_t i = 0; i < size; ++i)
+  {
+    s[i] = static_cast<unsigned int>(i);
+  }
+  my_memmove(&s[2], &s[4], sizeof(s[0]));
+  EXPECT(s[2] == s[4], "overlap error: {} {}", s[2], s[4]);
+  my_memmove(&s[15], &s[5], sizeof(s[0]));
```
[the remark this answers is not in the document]
devnexen: ah right I forgot to change in the beginning the sizes set were fixed :-)

src/test/func/memcpy/func-memcpy.cc
```diff
@@ -144,6 +160,49 @@ void check_bounds(size_t size, size_t out_of_bounds)
...
+  my_memmove(ptr, s, size * sizeof(s[0]));
+  EXPECT(ptr == s, "overlap error: {} {}", ptr, s);
+  my_free(s);
+}
+
+template<bool after>
+void check_overlaps2(size_t size)
+{
+  START_TEST("memmove overlaps2, size {}", size);
+  auto sz = size / 2;
```
mjp41, line 186: I wonder if it would be good to separate the offset from the size being copied. The current test only covers very small overlaps.
```suggestion
  auto sz = size / 2;
  auto offset = size / 3;
```
Then use `offset` in the calculations using `after`?

src/snmalloc/global/memcpy.h
mjp41: I guess my concern here is because you have dropped to a slow byte by byte copy.  Half of all non-overlapping `memmove`s will be slow.

mjp41 (commented):
Thanks for fixing the tests.

src/snmalloc/global/memcpy.h
devnexen: right .. I ll follow your suggestion.

mjp41:
@devnexen sorry for the delay in running the CI.  It looks like there are a few failures in the tests, and clangformat is not happy.

devnexen:
weird as the issue seem about check_bounds about using memcpy ?
Also my clang-format(9) does not report format issue..

mjp41:
I think it is the last line here: 

```
Check fail: dst[u] == i in /home/runner/work/snmalloc/snmalloc/src/test/func/memcpy/func-memcpy.cc on 200  
in test void check_overlaps2(size_t) [after = false] overlap error: 0x4 0x2
```
https://github.com/microsoft/snmalloc/actions/runs/4256879925/jobs/7476767702#step:9:16563

The clangformat might be it has cached an old file list.  If you try rerunning `cmake`, and then the `clangformat` target?

mjp41 (approved):
LGTM

PR #597, opened 2023-02-27
Fix incorrect aligned_size invocations in operator overrides

EAirPeter:
The order of the arguments passed to aligned_size is reversed, causing the assertion `bits::is_pow2(alignment)` to fail. The fix is to simply change the order.


mjp41 (approved):
LGTM.  Thank you for your contribution.

PR #598, opened 2023-02-27
Alter glibc override to work with RTL_DEEPBIND

mjp41:
When using dlopen with RTL_DEEPBIND the LD_PRELOAD used by the majority of allocators does not work as both libc and snmallocs allocators can be called by various dso.

This patch uses GLIBC's __malloc_hook to override the allocator as well as LD_PRELOAD.  This means that all libraries will call snmalloc when performing allocation.


mjp41:
Fixes #595.

davidchisnall (approved):

Issue #586, opened 2023-02-02
Alternative remote queue initialisation


mjp41:
## Problem

The current implementation of remote queue is designed to always contain at least one element.
This is essential for efficient operation as it removes branchs from fast paths, but leads to complexity. 
At initialiasation time this means we need to allocate a single 16 byte object to prime the queue:

https://github.com/microsoft/snmalloc/blob/6be63b1ea82676db10a64faeeda2b4421c38a419/src/snmalloc/mem/corealloc.h#L179-L185

This causes a 16 byte size class to come into existence, and this is unlikely to go away.

This issues is about alterating the way the queue works, so that it can still always appear to contain one element, but without 
the need to bring a 16 byte size class into being.

## Current implementation

We could perform an alternative approach to initialising the remote allocator.  Currently there are two fields that represent the remote allocator:
 https://github.com/microsoft/snmalloc/blob/6be63b1ea82676db10a64faeeda2b4421c38a419/src/snmalloc/mem/remoteallocator.h#L49-L54
 
Once, we initialise our queue it looks like:
```
              back
                |
                v
front ------> dummy -next-> nullptr
```
and then once it receives the first deallocation it looks like
```
                            back
                              |
                              v
front ------> dummy -next-> first -next-> nullptr

```
It is only when there are at least two elements in the queue that we can remove an element.  Hence after the first deallocation we can dequeue the dummy element.
```
                            back
                              |
                              v
front  ------------------>  first -next-> nullptr

```

This illustrates the current implementation.

To detect if a queue is empty, we currently perform:

https://github.com/microsoft/snmalloc/blob/6be63b1ea82676db10a64faeeda2b4421c38a419/src/snmalloc/mem/remoteallocator.h#L79-L84

However, it should work equally well to perform:

```C++
front->next == nullptr
```

## Proposal

I propose we implement this using two stub objects to build the free queue to start with:
```C++
  alignas(CACHELINE_SIZE) freelist::AtomicQueuePtr back{nullptr};
  alignas(CACHELINE_SIZE) freelist::Object::T stub1{};
  alignas(CACHELINE_SIZE) freelist::Object::T stub2{};
```
The state is initialised by:
```C++
  stub1.set_next(stub2);
  stub2.set_next(nullptr);
  back = &stub1;
```

This effectively looks like a single element queue from the point of view of next fields, however, back is pointing to 
`stub1`.

```
back
  |
  v
stub1 -next-> stub2 -next-> nullptr
```
[...]

PR #604, opened 2023-03-23
Remote stub

mjp41:
This partially address #586.  It provides a branch on the enqueue path, which simplifies the code.

This removes the need for the smallest size class to be create when the allocator is first initialised.


nwf-msr (approved):
LGTM.  I'm glad the domestication machinery was amenable. :)

PR #606, opened 2023-03-23
Optimise buddy allocator to not search too high

mjp41:
The buddy allocator doesn't need to look at sizes above the current highest size. This commit tracks the highest block that is stored in the buddy allocator.

This is factored out of #570.


nwf-msr (approved):
Seems straight-forward enough.

PR #607, opened 2023-03-23
Minor perf

mjp41:
A few minor changes extracted from #570.

### DIFF
--- a/src/snmalloc/backend_helpers/range_helpers.h
+++ b/src/snmalloc/backend_helpers/range_helpers.h
@@ -16,7 +16,7 @@ namespace snmalloc
 
     // Find the minimum set of maximally aligned blocks in this range.
     // Each block's alignment and size are equal.
-    while (length >= sizeof(void*))
+    while (length >= bits::one_at_bit(MIN_BITS))
     {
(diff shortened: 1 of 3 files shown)


src/snmalloc/backend_helpers/range_helpers.h
```diff
@@ -16,7 +16,7 @@ namespace snmalloc
 
     // Find the minimum set of maximally aligned blocks in this range.
     // Each block's alignment and size are equal.
-    while (length >= sizeof(void*))
+    while (length >= bits::one_at_bit(MIN_BITS))
```
nwf-msr, line 19: I guess we always use this to construct allocatable things at the moment, so there's not much point in going smaller, but I wonder if the lower bound should be a template parameter?

nwf-msr (approved):

src/snmalloc/backend_helpers/range_helpers.h
mjp41: `MIN_BITS` is actually a template parameter.  The global constant is `MIN_ALLOC_BITS`.  It was only when I went to change the code, I realised this though :)

PR #608, opened 2023-03-28
Move key_global into RemoteAllocator

mjp41:
There was a mis-compilation in a Verona configuration that lead to two instances of key_global existing.  This change moves it inside a struct that seems to fix the issue.

The rest of the changes are limiting the use of key_global as both RemoteCache and RemoteAllocator must use the same configuration, so there is no need to take the key_global as a parameter.


nwf (approved):
LGTM.  I wonder if it's worth having a comment somewhere in the tree (maybe just in docs/security?) about the choice of using a single key.  On the one hand:
- we don't have to read anything extra from the `RemoteAllocator*` in the pagemap
- we don't have to do any rewriting when forwarding a list of messages
On the other:
- well, it is a shared secret and its shared nature might, theoretically, make it easier to leak?

PR #611, opened 2023-04-19
Update name of header-only flag in documentation

jumaffre:
This PR renames the flag used to enable the import of snmalloc as header-only in the documentation, in line with the name of the flag in [`CMakeLists.txt`](https://github.com/microsoft/snmalloc/blob/7f368bd6dabd6349534550df54a4e559b455cfc7/CMakeLists.txt#L35).


mjp41 (approved):
LGTM, thanks

PR #617, opened 2023-06-08
Conditional range

mjp41:
This PR adds a new type of Range that can allow a range to be disabled dynamically. 

This is used to disable some backend caching if the heap is very small, such as in the OpenEnclave case.

### DIFF
--- a/src/snmalloc/backend/meta_protected_range.h
+++ b/src/snmalloc/backend/meta_protected_range.h
@@ -120,6 +120,11 @@ namespace snmalloc
       return meta_range;
     }
 
+    void set_small_heap()
+    {
+      error("Not supported! Cannot use a small heap with MetaProtectedRange.");
+    }
+
     // Create global range that can service small meta-data requests.
     // Don't want to add the SmallBuddyRange to the CentralMetaRange as that
     // would require committing memory inside the main global lock.
--- a/src/snmalloc/backend/meta_protected_range.h
+++ b/src/snmalloc/backend/meta_protected_range.h
@@ -120,11 +120,6 @@ namespace snmalloc
       return meta_range;
     }
 
-    void set_small_heap()
-    {
-      error("Not supported! Cannot use a small heap with MetaProtectedRange.");
-    }
-
     // Create global range that can service small meta-data requests.
     // Don't want to add the SmallBuddyRange to the CentralMetaRange as that
     // would require committing memory inside the main global lock.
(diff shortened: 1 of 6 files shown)


src/snmalloc/backend/meta_protected_range.h
```diff
@@ -120,6 +120,11 @@ namespace snmalloc
       return meta_range;
     }
 
+    void set_small_heap()
+    {
+      error("Not supported! Cannot use a small heap with MetaProtectedRange.");
```
nwf-msr, line 125: I wonder if we can make this a compile-time error instead... in principle I think we could require `fixedglobalconfig` to do something like:
```
    if constexpr (requires { { local_state->set_small_heap() } -> std::same_as<void> })
```
but that's more `concept`-ual than we've wanted the codebase to be.

src/snmalloc/backend_helpers/staticconditionalrange.h
```diff
@@ -0,0 +1,68 @@
...
+    // Disabling is global, and not local.
+    // This is used to allow disabling thread local buddy allocators when the
+    // initial fixed size heap is small.
+    // 
+    // The range builds a more complex parent 
+    //    Pipe<ParentRange, OptionalRange>
+    // and uses the ancestor functions to bypass the OptionalRange if the flag
+    // has been set.
+    template<typename ParentRange>
+    class Type : public ContainsParent<Pipe<ParentRange, OptionalRange>>
```
nwf-msr, line 21: I've nothing to add here except to quote Hackers:

    It's so neat and clean.
    Doesn't look like a hacker wrote it.

mjp41: Now I need to watch "Hackers" ;)

src/snmalloc/backend/meta_protected_range.h
mjp41: If I delete this method, then the code still compiles, so we convert it to a compile time error as there is already enough conditional template stuff to make this unreachable.

src/snmalloc/backend_helpers/staticconditionalrange.h
```diff
@@ -0,0 +1,69 @@
...
+      // Global flag specifying if the optional range should be disabled.
+      static inline bool disable_range_{false};
+
+    public:
+      static constexpr bool Aligned = ActualParentRange::Aligned;
+
+      static constexpr bool ConcurrencySafe =
+        ActualParentRange::ConcurrencySafe;
+
+      using ChunkBounds = typename ActualParentRange::ChunkBounds;
```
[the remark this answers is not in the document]
mjp41: I have used `&&` for the two booleans, `ConcurrencySafe` and `Aligned`.  This means they don't need to be the same, but compatible with the requirements.

Issue #618, opened 2023-06-20
Unknown failure in CI on Risc-V


mjp41:
For #617 the CI failed a couple of times on RISC-V in the following way.  

```
24/40 Test #25: func-first_operation-check .........Subprocess aborted***Exception:  11.82 sec
......
assert fail: (back != nullptr) || (address_cast(front.load()) == address_cast(&stub)) in /home/runner/work/snmalloc/snmalloc/src/snmalloc/backend_helpers/../mem/remoteallocator.h on 69 

/home/runner/work/snmalloc/snmalloc/build/func-first_operation-check(_ZN8snmalloc8PALPOSIXINS_8PALLinuxEXadL_Z6writevEEXadL_Z5fsyncEEE17print_stack_traceEv+0x3c)[0x4000036904]
```
https://github.com/microsoft/snmalloc/actions/runs/5221697554/jobs/9455200536

The PR did not affect the code in anyway.  The failure has not been seen outside of this PR, and a separate branch to exercise this more has not resulted in any failures.  I'm raising this issue in case we see this again.

PR #619, opened 2023-06-20
Correct order of test based on #618

mjp41:


nwf-msr:
I _think_ we will end up (at least in the Cornucopia prototypes and the eventual CHERI+MTE world) with double-free guards being rather towards the front, before we could even get to `enqueue()`ing to the remote queue.  In the current stochastic detection regime, we're tripping an assert when we detect double-frees anyway, and we're much more likely to cause double-frees to trip over the intended "congratulations, you corrupted the message queue" assert(s) rather than this one.

nwf-msr (approved):

mjp41:
> I _think_ we will end up (at least in the Cornucopia prototypes and the eventual CHERI+MTE world) with double-free guards being rather towards the front, before we could even get to `enqueue()`ing to the remote queue. In the current stochastic detection regime, we're tripping an assert when we detect double-frees anyway, and we're much more likely to cause double-frees to trip over the intended "congratulations, you corrupted the message queue" assert(s) rather than this one.

So this is a `SNMALLOC_ASSERT` not a `SNMALLOC_CHECK`, so is more for debugging the allocator rather than detecting user errors.  I think the double free protection would catch issues much better than this assert.

PR #624, opened 2023-07-20
Factor out libc code into a header.

mjp41:
This pulls the main definitions of the various libc malloc functions into a header for easier use and inclusion in other projects.

### DIFF
--- a/src/snmalloc/mem/localalloc.h
+++ b/src/snmalloc/mem/localalloc.h
@@ -184,6 +184,13 @@ namespace snmalloc
       }
 
       return check_init([&](CoreAlloc* core_alloc) {
+        if (size > bits::one_at_bit(bits::BITS - 1))
+        {
+          // Cannot allocate something that is more that half the size of the
+          // address space
+          errno = ENOMEM;
+          return capptr::Alloc<void>{nullptr};
+        }
         // Grab slab of correct size
         // Set remote as large allocator remote.
         auto [chunk, meta] = Config::Backend::alloc_chunk(
@@ -712,7 +719,7 @@ namespace snmalloc
         auto pm_size = sizeclass_full_to_size(pm_sc);
         snmalloc_check_client(
           mitigations(sanity_checks),
-          sc == pm_sc,
+          (sc == pm_sc) || (p == nullptr),
           "Dealloc rounded size mismatch: {} != {}",
           rsize,
           pm_size);
--- a/src/snmalloc/override/libc.h
+++ b/src/snmalloc/override/libc.h
@@ -142,7 +142,8 @@ namespace snmalloc::libc
 
   inline void* memalign(size_t alignment, size_t size)
   {
-    if (SNMALLOC_UNLIKELY(alignment < sizeof(uintptr_t) || !bits::is_pow2(alignment)))
+    if (SNMALLOC_UNLIKELY(
+          alignment < sizeof(uintptr_t) || !bits::is_pow2(alignment)))
     {
       return set_error(EINVAL);
     }
@@ -158,7 +159,8 @@ namespace snmalloc::libc
 
   inline int posix_memalign(void** memptr, size_t alignment, size_t size)
   {
-    if (SNMALLOC_UNLIKELY((alignment < sizeof(uintptr_t) || !bits::is_pow2(alignment))))
+    if (SNMALLOC_UNLIKELY(
+          (alignment < sizeof(uintptr_t) || !bits::is_pow2(alignment))))
     {
       return EINVAL;
     }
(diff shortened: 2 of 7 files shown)


davidchisnall:
Do they still get inlined in the .cc file?

mjp41:
It looks good:
```
00000000000054e0 g     F .text  00000000000000b9              free
00000000000054e0 g     F .text  00000000000000b9              cfree
00000000000054e0 g     F .text  00000000000000b9              operator delete(void*)
00000000000054e0 g     F .text  00000000000000b9              operator delete[](void*)
00000000000054e0 g     F .text  00000000000000b9              operator delete(void*, std::align_val_t)
00000000000054e0 g     F .text  00000000000000b9              operator delete[](void*, std::align_val_t)

0000000000005430 g     F .text  00000000000000a8              malloc
0000000000005430 g     F .text  00000000000000a8              operator new(unsigned long)
0000000000005430 g     F .text  00000000000000a8              operator new[](unsigned long)
0000000000005ed0 g     F .text  00000000000000b6              operator new(unsigned long, std::align_val_t)
0000000000005ed0 g     F .text  00000000000000b6              operator new[](unsigned long, std::align_val_t)
```
and
```
0000000000005430 <operator new[](unsigned long)>:
    5430:       48 8d 47 ff             lea    -0x1(%rdi),%rax
    5434:       48 3d ff df 00 00       cmp    $0xdfff,%rax
    543a:       77 3a                   ja     5476 <operator new[](unsigned long)+0x46>
    543c:       48 c1 e8 04             shr    $0x4,%rax
    5440:       48 8d 0d 51 cc ff ff    lea    -0x33af(%rip),%rcx        # 2098 <snmalloc::sizeclass_lookup>
    5447:       0f b6 14 08             movzbl (%rax,%rcx,1),%edx
    544b:       48 8b 35 e6 e0 00 00    mov    0xe0e6(%rip),%rsi        # 13538 <_DYNAMIC+0x1e0>
    5452:       64 48 8b 04 25 00 00    mov    %fs:0x0,%rax
    5459:       00 00
    545b:       48 01 f0                add    %rsi,%rax
    545e:       48 8d 0c d0             lea    (%rax,%rdx,8),%rcx
    5462:       64 48 8b 04 d6          mov    %fs:(%rsi,%rdx,8),%rax
    5467:       48 85 c0                test   %rax,%rax
    546a:       74 22                   je     548e <operator new[](unsigned long)+0x5e>
    546c:       48 8b 10                mov    (%rax),%rdx
    546f:       0f 18 0a                prefetcht0 (%rdx)
    5472:       48 89 11                mov    %rdx,(%rcx)
    5475:       c3                      ret
    ....
```

src/snmalloc/mem/localalloc.h
```diff
@@ -184,6 +184,13 @@ namespace snmalloc
       }
 
       return check_init([&](CoreAlloc* core_alloc) {
+        if (size > bits::one_at_bit(bits::BITS - 1))
```
nwf-msr, line 187: I guess I understand not initializing the frontend in response to an absurd request, but this seems like the kind of thing that will be pretty rare?

src/snmalloc/override/libc.h
```diff
@@ -0,0 +1,181 @@
...
+#include <errno.h>
+#include <string.h>
+
+namespace snmalloc::libc
+{
+  /**
+   * Clang was helpfully inlining the constant return value, and
+   * thus converting from a tail call to an ordinary call.
+   */
+  static inline void* snmalloc_not_allocated = nullptr;
```
nwf-msr, line 14: Does marking this `const` re-enable the problematic inlining?

nwf-msr (commented):
Two hopefully quick questions, but the bulk of it seems fine.

src/snmalloc/mem/localalloc.h
mjp41: So `check_init` does always calls the supplied lambda, not just in the init case.

https://github.com/microsoft/snmalloc/blob/9d4466093a7c42e4fe43e032aeca356674d6e55c/src/snmalloc/mem/localalloc.h#L103-L111

This is dealing with this case off the fast path, and then setting the error.

src/snmalloc/override/libc.h
mjp41: I had a look at the code a bit more, and it no longer seems like it is a problem.

mjp41: Also ended up adding some SNMALLOC_FAST_PATH to force some inlining as `calloc` and `realloc` had introduced an additional tail call than needed.

nwf-msr (approved):
LGTM

PR #626, opened 2023-07-25
risc-v aal tick implementation proposal.

devnexen:


devnexen:
My [playground](https://godbolt.org/z/xxdosbWvn).

nwf-msr:
Having a play in your playground, it looks like `__builtin_readcyclecounter()` generates equivalent code, identical at optimizations above `-O0`.  Is there some problem in your environment with the generic path at
https://github.com/microsoft/snmalloc/blob/9d4466093a7c42e4fe43e032aeca356674d6e55c/src/snmalloc/aal/aal.h#L179 ?  (Does your compiler not have that builtin?  If so, which?)

PR #627, opened 2023-08-02
Using exclusive mode prefetch

mjp41:
The prefetching is always used to move the cache line to the current core for writing.  This change makes it use exclusive mode prefetch and enables it as a feature flag for x64.


nwf-msr (approved):
LGTM; thanks for taking the time to chat about this a few days ago

PR #632, opened 2023-09-12
Adding an option to disable TLS

matajoh:


CMakeLists.txt
```diff
@@ -26,6 +26,7 @@ option(SNMALLOC_NO_REALLOCARR "Build without reallocarr exported" ON)
 option(SNMALLOC_LINK_ICF "Link with Identical Code Folding" ON)
 option(SNMALLOC_IPO "Link with IPO/LTO support" OFF)
 option(SNMALLOC_BENCHMARK_INDIVIDUAL_MITIGATIONS "Build tests and ld_preload for individual mitigations" OFF)
+option(SNMALLOC_NO_TLS "Build without thread local storage" OFF)
```
mjp41, line 29: Sorry, this name is not good.  Can we call it
```
SNMALLOC_ENABLE_DYNAMIC_LOADING
```
The code is still going to use TLS, just compiled in a different and much slower way.

matajoh: To be clear, it should otherwise stay the same or should it now be default ON (and the check below inverted)?

mjp41 (approved):

PR #633, opened 2023-09-14
Fix #631

### LINKED ISSUE #631: _msize is not provided as an override?
Cross-posting from [Z3](https://github.com/Z3Prover/z3/issues/6890) and grepping through the source-code doesn't seem to provide an override for it

Naville:
Add wrapper override for the Windows variant of maloc_usable_size : _msize

### DIFF
--- a/src/snmalloc/override/malloc.cc
+++ b/src/snmalloc/override/malloc.cc
@@ -40,12 +40,15 @@ extern "C"
   {
     return snmalloc::libc::malloc_usable_size(ptr);
   }
-  SNMALLOC_EXPORT 
-  size_t SNMALLOC_NAME_MANGLE(_msize)( 
-    MALLOC_USABLE_SIZE_QUALIFIER void* ptr) 
-  { 
-    return snmalloc::libc::malloc_usable_size(ptr); 
-  } 
+
+#ifdef _WIN32
+  SNMALLOC_EXPORT
+  size_t SNMALLOC_NAME_MANGLE(_msize)(
+    MALLOC_USABLE_SIZE_QUALIFIER void* ptr)
+  {
+    return snmalloc::libc::malloc_usable_size(ptr);
+  }
+#endif
 
   SNMALLOC_EXPORT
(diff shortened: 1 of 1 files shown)


src/snmalloc/override/malloc.cc
```diff
@@ -40,6 +40,12 @@ extern "C"
   {
     return snmalloc::libc::malloc_usable_size(ptr);
   }
+  SNMALLOC_EXPORT 
+  size_t SNMALLOC_NAME_MANGLE(_msize)( 
+    MALLOC_USABLE_SIZE_QUALIFIER void* ptr) 
+  { 
+    return snmalloc::libc::malloc_usable_size(ptr); 
+  }
```
mjp41, line 48: Thanks for doing this.  If you could put it in a `ifdef` that would be great.  Also clang-format is complaining about the whitespace changes. The below suggestions should fix both.

```suggestion

#ifdef _WIN32
  SNMALLOC_EXPORT
  size_t SNMALLOC_NAME_MANGLE(_msize)(
    MALLOC_USABLE_SIZE_QUALIFIER void* ptr)
  {
    return snmalloc::libc::malloc_usable_size(ptr);
  }
#endif
```

mjp41 (changes requested):
Thanks for doing this.  Sorry I have been swamped by another project so didn't have time to replicate your issue.

Naville:
Done

src/snmalloc/override/malloc.cc
```diff
@@ -41,6 +41,15 @@ extern "C"
     return snmalloc::libc::malloc_usable_size(ptr);
   }
 
+#ifdef _WIN32
+  SNMALLOC_EXPORT
+  size_t SNMALLOC_NAME_MANGLE(_msize)(
+    MALLOC_USABLE_SIZE_QUALIFIER void* ptr)
```
mjp41, line 47:
```suggestion
  size_t SNMALLOC_NAME_MANGLE(_msize)(MALLOC_USABLE_SIZE_QUALIFIER void* ptr)
```

mjp41 (approved):

Issue #634, opened 2023-09-14
Implement slab level cache for remote frees 


mjp41:
@nwf-msr observed that we could improve the performance of remote deallocation if the producer does more work on building lists for each slab before returning to the original allocator.  This could improve producer/consumer scenarios further.

```[tasklist]
### Tasks
- [ ] Implementing statistics to estimate performance wins.
- [ ] Implement version that doesn't work with snmalloc hardening.
- [ ] Implement version with hardening as well 
```

mjp41:
@licenser @darach are you still using snmalloc. If so do you have any benchmarks that represent your workload?  We have some ideas that would benefit your kind of workload for Tremor.

darach:
Hi Matt,

We sure are. In tremor, and I have a project at work ( axiom now ) where I
think it's a great fit.

We'll ask the community as well

Cheers,

Darach.

mjp41:
Awesome thanks. Any benchmarks we can run would really help us in justifying the engineering work.

darach:
Our CI benchmarks are snmalloc based https://www.tremor.rs/benchmarks/ - relatively boring UI there.
The benchmarks we ran a year or two ago for tremor ( and mimalloc before it, and jemalloc before that ) are here:
https://github.com/tremor-rs/tremor-runtime/blob/main/bench/README.md

They have changed enough though in themselves and we rewrote the runtime and connectors so the benchmarking
code works differently so YMMV

nwf-msr:
I've spent a while playing with various cache strategies (though nothing very sophisticated, just sort of seeing what the low-hanging fruit was like).  For the two workloads I've tried, for three interesting choices of hashes, I currently have these message counts:

| Workload       | No caching | Perfect assembly  | 4-way direct hash |
|----------------|------------|-------------------|-------------------|
| `msgpass`      | 7,205,380  | 552,781; 35 rings | 1,076,843         |
| `xmalloc-test` | 2,388,551  | 317,057; 5 rings  | 317,065           |

The hash here is inspired by https://github.com/skeeto/hash-prospector but is sort of "a third" of that:
```
  hash = slab;
  hash *= 0x7feb352d;
  hash ^= hash >> 16;
  return (hash & 3)
```

mjp41:
[...]

PR #637, opened 2023-09-21
WIP: BatchIt

nwf-msr:
As part of the assessment of #634, but also perhaps more generally useful.  Opinions welcome.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -65,6 +65,9 @@ endif()
 set(SNMALLOC_MIN_ALLOC_SIZE "" CACHE STRING "Minimum allocation bytes (power of 2)")
 set(SNMALLOC_MIN_ALLOC_STEP_SIZE "" CACHE STRING "Minimum allocation step (power of 2)")
 
+set(SNMALLOC_DEALLOC_BATCH_RING_ASSOC "" CACHE STRING "Associativity of deallocation batch cache; 0 to disable")
+set(SNMALLOC_DEALLOC_BATCH_RING_SET_BITS "" CACHE STRING "Logarithm of number of deallocation batch cache associativity sets")
+
 if(MSVC AND SNMALLOC_STATIC_LIBRARY AND (SNMALLOC_STATIC_LIBRARY_PREFIX STREQUAL ""))
   message(FATAL_ERROR "Empty static library prefix not supported on MSVC")
 endif()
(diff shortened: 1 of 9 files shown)


mjp41:
Looks like it is leaking in some case
https://github.com/microsoft/snmalloc/actions/runs/6279469686/job/17055222812?pr=637#step:7:149

nwf-msr:
> Looks like it is leaking in some case https://github.com/microsoft/snmalloc/actions/runs/6279469686/job/17055222812?pr=637#step:7:149

Whoops; I had the loop termination conditions wrong.  They're fixed now, I think.  Let's see if CI agrees.

nwf-msr:
After discussions with @mjp41 yesterday, I've introduced a notion of "tweakable obfuscation" and have made all the intra-slab free lists' backwards signatures use the address of the slab metadata as the "tweak".  The next step would be to remove the per-thread keys and have everyone use a common global key (probably not `RemoteAllocator::key_global`!) and apply the same tweaking.  This opens the door to sending threads being able to build up segments of slab free lists that can be spliced in by the recipient in O(1) rather than  O(n).

nwf-msr:
I've (at long last) got things flying end to end with a very simple "cache" on the sending side -- a single open ring -- but I think some review and investigation is a good idea.  Here's what `mimalloc-bench` makes of the current state of things in terms of time
image
and memory
image

nwf-msr:
We should:

1. figure out how to make the caching layer optional, which probably just means some more
`std::conditional_t` use.
2. get someone with chops to assess the "tweaked obfuscation" changes.
3. offer to randomize deallocator cache construction order (Matt writes: "as we are building a ring, we can add to the start or the end, so perhaps we could at least build an unpredictable order in the ring")
4. offer probabilistic premature eviction from the deallocator caches to further thwart attempts to control free-list order

mjp41:
Two things to address:
* Randomisation - this might break some of the randomisation, can we use the ways to build multiple queues for the same slab.
* Can we disable this feature as feature flag and constexpr/conditional_t, so we can analyse performance more in the future.

nwf-msr:
Just rebasing after #659 landed.  Todo-s remain to be addressed.

nwf-msr:
And the novelty of `[[no_unique_address]]` continues to sting.  Hm.

nwf-msr:
Start addressing to-dos, specifically being able to turn BatchIt off: rewrite history to have always had a `RemoteDeallocCacheBatching` structure that encapsulates the client-side logic.  We can pair this with the current `RemoteMessage` structure, then add parallel non-batching implementations of the client-side logic and the `RemoteMessage` internals.

nwf-msr:
OK, well, apparently `MAX_CAPACITY_BITS` needs to be at least `13`, were it to be universal:  64-bit cross-builds under qemu use a `MIN_CHUNK_SIZE` of `17` and a smallest sizeclass of `16` bytes.

But that breaks all the 32-bit builds, because `MAX_SMALL_SIZECLASS_BITS` is `16` and `MIN_OBJECT_COUNT` is sometimes `13`, and `16 + ceil(log_2(13)) + 13` is `33`.

I'm not sure why some other Windows builds are still broken.

nwf-msr:
OK, here's a version with, I think, all todo-s at least prototyped.  BatchIt can be turned off (see the "Make BatchIt optional" commit and/or CMakeLists.txt) and it now attempts to play nicely with `mitigations(random_preserve)` by stochastically evicting batches during construction.

nwf-msr:
Rebased after merging #675.

PR #639, opened 2023-09-27
Startup improvements

mjp41:
This PR improves the startup cost of snmalloc.

It adds a benchmark for profiling startup by starting a lot of threads, waiting for them to reach a stable point, each perform a single allocation deallocation pair, and terminate.  It measure the time taken from all threads starting allocation until the last one has completed its deallocation.

The second commit makes the Pool pass the `spare` memory in a meta-data allocation to the constructor of that type. This is then used to seed the local state in the backend with a small amount of meta-data, and means the first allocation does not need to request more meta-data memory.

The benchmark goes from taking ~2.5M cycles to ~1.3M cycles for the checked build, and ~1.4M cycles to ~1.1M cycles for the fast build (wall clock cycles).

### DIFF
--- a/src/snmalloc/mem/corealloc.h
+++ b/src/snmalloc/mem/corealloc.h
@@ -555,8 +555,11 @@ namespace snmalloc
     /**
      * Initialiser, shared code between the constructors for different
      * configurations.
+     *
+     * spare is the amount of space directly after the allocator that is
+     * reserved as meta-data, but is not required by this CoreAllocator.
      */
-    void init()
+    void init(size_t spare)
     {
 #ifdef SNMALLOC_TRACING
       message<1024>("Making an allocator.");
@@ -565,6 +568,14 @@ namespace snmalloc
       // it generates.
       // This must occur before any freelists are constructed.
       entropy.init<typename Config::Pal>();
+
+      if (spare != 0)
+      {
+        capptr::Arena<void> spare_start = capptr::Arena<void>::unsafe_from((
+          void*)pointer_offset((uintptr_t)this, sizeof(CoreAllocator<Config>)));
+        Config::Backend::dealloc_meta_data(
+          get_backend_local_state(), spare_start, spare);
+      }
 
       // Ignoring stats for now.
       //      stats().start();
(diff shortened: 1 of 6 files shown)


src/snmalloc/mem/pool.h
```diff
@@ -152,8 +152,14 @@ namespace snmalloc
...
 
+      capptr::Alloc<void> spare_start = pointer_offset(raw, sizeof(T));
+      Range<capptr::bounds::Alloc> r{spare_start, spare};
+
       auto p = capptr::Alloc<T>::unsafe_from(
-        new (raw.unsafe_ptr()) T(spare, std::forward<Args>(args)...));
+        new (raw.unsafe_ptr()) T(r, std::forward<Args>(args)...));
+
+      // Remove excess from the permissions.
+      p = Aal::capptr_bound<T, capptr::bounds::Alloc>(p, sizeof(T));
```
nwf-msr: Are we guaranteed that `p` is sufficiently aligned that `sizeof(T)` is representable?

src/snmalloc/aal/aal.h
```diff
@@ -205,7 +205,8 @@ namespace snmalloc
     capptr_bound(CapPtr<U, BIn> a, size_t size) noexcept
     {
       static_assert(
-        BIn::spatial > capptr::dimension::Spatial::Alloc,
+        (BIn::spatial > capptr::dimension::Spatial::Alloc) ||
+          (capptr::dimension::Spatial::Alloc == BOut::spatial),
         "Refusing to re-bound Spatial::Alloc CapPtr");
```
nwf-msr, line 210: That largely de-fangs this assert, so it could be worth removing instead.

src/snmalloc/mem/pool.h
mjp41: Good catch, I've rounded the size up now.

src/snmalloc/aal/aal.h
mjp41: Removed.

src/snmalloc/mem/corealloc.h
```diff
@@ -566,6 +569,12 @@ namespace snmalloc
       // This must occur before any freelists are constructed.
       entropy.init<typename Config::Pal>();
 
+      if (spare.length != 0)
+      {
+        Config::Backend::dealloc_meta_data(
```
nwf-msr, line 574: It might not hurt to have a comment here, for mere mortals like me that don't always readily recall that "backend" things come in per-frontend and shared state flavors.  Maybe something like...

```
/*
 * Seed this frontend's private metadata allocation cache with any
 * excess space from the metadata allocation holding the frontend
 * Allocator object itself.  This alleviates thundering herd
 * contention on the backend during startup: each slab opened now
 * makes one trip to the backend, for the slab itself, rather than
 * two, for the slab and its metadata.
 */
```

src/snmalloc/mem/pool.h
```diff
@@ -141,16 +141,26 @@ namespace snmalloc
         }
       }
 
+      size_t request_size = bits::next_pow2(sizeof(T));
+      size_t round_sizeof = Aal::capptr_size_round(sizeof(T));
+      size_t spare = request_size - round_sizeof;
```
nwf-msr, line 146: Hm.  I'm not quite sure what the theorem for CHERI Concentrate looks like, exactly, but in general, `spare` bytes at `spare_start` is not guaranteed to be representable even if `request_size` and `round_sizeof` (both at `raw`) are.  But maybe that's OK, since we are just returning the `Range` to the cache, not actually constructing a capability with those bounds?

Presumably rounding up to powers of two is maximally convenient for the backend and, in addition, tends to give us a reasonable amount of `spare` space?  (If it got to be excessive, we could instead pad upwards by smaller powers of two, namely `__builtin_cheri_representable_alignment_mask(sizeof(T)) + 1`, and still not risk the backend trying to build a capability to a non-representable range.)

nwf-msr (approved):
Despite the comment I just raised, I think this is good to go as it stands.

src/snmalloc/mem/pool.h
mjp41: The smallbuddy will split the returned amount into a collection of powers of two, so all nice and representable.

PR #641, opened 2023-09-29
Template construction of Pool elements

### LINKED ISSUE #603: Enable Pool to use a malloc
The `Pool` setup is primarily designed to be a trivial allocator that exists before the allocator functionality has come online.  It is the bootstrap allocator.

It is also used by verona-rt as a way of knowing all the elements of a particular allocation.  I.e. the Epoch code.  We should refactor the Pool to either support memory coming from `malloc` or factor the bootstrap allocation from the set of things code.

mjp41:
The Pool class is used by verona-rt.  The recent changes made this less nice to consume as an API.

This change makes the construction logic a template parameter to the Pool. This enables standard allocation to be used from Verona.

Fixes #603

### DIFF
--- a/src/snmalloc/mem/corealloc.h
+++ b/src/snmalloc/mem/corealloc.h
@@ -1040,7 +1040,7 @@ namespace snmalloc
 
       auto p = capptr::Alloc<CA>::unsafe_from(new (raw.unsafe_ptr()) CA(r));
 
-      // Remove excess from the permissions.
+      // Remove excess from the bounds.
       p = Aal::capptr_bound<CA, capptr::bounds::Alloc>(p, round_sizeof);
       return p;
     }
--- a/src/snmalloc/mem/corealloc.h
+++ b/src/snmalloc/mem/corealloc.h
@@ -621,8 +621,7 @@ namespace snmalloc
     template<
       typename Config_ = Config,
       typename = std::enable_if_t<Config_::Options.CoreAllocOwnsLocalState>>
-    CoreAllocator(Range<capptr::bounds::Alloc>& spare, LocalCache* cache)
-    : attached_cache(cache)
+    CoreAllocator(Range<capptr::bounds::Alloc>& spare)
     {
       init(spare);
     }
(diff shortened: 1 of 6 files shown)


src/snmalloc/mem/pool.h
```diff
@@ -116,8 +90,8 @@ namespace snmalloc
    */
   template<
     typename T,
-    SNMALLOC_CONCEPT(IsConfig) Config,
-    PoolState<T>& get_state() = SingletonPoolState<T, Config>::pool>
+    typename ConstructT = DefaultConstruct<T>,
```
nwf, line 93: Do you want to `concept`-ualize your `ConstructT`-ors?

src/snmalloc/mem/corealloc.h
```diff
@@ -1015,9 +1014,42 @@ namespace snmalloc
...
+      {
+        Config::Pal::error("Failed to initialise thread local allocator.");
+      }
+
+      capptr::Alloc<void> spare_start = pointer_offset(raw, round_sizeof);
+      Range<capptr::bounds::Alloc> r{spare_start, spare};
+
+      auto p = capptr::Alloc<CA>::unsafe_from(new (raw.unsafe_ptr()) CA(r));
+
+      // Remove excess from the permissions.
```
nwf-msr, line 1043: nit: bounds, rather than permissions?

src/snmalloc/mem/corealloc.h
```diff
@@ -621,8 +621,7 @@ namespace snmalloc
     template<
       typename Config_ = Config,
       typename = std::enable_if_t<Config_::Options.CoreAllocOwnsLocalState>>
-    CoreAllocator(Range<capptr::bounds::Alloc>& spare, LocalCache* cache)
-    : attached_cache(cache)
+    CoreAllocator(Range<capptr::bounds::Alloc>& spare)
```
nwf-msr, line 624: This is going to leave `attached_cache` `nullptr` "more often" after construction (I guess you could always have passed in `nullptr` for the 2nd arg).  Are we guaranteed that `attach()` is called before `flush()` or before `handle_message_queue_inner()` decides to `post()`?

nwf-msr (approved):

src/snmalloc/mem/corealloc.h
mjp41:
```suggestion
      // Remove excess from the bounds.
```

PR #643, opened 2023-10-10
Add _base variants of Windows Heap API, partially close #638

### LINKED ISSUE #638: Maybe missing _base variants in overrides
Continuing the discussion from #631 since that one is closed....

After applying the patch that also overrides ``_msize_base``, the project compiles. but got greeted with many "Not allocated by this allocator" error. 

Running ``dumpbin /SYMBOLS`` on the ucrt lib shows there is a ``_base`` variant for many other heap related APIs.

I wonder these information that might requires some microsoft internal employee's knowledge:

- Why these _base variant calls appeared in the final binary when no source code referenced them?
-  Should snmalloc provide overrides for them too?

Naville:
I'm still not quite sure how the ``_recalloc`` implementation works so I kept that part out

### DIFF
--- a/src/snmalloc/override/malloc.cc
+++ b/src/snmalloc/override/malloc.cc
@@ -20,11 +20,23 @@ extern "C"
   {
     return ThreadAlloc::get().alloc(size);
   }
+  #ifdef _WIN32
+   SNMALLOC_EXPORT void* SNMALLOC_NAME_MANGLE(_malloc_base)(size_t size)
+   {
+     return ThreadAlloc::get().alloc(size);
+   }
+   #endif
 
   SNMALLOC_EXPORT void SNMALLOC_NAME_MANGLE(free)(void* ptr)
   {
     ThreadAlloc::get().dealloc(ptr);
   }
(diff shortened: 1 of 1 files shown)


mjp41:
Thanks for doing this.  I think your branch is a bit behind `main`.  In particular, I think it is behind #624 which refactored this code to be a bit cleaner.  Would you be change in line with that?  Thanks

davidchisnall:
Do you have a link to the documentation for these?  I could not find them on MSDN.

mjp41:
Documentation on this is hard to come by.  When I have talked to people they have been unkeen to officially support anything, but they try not to break the tricks other people are using.

PartititonAlloc provides these:
https://source.chromium.org/chromium/chromium/src/+/main:base/allocator/partition_allocator/src/partition_alloc/shim/allocator_shim_override_ucrt_symbols_win.h

Naville:
> Do you have a link to the documentation for these? I could not find them on MSDN.

In my use case, without these variants the linker still tries to pull them in from ucrt, which results in duplicate definition error

Issue #646, opened 2023-11-07
FreeBSD and NetBSD CI broken


mjp41:
The CI for FreeBSD and NetBSD is broken.  The vmaction repo appear to be undergoing some issues:

https://github.com/vmactions/freebsd-vm/issues/74

PR #647, opened 2023-11-07
FreeBSD CI.

mjp41:
Fixes the FreeBSD side of #646


davidchisnall (approved):
Didn't we move away from cross-platform-actions because they were flaky?

Maybe we can just add Cirrus-CI and get native FreeBSD on Google Computer Engine VMs?

mjp41:
> Didn't we move away from cross-platform-actions because they were flaky?
> 
> Maybe we can just add Cirrus-CI and get native FreeBSD on Google Computer Engine VMs?

Still flaky.  It didn't work on first try.

PR #648, opened 2023-11-08
Update NetBSD runner

mjp41:
Addresses second part of #646


nwf-msr (approved):
If CI's happy, I'm happy.  Odd that `/usr/sbin` was removed from `$PATH` but whatever.

PR #651, opened 2024-01-04
NFC: sizeclass: differentiate minimum step size and minimum allocation sizes

nwf-msr:
No functional change intended yet. Just going through the exercise of separating `MIN_ALLOC_SIZE`'s two hats, but leaving the two at the same value.

So why?

We think this might be useful, letting us use less space in #637 (by leaving `MIN_ALLOC_STEP_SIZE` at `16` but raising `MIN_ALLOC_SIZE` to `32`, meaning that sizeclasses 0 and 1 would not be allocable), but it might be more generally useful.  It's also probably easier to review without all the rest of that mess at the same time.  Speaking of review, as per usual, this might be best read in commit order rather than all at once, and the punchline is in the last commit.


mjp41 (approved):
LGTM

I wonder if we should make STEP or SIZE configurable from CMake, so we can experiment with configurations.

It would be interesting to reduce step to 8, which would get us 16, 24, 32, 40, 48, 56, 64, 80, ...

nwf-msr:
Like that?

mjp41:
> Like that?

Is there a static_assert that the main is at least two pointers?

nwf-msr:
> Is there a static_assert that the main is at least two pointers?

Yes, two, sort of.  Once directly by the configuration:

https://github.com/nwf-msr/snmalloc/blob/ab87e2b49b7371a1fa93a96f131703a87925fcd3/src/snmalloc/ds/allocconfig.h#L104-L106

And once by the `freelist::Object` that actually imposes the requirement:

https://github.com/nwf-msr/snmalloc/blob/ab87e2b49b7371a1fa93a96f131703a87925fcd3/src/snmalloc/mem/freelist.h#L429-L431

mjp41 (approved):
LGTM

nwf-msr:
Out of curiosity, I did a little parameter sweeping with `mimalloc-bench`.  `aNN` is the minimum size, `sNN` is the step size.  (That is, `a16-s16` is the current policy on amd64.)  Don't take these results too seriously, they're from my laptop and the machine was not particularly quiesced at the time.

image

It could be more interesting to do an analogous sweep on CHERI, where `a32-s8` or `a32-s16` might be of interest, or with #637, where `a64-s16` or `a64-s32` might be.

mjp41:
Looks pretty noisy, but I can run some experiments on Monday

mjp41:
@nwf-msr would it make sense to add something like:
```CMake
  if (SNMALLOC_BENCHMARK_INDIVIDUAL_CONFIGS)
    foreach (STEP 8 16 32)
      foreach (MIN 16 32 64)
        if ((${STEP} EQUAL 8) AND (${MIN} EQUAL 64))
          continue()
        endif()
        set(DEFINES "-DSNMALLOC_MIN_ALLOC_STEP_SIZE=${STEP}" "-DSNMALLOC_MIN_ALLOC_SIZE=${MIN}")
        set(NAME "step-${STEP}-min-${MIN}")
        add_shim(snmallocshim-${NAME} SHARED ${SHIM_FILES})
        target_compile_definitions(snmallocshim-${NAME} PRIVATE ${DEFINES})
        if (SNMALLOC_BUILD_TESTING)
          make_tests(${NAME} ${DEFINES})
        endif()
      endforeach()
    endforeach()
  endif() # SNMALLOC_BENCHMARK_INDIVIDUAL_CONFIGS
```
So that we can rebuild the libraries for the test?

mjp41:
Here is running on my perf VM in Azure:

memory_nwf
time_nwf

I dropped a couple of the noisier ones, but overall it doesn't have much change.

mjp41:
memory_nwf
time_nwf

src/snmalloc/mem/sizeclasstable.h
```diff
@@ -421,13 +422,29 @@ namespace snmalloc
...
       size_t curr = 1;
-      for (sizeclass_compress_t sizeclass = 0;
-           sizeclass < NUM_SMALL_SIZECLASSES;
-           sizeclass++)
+
+      sizeclass_compress_t sizeclass = 0;
+      for (; sizeclass < minimum_class; sizeclass++)
+      {
+        for (; curr <= sizeclass_metadata.fast_small(sizeclass).size;
+             curr += 1 << MIN_ALLOC_STEP_BITS)
```
mjp41, line 438:
```suggestion
             curr += MIN_ALLOC_STEP_SIZE)
```

src/snmalloc/mem/sizeclasstable.h
```diff
@@ -421,13 +422,29 @@ namespace snmalloc
...
+        {
+          table[sizeclass_lookup_index(curr)] = minimum_class;
+        }
+      }
+
+      for (; sizeclass < NUM_SMALL_SIZECLASSES; sizeclass++)
       {
         for (; curr <= sizeclass_metadata.fast_small(sizeclass).size;
-             curr += 1 << MIN_ALLOC_BITS)
+             curr += 1 << MIN_ALLOC_STEP_BITS)
```
mjp41, line 447:
```suggestion
             curr += MIN_ALLOC_STEP_SIZE)
```

mjp41 (approved):
Lgtm

nwf-msr:
Taking those two suggestions; thanks!  Will merge as soon as CI's happy.

PR #656, opened 2024-04-07
Add Windows ARM64 support

jblazquez:
This PR solves a few issues when building for Windows on ARM:

1. No detection for the Windows ARM64 architecture, resulting in no AAL being selected.
2. A compilation error in `bits.h` due to the use of Intel intrinsics that don't exist on ARM.
3. A misdetection of the ARM64EC ABI as AMD64, leading to the use of an unsupported Intel intrinsic.

The first two issues manifest as follows. You can repro in an [ARM64 Visual Studio Developer Command Prompt](https://devblogs.microsoft.com/cppblog/arm64ec-support-in-visual-studio/#developer-command-prompt):

```
> cl /nologo /diagnostics:caret /EHsc /std:c++20 /c malloc.cc

ds_core\bits.h(162,14): error C3861: '_tzcnt_u64': identifier not found
      return _tzcnt_u64(static_cast<unsigned __int64>(x));
             ^
ds_core\bits.h(208,21): error C3861: '_umul128': identifier not found
      size_t prod = _umul128(x, y, &high_prod);
                    ^
pal\../aal/aal.h(259,50): error C2065: 'AAL_Arch': undeclared identifier
  using Aal = AAL_Generic<AAL_NoStrictProvenance<AAL_Arch>>;
```

To fix these two issues, this change adds detection of the Windows ARM64 architecture so the ARM AAL is selected. It also adds implementations of `snmalloc::bits::ctz` and `snmalloc::bits::umul` that work on Windows ARM64 under MSVC.

The third issue shows up like this:

```
> cl /nologo /diagnostics:caret /EHsc /std:c++20 /c /arm64EC malloc.cc

aal\aal_x86.h(82,7): error C3861: '_m_prefetchw': identifier not found
      _m_prefetchw(ptr);
      ^
```

When building for the [ARM64EC ABI](https://blogs.windows.com/windowsdeveloper/2021/06/28/announcing-arm64ec-building-native-and-interoperable-apps-for-windows-11-on-arm/), the compiler [defines the `_M_X64` macro and _not_ the `_M_ARM64` macro](https://techcommunity.microsoft.com/t5/windows-os-platform-blog/getting-to-know-arm64ec-defines-and-intrinsic-functions/ba-p/2957235), to make incremental porting existing X64 Windows code to ARM easier.

This means that the library was detecting ARM64EC as AMD64 and defining `PLATFORM_IS_X86` instead of `PLATFORM_IS_ARM`. This causes the library to compile in the x86 AAL and use several Intel intrinsics, which actually does work in almost all cases because Windows provides [an emulated implementation of Intel SIMD instruction sets](http://www.emulators.com/docs/abc_arm64ec_explained.htm) up to SSE4.2 in the `softintrin.lib` library. However, it does not provide an implementation of `_m_prefetchw`.

To fix this issue, the CPU architecture detection in `aal.h` is now aware of the ARM64EC ABI.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -361,7 +361,7 @@ jobs:
         # Build each combination of OS and release/debug variants
         os: [ windows-2019 ]
         build-type: [ Release, Debug ]
-        arch: [ Win32, x64 ]
+        arch: [ Win32, x64, ARM64, ARM64EC ]
         toolchain: [ "", "-T ClangCL" ]
         extra-cmake-flags: [ "" ]
         # Add an extra check for the Windows 8 compatible PAL
@@ -388,6 +388,24 @@ jobs:
             build-type: Debug
             arch: x64
             toolchain: ""
+          - os: windows-2022
+            build-type: Release
+            arch: ARM64
+            toolchain: ""
+          - os: windows-2022
+            build-type: Debug
+            arch: ARM64
+            toolchain: ""
+          - os: windows-2022
+            build-type: Release
+            arch: ARM64EC
+            toolchain: ""
+            extra-cmake-flags: -DCMAKE_SYSTEM_VERSION="10.0.22621.0"
+          - os: windows-2022
+            build-type: Debug
+            arch: ARM64EC
+            toolchain: ""
+            extra-cmake-flags: -DCMAKE_SYSTEM_VERSION="10.0.22621.0"
         
       # Don't abort runners if a single one fails
       fail-fast: false
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -361,7 +361,7 @@ jobs:
         # Build each combination of OS and release/debug variants
         os: [ windows-2019 ]
         build-type: [ Release, Debug ]
-        arch: [ Win32, x64, ARM64, ARM64EC ]
+        arch: [ Win32, x64 ]
         toolchain: [ "", "-T ClangCL" ]
         extra-cmake-flags: [ "" ]
         # Add an extra check for the Windows 8 compatible PAL
@@ -392,20 +392,24 @@ jobs:
             build-type: Release
             arch: ARM64
             toolchain: ""
+            build-only: yes
           - os: windows-2022
             build-type: Debug
             arch: ARM64
             toolchain: ""
+            build-only: yes
           - os: windows-2022
             build-type: Release
             arch: ARM64EC
             toolchain: ""
             extra-cmake-flags: -DCMAKE_SYSTEM_VERSION="10.0.22621.0"
+            build-only: yes
           - os: windows-2022
             build-type: Debug
             arch: ARM64EC
             toolchain: ""
             extra-cmake-flags: -DCMAKE_SYSTEM_VERSION="10.0.22621.0"
+            build-only: yes
         
       # Don't abort runners if a single one fails
       fail-fast: false
@@ -421,6 +425,7 @@ jobs:
       run: cmake --build ${{github.workspace}}/build -- /m /p:Configuration=${{ matrix.build-type }}
       # Run the tests.
     - name: Test
+      if: ${{ matrix.build-only != 'yes' }}
       working-directory: ${{ github.workspace }}/build
       run: ctest -j 2 --interactive-debug-mode 0 --output-on-failure -C ${{ matrix.build-type }} --timeout 400
       timeout-minutes: 20
--- a/src/snmalloc/ds_core/bits.h
+++ b/src/snmalloc/ds_core/bits.h
@@ -158,7 +158,11 @@ namespace snmalloc
       SNMALLOC_ASSERT(x != 0); // Calling with 0 is UB on some implementations
 
 #if defined(_MSC_VER) && !defined(__clang__)
-#  ifdef _WIN64
+#  if defined(_M_ARM64) || defined(_M_ARM64EC)
+      unsigned long n = 0;
+      _BitScanForward64(&n, static_cast<unsigned __int64>(x));
+      return static_cast<size_t>(n);
+#  elif defined(_WIN64)
       return _tzcnt_u64(static_cast<unsigned __int64>(x));
 #  else
       return _tzcnt_u32(static_cast<unsigned int>(x));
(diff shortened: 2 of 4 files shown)


jblazquez:
@microsoft-github-policy-service agree company="Riot Games Inc."

CC @paulgray

src/snmalloc/ds_core/bits.h
```diff
@@ -203,7 +207,12 @@ namespace snmalloc
       overflow = __builtin_mul_overflow(x, y, &prod);
       return prod;
 #elif defined(_MSC_VER)
-#  ifdef _WIN64
+#  if defined(_M_ARM64) || defined(_M_ARM64EC)
+      size_t high_prod = __umulh(x, y);
+      size_t prod = x * y;
+      overflow = high_prod != 0;
+      return prod;
+#  elif defined(_WIN64)
```
mjp41, line 215: Would it be better to use the same code with `__umulh` for the x64 version as well?  It seems that would be optimised to the same code, and then we have slightly less branching in the `ifdef`s.

mjp41 (approved):
LGTM.  Thank you for the very high quality PR with the explanations of the changes.  It is really appreciated.

I'm happy to take this as is, but if you think the slight simplification for `umul` works, then I would prefer that. 

Also, if you have any thoughts on how we might add CI to cover this case then I would be happy to add that.  If it isn't built in CI, it will often get broken.

src/snmalloc/ds_core/bits.h
jblazquez: Sure, I'll take a look.

jblazquez: @mjp41, on X64 the code generated for the `__umulh` version contains two multiplications, whereas the one for `_umul128` contains only one. Would you like me to use `__umulh` in both cases or keep it as is?

```
_umul128
00007FF663A51000  mov         r9,rdx  
00007FF663A51003  mov         rax,rdx  
00007FF663A51006  mul         rax,rcx  
00007FF663A51009  test        rdx,rdx  
00007FF663A5100C  setne       al  
00007FF663A5100F  imul        rcx,r9  
00007FF663A51013  mov         byte ptr [r8],al  
00007FF663A51016  mov         rax,rcx  

__umulh
00007FF663A51020  mov         rax,rdx  
00007FF663A51023  mul         rax,rcx  
00007FF663A51026  test        rdx,rdx  
00007FF663A51029  setne       cl  
00007FF663A5102C  mov         byte ptr [r8],cl  
00007FF663A5102F  ret  
```

jblazquez:
Added a new commit that fixes the clang-format issues and also adds ARM64/ARM64EC CI workflows. I wasn't able to test those workflows in my fork, even after enabling GitHub Actions, so they might not work just yet.

.github/workflows/main.yml
```diff
@@ -388,6 +388,24 @@ jobs:
...
+            toolchain: ""
+          - os: windows-2022
+            build-type: Debug
+            arch: ARM64
+            toolchain: ""
+          - os: windows-2022
+            build-type: Release
+            arch: ARM64EC
+            toolchain: ""
+            extra-cmake-flags: -DCMAKE_SYSTEM_VERSION="10.0.22621.0"
```
jblazquez, line 403: Using the Windows 11 SDK is required for CMake to build ARM64EC correctly: https://stackoverflow.com/a/77651884

src/snmalloc/ds_core/bits.h
mjp41: Thanks. Let's leave it as is. That is on the fast path for `calloc`, so code gen is important.

.github/workflows/main.yml
```diff
@@ -388,6 +388,24 @@ jobs:
             build-type: Debug
             arch: x64
             toolchain: ""
+          - os: windows-2022
+            build-type: Release
```
mjp41, line 392: I believe if you don't specify build type it will apply to both.

jblazquez:
As expected (since I was unable to test them) the new ARM64 CI workflows aren't working. I'll see if I can iterate on those in my fork, otherwise I'll add a few more commits here to fix them.

.github/workflows/main.yml
```diff
@@ -388,6 +388,24 @@ jobs:
...
+            arch: ARM64
+            toolchain: ""
+          - os: windows-2022
+            build-type: Debug
+            arch: ARM64
```
mjp41, line 397: I'm assuming the tests will fail to run. We need a `build-only` option in the matrix.  There is one above
https://github.com/microsoft/snmalloc/blob/b8e9e99cf096357d37c67d423ccb657652879ba1/.github/workflows/main.yml#L89

mjp41 (changes requested):

.github/workflows/main.yml
jblazquez: Thanks, that should be fixed in the latest commit. I addressed this on my latest commit. It looks like you have to manually allow the workflows to run again. I expect them to pass now.

mjp41:
> As expected (since I was unable to test them) the new ARM64 CI workflows aren't working. I'll see if I can iterate on those in my fork, otherwise I'll add a few more commits here to fix them.

If it's too painful just say, and I can fix tomorrow and push the changes onto your branch.

The Microsoft org on GitHub has very strict requirements for CI runs from new contributors, so sorry about the pain in getting it through CI.

.github/workflows/main.yml
jblazquez: I forgot to test whether removing `build-type` would work. I was following the pattern of the other explicitly included workflows:

```
          - os: windows-2022
            build-type: Release
            arch: Win32
            toolchain: ""
          - os: windows-2022
            build-type: Debug
            arch: Win32
            toolchain: ""
          - os: windows-2022
            build-type: Release
            arch: x64
            toolchain: ""
          - os: windows-2022
            build-type: Debug
            arch: x64
            toolchain: ""
```

mjp41 (approved):
LGTM. Looks like CI is going to pass. Great job. Thanks. I'll merge when CI passes.

jblazquez:
Thanks! By the way, I just run the test suite in a Windows 11 ARM64 VM and all tests passed.

PR #658, opened 2024-05-23
Remove unnecessary assertion from aligned_alloc

mjp41:
With [DR460](https://open-std.org/JTC1/SC22/WG14/www/docs/summary.htm#dr_460) the undefined behaviour for size not being a multiple of alignment was remove.

The [N2072](http://www.open-std.org/jtc1/sc22/wg14/www/docs/n2072.htm) allows for this to return memory.

So we can remove this assert and keep the same conditions as memalign.


mjp41:
See https://github.com/openenclave/openenclave/pull/4981#issuecomment-2125687166 for more discussion.

PR #659, opened 2024-05-24
msgpass benchmark and its refactoring dependencies

nwf-msr:
This is just enough of #637 to get the benchmark itself landed, but that also pulls in the freelist refactoring and the option of tweaking obfuscation keys (but not any usage of that internally).

It's probably worth getting someone with more cryptographic and/or red-teaming chops to think about the particular tweaking function, too.

### DIFF
--- a/src/test/perf/msgpass/msgpass.cc
+++ b/src/test/perf/msgpass/msgpass.cc
@@ -1,0 +1,302 @@
+/**
+ * A simulation of a message-passing application workload for snmalloc.
+ *
+ * - N_PRODUCER producer threads allocate and queue spans of messages randomly,
+ * - to N_CONSUMER consumer threads, which dequeue messages and free() them.
+ *
+ * Optionally, N_PROXY threads act as both producers and consumers, forwarding
+ * received messages back to another queue rather than freeing them.
+ */
+
+#include "test/opt.h"
+#include "test/setup.h"
+#include "test/usage.h"
+#include "test/xoroshiro.h"
+
+constexpr static bool be_chatty = false;
+
+#include <chrono>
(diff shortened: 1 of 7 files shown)


src/test/perf/msgpass/msgpass.cc
```diff
@@ -0,0 +1,302 @@
...
+  void* alloc(size_t sz)
+  {
+    return a.alloc(sz);
+  }
+  void dealloc(void* p)
+  {
+    a.dealloc(p);
+  }
+};
+#endif
```
mjp41, line 72: Could we just use malloc and free in the example and do something like

```C++
#ifndef SNMALLOC_PASS_THROUGH
#include <snmalloc/override/malloc.cc>
#endif
```

I have been wondering about doing this for more of the examples, so that the tests could be less dependent on snmalloc.

mjp41 (approved):
LGTM. I think it can be merged as is, and we can make the changes in a future PR.

mjp41 (approved):
LGTM. I think it can be merged as is, and we can make the changes in a future PR.

src/test/perf/msgpass/msgpass.cc
nwf-msr: Could do.  The ability to grab the thread allocator once per thread in the statically-`snmalloc` case seemed kind of nice (and you can get the dynamic behavior by using the malloc-passthrough and `LD_PRELOAD`-ing `snmallocshim`), but maybe it's not actually useful.

nwf-msr:
Matt makes the good point that tweaking just one side of the obfuscation feels less wrong than tweaking both, so I'll make that change and force push.

Issue #661, opened 2024-06-06
Configurable client meta-data


mjp41:
Various systems want meta-data per allocation

* Asan has a bit mask for whether allocations are live or not.
* MiraclePtr wants to store a reference count per object. 
* Memory tagging wants a tag colour per object
* Memory tagging with revocation wants to store colour at last revocation

This issue is to support that kind of use case with custom version of `snmalloc`.

# Design

`snmalloc` stores meta-data at the chunk granularity. For each 16KiB (configurable) of memory, there is a 16 byte entry in the pagemap.  This contains the follow things,

* A ptr to the owning allocator
* The sizeclass
* A ptr to additional meta-data
* A few bits to signify if this is in use by the front-end or backend of snmalloc

The third entry, the additional meta-data is used to store the free lists and other data structures associated with this chunk of memory.  It can be shared between chunks of memory if the slab is more than 16KiB. 

This issue proposes to allow the `Config` template parameter to specify additional meta-data that can be stored per allocation.  Due to wanting potentially sub-byte storage, e.g. 1-bit allocation map, or a 4-bit colour, we will provide a class in Config that controls this meta-data.

```C++
class ClientMetaDataProvider {
  static constexpr size_t required_size(size_t max_index);

  static constexpr void init(char* base, size_t max_index);

  static Tref get(char* base, size_t index);
};
```

Given a type like this then snmalloc would expose:

```C++
Tref snmalloc::ref_meta_data(void* ptr);
```

The client can decide what operations are permitted on `Tref`.  For example, we might use an array as:

```C++
template <typename T>
class ArrayMetaDataProvider {
  static constexpr size_t required_size(size_t max_index) {
    return sizeof(T) * max_index;
  }

  static constexpr void init(char* base, size_t max_index) {
    new (base) T[max_index];
  }

  static T& get(char* base, size_t index) {
    return reinterpret_cast<T[]>(base)[index];
  }  
};
```
or a bit field as
```C++
// Assuming char is a byte.
class BitRef {
  char* base;
  size_t index;
  
public:
  BitRef(char* base, size_t index): base(base), index(index) {} 

  bool get_bit() {
    return *(base + (index / 8)) >> (index % 8) & 1;
  }

  bool set_bit(bool bit) {
    //  Really should handle concurrency properly.
    char* elem = base + (index / 8);
    size_t shift = (index % 8);
    size_t mask = 1 << shift;
    if (bit)
     *elem = *elem | mask;
    else
[...]

PR #662, opened 2024-06-11
Configurable client meta-data

mjp41:
This addresses issue #661.

This provide a way to configure snmalloc to provide per object meta-data that is out of band.  This can be used to provide different mitigations on top of snmalloc, such as storing memory tags in a compressed form, or provide a miracle pointer like feature.


src/snmalloc/mem/corealloc.h
```diff
@@ -401,7 +405,7 @@ namespace snmalloc
         meta->node.remove();
 
         Config::Backend::dealloc_chunk(
-          get_backend_local_state(), *meta, p, size);
+          get_backend_local_state(), *meta, p, size, 0);
```
nwf-msr, line 408: Large objects always have zero metadata space?

src/snmalloc/mem/corealloc.h
```diff
@@ -792,11 +796,16 @@ namespace snmalloc
       message<1024>("small_alloc_slow rsize={} slab size={}", rsize, slab_size);
 #endif
 
+      // Calculate the extra bytes required to store the client meta-data.
+      size_t extra_bytes = BackendSlabMetadata::get_extra_bytes(sizeclass);
```
nwf-msr, line 800: Hm.  The reason we're having to do this math here in the frontend is because `Config::Backend::alloc_chunk` doesn't know (and isn't told) the sizeclass?  (The `dealloc_chunk` case above could presumably read out the sizeclass from the pagemap if we didn't want to tell it... but we could also probably just tell it... but in any case, just reading from the pagemap doesn't work for the allocate half anyway.)

Should we be passing in the sizeclass and having the backend do the `get_extra_bytes`?  (And maybe also `public_state()` and having the backend do the `from_small_class` too?)

nwf-msr (commented):
Two comments having glanced only at the first commit.  Sorry if they're obviated by later changes!

src/snmalloc/mem/metadata.h
```diff
@@ -685,7 +685,9 @@ namespace snmalloc
      */
     [[nodiscard]] SNMALLOC_FAST_PATH SlabMetadata* get_slab_metadata() const
     {
-      //SNMALLOC_ASSERT(get_remote() != nullptr);
+      // TODO Following assertion removed for client meta-data use case.
+      // Think about possible UAF scenarios.
+      // SNMALLOC_ASSERT(get_remote() != nullptr);
```
nwf-msr, line 690: The concern is that the client has retained a pointer to a free object for so long that the containing slab has been given back to the backend and said client is now asking for the metadata once associated with that free object?  At that point `meta` isn't even guaranteed to point to a metaslab, right?

src/snmalloc/mem/corealloc.h
mjp41: There is always one in the metaslab by default.  But I think that might be giving me UB.

src/snmalloc/mem/corealloc.h
mjp41: That makes sense I will move it.  What I did was for expediency in getting something working, and I should have revisited this part.

src/test/func/miracle_ptr/miracle_ptr.cc
```diff
@@ -0,0 +1,188 @@
+/**
+ * This file demonstrates how the snmalloc library could be implemented to
+ * provide a miracle pointer like feature.  This is not a hardened
```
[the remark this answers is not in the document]
mjp41: I've made it all in a namespace

src/snmalloc/mem/metadata.h
mjp41: Yeah, this could do with some thinking about. What we might do to prevent a UAF being escalated.

src/snmalloc/mem/corealloc.h
mjp41: Made this clearer, and added a comment when I refactored where the extra_bytes are calculated.

src/snmalloc/backend_helpers/commonconfig.h
```diff
@@ -98,12 +98,14 @@ namespace snmalloc
   struct NoClientMetaDataProvider
   {
     using StorageType = Empty;
+    using DataRef = Empty&;
+    using ConstDataRef = const Empty&;
```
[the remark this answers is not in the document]
mjp41: Thanks, changed.

src/snmalloc/mem/metadata.h
mjp41: https://github.com/microsoft/snmalloc/pull/662/commits/a93cd15d740f450b48fc770929242a332bf82a47

So this checks if it is a backend meta-data.  There is a gap between the check and following pointers that could potentially be exploited to get an interesting write gadget, but this is a short window, and require control over interleavings to make reliable.

nwf-msr (approved):

Issue #663, opened 2024-06-17
Segmentation Fault on Apple Silicon Asahi Linux


SchrodingerZhu:
```
30/87 Test #30: perf-contention-check ...............   Passed    0.90 sec
      Start 31: perf-external_pointer-check
31/87 Test #31: perf-external_pointer-check .........***Exception: SegFault  0.28 sec
      Start 32: perf-low_memory-check
32/87 Test #32: perf-low_memory-check ...............   Passed    0.00 sec
      Start 33: perf-memcpy-check
33/87 Test #33: perf-memcpy-check ...................***Exception: SegFault  0.08 sec
      Start 34: perf-msgpass-check
34/87 Test #34: perf-msgpass-check ..................***Exception: SegFault  0.15 sec
      Start 35: perf-singlethread-check
35/87 Test #35: perf-singlethread-check .............***Exception: SegFault  0.10 sec
      Start 36: perf-startup-check
36/87 Test #36: perf-startup-check ..................***Exception: SegFault  0.13 sec
      Start 37: func-bits-check
37/87 Test #37: func-bits-check .....................   Passed    0.00 sec
      Start 38: func-cheri-check
38/87 Test #38: func-cheri-check ....................   Passed    0.00 sec
      Start 39: func-client_meta-check
39/87 Test #39: func-client_meta-check ..............***Exception: SegFault  0.07 sec
      Start 40: func-domestication-check
40/87 Test #40: func-domestication-check ............***Exception: SegFault  0.06 sec
      Start 41: func-external_pagemap-check
41/87 Test #41: func-external_pagemap-check .........   Passed    0.00 sec
      Start 42: func-first_operation-check
42/87 Test #42: func-first_operation-check ..........***Exception: SegFault  0.06 sec
      Start 43: func-fixed_region-check
43/87 Test #43: func-fixed_region-check .............   Passed    0.06 sec
      Start 44: func-jemalloc-check
44/87 Test #44: func-jemalloc-check .................***Exception: SegFault  0.05 sec
      Start 45: func-malloc-check
45/87 Test #45: func-malloc-check ...................***Exception: SegFault  0.07 sec
      Start 46: func-memcpy-check
46/87 Test #46: func-memcpy-check ...................***Exception: SegFault  0.05 sec
      Start 47: func-memory-check
47/87 Test #47: func-memory-check ...................***Exception: SegFault  0.06 sec
      Start 48: func-memory_usage-check
[...]

PR #664, opened 2024-06-17
Provide option to CMake for Page Size

mjp41:
This uses the PAGESIZE constant from the unistd.h on POSIX. This should make the code more resilient to being compiled on platforms with different page sizes.

### DIFF
--- a/src/snmalloc/pal/pal_posix.h
+++ b/src/snmalloc/pal/pal_posix.h
@@ -130,7 +130,11 @@ namespace snmalloc
 #endif
       ;
 
+#ifdef PAGESIZE
+    static constexpr size_t page_size = PAGESIZE;
+#else
     static constexpr size_t page_size = Aal::smallest_page_size;
+#endif
 
     /**
      * Address bits are potentially mediated by some POSIX OSes, but generally
--- a/src/snmalloc/pal/pal_posix.h
+++ b/src/snmalloc/pal/pal_posix.h
@@ -8,6 +8,7 @@
 #endif
 #include <errno.h>
 #include <fcntl.h>
+#include <limits.h>
 #include <stdio.h>
 #include <string.h>
 #include <strings.h>
@@ -129,8 +130,11 @@ namespace snmalloc
       | Entropy
 #endif
       ;
-
-#ifdef PAGESIZE
+#ifdef SNMALLOC_PAGESIZE
+    static_assert(
+      bits::is_pow2(SNMALLOC_PAGESIZE), "Page size must be a power of 2");
+    static constexpr size_t page_size = SNMALLOC_PAGESIZE;
+#elif defined(PAGESIZE)
     static constexpr size_t page_size = PAGESIZE;
 #else
     static constexpr size_t page_size = Aal::smallest_page_size;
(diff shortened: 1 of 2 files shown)


mjp41:
Hopefully addresses #663

src/snmalloc/pal/pal_posix.h
```diff
@@ -130,7 +131,11 @@ namespace snmalloc
 #endif
       ;
 
+#ifdef PAGESIZE
+    static constexpr size_t page_size = PAGESIZE;
```
nwf-msr, line 135: This works on my PowerPC machine and passes CI, but I wonder if we want to do something like...

```suggestion
    static constexpr size_t page_size = max(Aal::smallest_page_size, PAGESIZE);
```

SchrodingerZhu:
will try this after work

SchrodingerZhu:
I confirm the fix

SchrodingerZhu:
Sorry for the noise. The fix is not working. Such macro is undefined on platforms without a fixed page size.

```
#ifdef PAGESIZE
    #warning "defined"
    static constexpr size_t page_size = max(Aal::smallest_page_size, PAGESIZE);
#else
    #warning "undefined"
#endif
```

Gives `undefined`.

Actually, the corresponding header says

```
/* We do not provide fixed values for

   ARG_MAX      Maximum length of argument to the `exec' function
                including environment data.

   ATEXIT_MAX   Maximum number of functions that may be registered
                with `atexit'.

   CHILD_MAX    Maximum number of simultaneous processes per real
                user ID.

   OPEN_MAX     Maximum number of files that one process can have open
                at anyone time.

   PAGESIZE
   PAGE_SIZE    Size of bytes of a page.

   PASS_MAX     Maximum number of significant bytes in a password.

   We only provide a fixed limit for

   IOV_MAX      Maximum number of `iovec' structures that one process has
                available for use with `readv' or writev'.

   if this is indeed fixed by the underlying system.
*/
```

If this is not the on the perf-critical path, I suggest use `_SC_PAGESIZE` approach instead. It is the most portable way. In libc on linux, this value is usually populated via aux vector and cached by libc. There is no syscall involved. If you are concern that `sysconf` has a super large jumptable, you can also cache it inside snmalloc.

SchrodingerZhu:
Inspected the code a little, it seems that a dynamic page demands too many changes and may impact negatively on the performance. Is it possible to allow `cmake` to run `sysconf` or let user override page size during configuration phase? All these can be set as additional options.

mjp41:
> Inspected the code a little, it seems that a dynamic page size demands too many changes and may impact negatively on the performance. Is it possible to allow `cmake` to run `sysconf` or let user override page size during configuration phase? All these can be set as additional options.

I would prefer the second.  The first might have issues with cross compiling, so I wouldn't want to always run `sysconf` at configure time. 

I think we could make the dynamic page size work, but it is more effort than I have time for, and would be pretty subtle to review the changes.

src/snmalloc/pal/pal_posix.h
mjp41: Sorry, I messed up the force push.  Put it back in now

SchrodingerZhu:
Yes. I can workaround the issue with this patch

PR #665, opened 2024-06-18
Startup speed

mjp41:
This improves the startup time for a heavily multi-threaded workload.

There are two improvements that are independent of each other:

1.  Buddy allocator has local cache.  The buddy allocator uses the pagemap as backing store.  This means that a lot of pages are faulted in at startup for the pagemap, when a large unaligned block is added.  This was causing a noticeable time in micro-profiling. The change in this PR makes there be a few entries at the top-level, so we only use the pagemap once a reasonable amount entries are required.
2. DODUMP/DONTDUMP on Linux were costing a noticeable amount of time, and causing a significant slow down.  This PR also exposes a new couple of optional operations on the PAL, and only applies these operations to the Pagemap.

### DIFF
--- a/src/snmalloc/backend_helpers/buddy.h
+++ b/src/snmalloc/backend_helpers/buddy.h
@@ -15,7 +15,13 @@ namespace snmalloc
   template<typename Rep, size_t MIN_SIZE_BITS, size_t MAX_SIZE_BITS>
   class Buddy
   {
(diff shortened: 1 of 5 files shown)


mjp41:
After private feedback from @vishalgupta97 and @nwf-msr I am going to put the lock in a separate PR.

nwf-msr (approved):
LGTM

PR #666, opened 2024-06-19
Implement MCS Combining lock

mjp41:
This is hybrid of Flat Combining and the MCS queue lock. It uses
the queue like the MCS queue lock, but each item additionally
contains a thunk to perform the body of the lock. This enables
other threads to perform the work that initially issued the request.

### DIFF
--- a/src/snmalloc/ds/combininglock.h
+++ b/src/snmalloc/ds/combininglock.h
@@ -1,0 +1,168 @@
+#pragma once
+
+#include "../aal/aal.h"
+#include "../pal/pal.h"
+
+#include <atomic>
+#include <functional>
+
+namespace snmalloc
+{
+  class CombineLockNode;
+  using CombiningLock = std::atomic<CombineLockNode*>;
+
+  /**
+   * @brief Combinations of MCS queue lock with Flat Combining
+   *
+   * Each element in the queue has a pointer to a work item.
+   * This means when under contention the thread holding the lock
+   * can perform the work.
+   *
+   * As the work items are arbitrary lambdas there are no simplifications
+   * for combining related work items.  I.e. original Flat Combining paper
+   * might sort a collection of inserts, and perform them in a single traversal.
+   *
+   * Note that, we should perhaps add a Futex/WakeOnAddress mode to improve
+   * performance in the contended case, rather than spinning.
+   */
+  class CombineLockNode
+  {
+    enum class LockStatus
+    {
+      // The work for this node has not been completed.
+      WAITING,
+
+      // The work for this thread has been completed, and it is not the
+      // last element in the queue.
+      DONE,
+
+      // The work for this thread has not been completed, and it is the
+      // head of the queue.
+      READY
+    };
+
+    // Status of the queue, set by the thread at the head of the queue,
+    // When it makes the thread for this node either the head of the queue
+    // or completes its work.
+    std::atomic<LockStatus> status{LockStatus::WAITING};
+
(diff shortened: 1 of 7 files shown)


mjp41:
I attempted to integrate `atomic::wait` and `notify_one`, but it went much slower.  So avoiding that for now.  I don't think we hold the lock long enough that the cost is worth it.

src/snmalloc/ds/combininglock.h
```diff
@@ -0,0 +1,215 @@
...
+          curr->set_status(LockStatus::DONE);
+          curr = n;
+          continue;
+        }
+
+        // This could be the end of the queue, attempt to close the
+        // queue.
+        auto curr_c = curr;
+        if (lock.head.compare_exchange_strong(
+              curr_c, nullptr, std::memory_order_acq_rel))
```
SchrodingerZhu, line 154: The failure order should be `Relaxed`

src/snmalloc/ds/combininglock.h
```diff
@@ -0,0 +1,215 @@
...
+      status.store(s, std::memory_order_release);
+    }
+
+  public:
+    constexpr CombineLockNode(void (*f)(CombineLockNode*)) : f_raw(f) {}
+
+    void attach(CombiningLock& lock)
+    {
+      // Test if no one is waiting
+      if (lock.head.load(std::memory_order_relaxed) == nullptr)
```
SchrodingerZhu, line 79: Maybe it is beneficial to separate this region out as fast path subject to be inlined. The remaining parts can be put into a `attach_slow` function.

mjp41: I don't think the this code is used here enough to warrant that, but maybe I should for future use.

src/snmalloc/ds/combininglock.h
mjp41: Good catch, thanks

src/snmalloc/ds/flaglock.h
```diff
@@ -133,4 +133,11 @@ namespace snmalloc
       lock.flag.store(false, std::memory_order_release);
     }
   };
+
+  template <typename F>
+  inline void with(FlagWord& lock, F&& f)
```
nwf-msr, line 138: Not `when`? ;)

src/snmalloc/ds/combininglock.h
```diff
@@ -0,0 +1,168 @@
...
+   * Each element in the queue has a pointer to a work item.
+   * This means when under contention the thread holding the lock
+   * can perform the work.
+   *
+   * As the work items are arbitrary lambdas there are no simplifications
+   * for combining related work items.  I.e. original Flat Combining paper
+   * might sort a collection of inserts, and perform them in a single traversal.
+   *
+   * Note that, we should perhaps add a Futex/WakeOnAddress mode to improve
+   * performance in the contended case, rather than spinning.
```
nwf-msr, line 34: Could perhaps templatize the class on `<bool notify = FALSE>` or such?

nwf-msr (approved):
One thought for the future, and one bit of humor, but this LGTM.  Took me a moment to understand again how the MCS lock's queue worked; one of these days it'll stick.

SchrodingerZhu:
Another interesting synchronization technique using on stack nodes is the HemLock from oracle. https://arxiv.org/pdf/2102.03863

mjp41:
Thanks for the link. I hadn't seen that algorithm.

PR #667, opened 2024-06-20
Remove the SNMALLOC_USE_CXX17 C preprocessor symbol

nwf-msr:
An attempt to simplify some `#ifdef` mazes (for `[[no_unique_address]]`) led me to want to test the compiler more directly, which meant that `SNMALLOC_USE_CXX17` was largely superfluous, so we can remove it in favor of testing pre-defined preprocessor symbols directly.

In a little more detail, the prior definitions were not sufficient to detect when `[[no_unique_address]]` would be accepted but ignored.  That made deciding when we could kick on #637 by default more... interesting.  With this, I hope to have a simpler time of things.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -184,7 +184,6 @@ endfunction()
 add_library(snmalloc INTERFACE)
 
 if(SNMALLOC_USE_CXX17)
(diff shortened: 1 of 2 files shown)


mjp41 (approved):
LGTM

PR #671, opened 2024-08-26
aligned_alloc: Permit alignment values smaller than sizeof(uintptr_t)

### LINKED ISSUE #668: aligned_alloc doesn't permit alignments less than `sizeof(uintptr_t)`
C11 says:

> The value of alignment shall be a valid alignment supported by the implementation and the value of size shall be an integral multiple of alignment.

where:

> Alignments are represented as values of the type size_t. Valid alignments include only those values returned by an _Alignof expression for fundamental types, plus an additional implementation-defined set of values, which may be empty. Every valid alignment value shall be a nonnegative integral power of two.

However, snmalloc checks for `alignment < sizeof(uintptr_t)` and returns an error if so: https://github.com/microsoft/snmalloc/blob/6af38acd94d401313b7237bb7c2df5e662e135cf/src/snmalloc/global/libc.h#L148-L152

bsdjhb:
posix_memalign() requires alignment values of at least sizeof(uintptr_t), but aligned_alloc() does not.  memalign() regressed to require larger alignment in commit
6cbc50fe2c255b5b47ea63f4cf759a6a20f6e7a6.

Fixes #668

### DIFF
--- a/src/snmalloc/global/libc.h
+++ b/src/snmalloc/global/libc.h
@@ -145,8 +145,7 @@ namespace snmalloc::libc
 
   inline void* memalign(size_t alignment, size_t size)
   {
-    if (SNMALLOC_UNLIKELY(
-          alignment < sizeof(uintptr_t) || !bits::is_pow2(alignment)))
+    if (SNMALLOC_UNLIKELY(alignment == 0 || !bits::is_pow2(alignment)))
     {
       return set_error(EINVAL);
     }


mjp41:
Thanks for doing this. If you have a chance to add a test to malloc.cc, that would be awesome, but if you don't I'll just merge tomorrow.

mjp41 (approved):

PR #672, opened 2024-09-04
0-length arrays in Buddy ranges

nwf-msr:
`Buddy` ranges can be instantiated with `MAX_SIZE_BITS == MIN_SIZE_BITS` (or, in principle, `<=`).  If this happens, the resulting class will contain 0-length arrays, and code that appears to attempt to access an element therein trips gcc's `-Warray-bounds`.

This PR contains two commits: one that adds a `static_assert` to ensure that `Buddy`'s arrays are always of positive size, and one possible way to fix the case I'm tripping over.  The latter works because the Buddy's `MIN_SIZE_BITS` is instantiated at `MIN_CHUNK_BITS`, and so the change herein then ensures that its `MAX_SIZE_BITS` is strictly greater than `MIN_CHUNK_BITS`.

A different, and possibly better, solution would be to drop the `LargeBuddyRange` from `CentralMetaRange` when `max_page_chunk_size_bits` is computed to be equal to `MIN_CHUNK_BITS`.

Please advise.

### DIFF
--- a/src/snmalloc/backend/meta_protected_range.h
+++ b/src/snmalloc/backend/meta_protected_range.h
@@ -75,11 +75,14 @@ namespace snmalloc
       CommitRange<PAL>,
       // In case of huge pages, we don't want to give each thread its own huge
       // page, so commit in the global range.
-      LargeBuddyRange<
-        max_page_chunk_size_bits,
-        max_page_chunk_size_bits,
-        Pagemap,
-        page_size_bits>,
+      std::conditional_t<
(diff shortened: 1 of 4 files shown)


nwf:
You OK to merge even though macos CI is still pending?

mjp41:
> You OK to merge even though macos CI is still pending?

Yeah. I think Mac os 11 support has been removed. I think we need to remove and possibly add the next version

PR #673, opened 2024-09-06
[src/snmalloc/ds_core/redblacktree.h] Make free-standing

### LINKED ISSUE #670: Is it possible to make `snmalloc` freestanding?
I know the build is actually free of STL linkage. However, snmalloc is not freestanding buildable because of the following:

https://github.com/microsoft/snmalloc/blob/6af38acd94d401313b7237bb7c2df5e662e135cf/src/snmalloc/ds_core/redblacktree.h#L6

Could we include the `<string>` header conditionally?

SamuelMarks:
Closes #670.

Taking a quick look:
```sh
$ rg -IFNth --trim '#include <' | sort -u
#include <array>
#include <atomic>
#include <chrono>
#include <climits>
#include <cstddef>
#include <cstdint>
#include <cstdlib>
#include <cstring>
#include <errno.h>
#include <fcntl.h>
#include <functional>
#include <iomanip>
#include <iostream>
#include <limits>
#include <new>
#include <sstream>
#include <stddef.h>
#include <stdio.h>
#include <string>
#include <string.h>
#include <strings.h>
#include <string_view>
#include <sys/mman.h>
#include <sys/uio.h>
#include <type_traits>
#include <unistd.h>
#include <utility>
```

I'm not sure how much of this can obviously be removed without looking really wonky in the C++ world.

Maintainers / @SchrodingerZhu - do you just want a C port of this?

### DIFF
--- a/src/snmalloc/ds_core/redblacktree.h
+++ b/src/snmalloc/ds_core/redblacktree.h
@@ -3,7 +3,7 @@
 #include <array>
 #include <cstddef>
 #include <cstdint>
-#include <string>
+#include <cstring>
 
 namespace snmalloc
 {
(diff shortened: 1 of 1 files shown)


nwf:
I think I'd prefer that the debugging code in `redblacktree.h`, which seems to be the sole consumer of `<string>`, as well as its `#include`, be made conditional on `__STDC_HOSTED__` being nonzero.  (See https://en.cppreference.com/w/cpp/freestanding).  @mjp41?

mjp41:
In this case this is just code that is executed by the test harness to unit test the RB tree.  I would sooner this be under and `#ifdef`, so that it isn't compiled with the library just the unit tests.

mjp41:
I have redone your idea, but with stack allocation in #680, and also copied the original contents into the string.

I would sooner not take a dependence on malloc/free in this code.  The std::string did have a dependence on an allocator, which is not nice for allocator debug code.

PR #674, opened 2024-09-10
Rework CI

nwf-msr:


.github/workflows/main.yml
```diff
@@ -94,7 +79,52 @@ jobs:
...
+    steps:
+    - uses: actions/checkout@v3
+    - name: Install build dependencies
+            # The homebrew packages are broken at the moment and error out
+            # after trying to install Python as a dependency of ninja because
+            # 2to3 exists.  As a quick hack, delete it first.  This should be
+            # removed once the homebrew install is fixed.
+      run: "rm -f /usr/local/bin/2to3 ; brew update && brew install ninja"
     
     - uses: mjp41/workaround8649@c8550b715ccdc17f89c8d5c28d7a48eeff9c94a8
```
mjp41, line 129: I don't think this is needed on Mac OS

mjp41 (approved):
LGTM

.github/workflows/main.yml
nwf-msr: OK, let's give that a shot. :)

nwf-msr: Looks like you're right. :)

PR #677, opened 2024-09-18
WIP: BatchIt

nwf:
This recreates the current contents of #637 from a different repository, since the lifetime of the @nwf-msr account is now pretty limited.  (I couldn't find a way to move _branches_ between repos in a way that swung the PR, so had to settle for this low-tech solution.)

### DIFF
--- a/src/snmalloc/ds/allocconfig.h
+++ b/src/snmalloc/ds/allocconfig.h
@@ -43,7 +43,7 @@ namespace snmalloc
 #if defined(SNMALLOC_MIN_ALLOC_SIZE)
     SNMALLOC_MIN_ALLOC_SIZE;
 #else
-    sizeof(void*) * (mitigations(freelist_backward_edge) ? 4 : 2);
+    2 * sizeof(void*);
 #endif
 
   // Minimum slab size.
@@ -119,6 +119,45 @@ namespace snmalloc
   static constexpr size_t REMOTE_SLOT_BITS = 8;
   static constexpr size_t REMOTE_SLOTS = 1 << REMOTE_SLOT_BITS;
   static constexpr size_t REMOTE_MASK = REMOTE_SLOTS - 1;
+
+#if defined(SNMALLOC_DEALLOC_BATCH_RING_ASSOC)
+  static constexpr size_t DEALLOC_BATCH_RING_ASSOC =
+    SNMALLOC_DEALLOC_BATCH_RING_ASSOC;
+#else
+#  if defined(__has_cpp_attribute)
+#    if ( \
+      __has_cpp_attribute(msvc::no_unique_address) && \
+      (__cplusplus >= 201803L || _MSVC_LANG >= 201803L)) || \
+      __has_cpp_attribute(no_unique_address)
+  // For C++20 or later, we do have [[no_unique_address]] and so can also do
+  // batching if we aren't turning on the backward-pointer mitigations
+  static constexpr size_t DEALLOC_BATCH_MIN_ALLOC_WORDS =
+    mitigations(freelist_backward_edge) ? 4 : 2;
+#    else
+  // For C++17, we don't have [[no_unique_address]] and so we always end up
+  // needing all four pointers' worth of space (because BatchedRemoteMessage has
+  // two freelist::Object::T<> links within, each of which will have two fields
+  // and will be padded to two pointers).
+  static constexpr size_t DEALLOC_BATCH_MIN_ALLOC_WORDS = 4;
+#    endif
+#  else
+  // If we don't even have the feature test macro, we're C++17 or earlier.
+  static constexpr size_t DEALLOC_BATCH_MIN_ALLOC_WORDS = 4;
+#  endif
+
+  static constexpr size_t DEALLOC_BATCH_RING_ASSOC =
+    (MIN_ALLOC_SIZE >= (DEALLOC_BATCH_MIN_ALLOC_WORDS * sizeof(void*))) ? 2 : 0;
+#endif
+
+#if defined(SNMALLOC_DEALLOC_BATCH_RING_SET_BITS)
+  static constexpr size_t DEALLOC_BATCH_RING_SET_BITS =
+    SNMALLOC_DEALLOC_BATCH_RING_SET_BITS;
+#else
+  static constexpr size_t DEALLOC_BATCH_RING_SET_BITS = 3;
+#endif
+
+  static constexpr size_t DEALLOC_BATCH_RINGS =
+    DEALLOC_BATCH_RING_ASSOC * bits::one_at_bit(DEALLOC_BATCH_RING_SET_BITS);
 
   static_assert(
     INTERMEDIATE_BITS < MIN_ALLOC_STEP_BITS,
--- a/src/snmalloc/ds/allocconfig.h
+++ b/src/snmalloc/ds/allocconfig.h
@@ -43,7 +43,7 @@ namespace snmalloc
 #if defined(SNMALLOC_MIN_ALLOC_SIZE)
     SNMALLOC_MIN_ALLOC_SIZE;
 #else
-    2 * sizeof(void*);
+    sizeof(void*) * (mitigations(freelist_backward_edge) ? 4 : 2);
 #endif
 
   // Minimum slab size.
--- a/src/snmalloc/mem/corealloc.h
+++ b/src/snmalloc/mem/corealloc.h
@@ -513,13 +513,11 @@ namespace snmalloc
                            SNMALLOC_FAST_PATH_LAMBDA {
                              return capptr_domesticate<Config>(local_state, p);
                            };
-      auto cb = [this, &need_post](
+      auto cb = [this, domesticate, &need_post](
                   capptr::Alloc<RemoteMessage> msg) SNMALLOC_FAST_PATH_LAMBDA {
         auto& entry =
           Config::Backend::template get_metaentry(snmalloc::address_cast(msg));
-
-        handle_dealloc_remote(entry, msg, need_post);
-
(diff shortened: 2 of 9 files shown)


src/snmalloc/mem/remoteallocator.h
```diff
@@ -1,10 +1,271 @@
...
+      (sizeof(BatchedRemoteMessage) <= MIN_ALLOC_SIZE));
+    static_assert(offsetof(BatchedRemoteMessage, free_ring) == 0);
+
+    static_assert(
+      (DEALLOC_BATCH_RINGS == 0) ||
+        (MAX_SLAB_SPAN_BITS + MAX_CAPACITY_BITS < 8 * sizeof(void*)),
+      "Ring bit-stuffing trick can't reach far enough to enclose a slab");
+  };
+
+  class SingletonRemoteMessage
```
mjp41, line 200: Add a comment on use of this class for cases where we don't batch messages.

src/snmalloc/mem/remotecache.h
```diff
@@ -12,6 +12,164 @@
...
+
+    SNMALLOC_FAST_PATH size_t
+    ring_set(typename Config::PagemapEntry::SlabMetadata* meta)
+    {
+      // See https://github.com/skeeto/hash-prospector for choice of constant
+      return ((meta->as_key_tweak() * 0x7EFB352D) >> 16) &
+        bits::mask_bits(DEALLOC_BATCH_RING_SET_BITS);
+    }
+
+    template<typename Forward>
```
mjp41, line 34: Perhaps explain use of `Forward`?  Basically continuation?

src/snmalloc/mem/remotecache.h
```diff
@@ -12,6 +12,164 @@
...
+            }
+          }
+          else
+          {
+            UNUSED(entropy);
+          }
+          return;
+        }
+      }
+
```
mjp41, line 93:
```suggestion

      // No hit in Cache, so find a victim to evict.
```

mjp41 (changes requested):
This is great stuff, @nwf-msr and @nwf.  A few minor things, but it is really clean.  Thanks for making it so easy to review.

src/snmalloc/ds/allocconfig.h
```diff
@@ -43,7 +43,7 @@ namespace snmalloc
 #if defined(SNMALLOC_MIN_ALLOC_SIZE)
     SNMALLOC_MIN_ALLOC_SIZE;
 #else
-    2 * sizeof(void*);
+    sizeof(void*) * (mitigations(freelist_backward_edge) ? 4 : 2);
```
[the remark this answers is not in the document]
nwf-msr: Ah, this is probably stale.  Good catch.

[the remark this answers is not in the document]
nwf-msr: Added revert of this hunk to "Make BatchIt optional"

src/snmalloc/mem/corealloc.h
```diff
@@ -548,31 +546,52 @@ namespace snmalloc
...
+              unreturned = meta->return_objects(unreturned.batch_size());
+              if (unreturned.needs_work())
+                dealloc_local_object_slower(entry, meta);
+            } while (unreturned.batch_size() > 0);
+          }
+        }
       }
       else
       {
+        auto nelem = RemoteMessage::template ring_size<Config>(
```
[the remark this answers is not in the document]
nwf: Fixed in "corealloc: operate ring-at-a-time on remote queues" and preserved in rebased "Initial CHERI support for free rings".

src/snmalloc/mem/remoteallocator.h
```diff
@@ -1,10 +1,271 @@
...
+      SNMALLOC_ASSERT(size < bits::one_at_bit(MAX_CAPACITY_BITS));
+
+      auto [first, last] = flb.extract_segment(key, key_tweak);
+
+      // Preserve the last node's backpointer and change its type.
+      auto last_prev = last->prev;
+      auto self =
+        CapPtr<BatchedRemoteMessage, capptr::bounds::Alloc>::unsafe_from(
+          new (last.unsafe_ptr()) BatchedRemoteMessage());
+      self->free_ring.prev = last_prev;
```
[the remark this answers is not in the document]
nwf: Comment put in "Initial steps in batched remote messages" which introduced this excitement.

src/snmalloc/mem/remoteallocator.h
nwf: In "Make BatchIt optional"

src/snmalloc/mem/remotecache.h
nwf: Comment added to "Initial dealloc-side batching machinery"

nwf:
Pushing with some changes applied to get CI's opinion.  Will address the rest of your review comments soon.

src/snmalloc/mem/corealloc.h
```diff
@@ -548,31 +546,52 @@ namespace snmalloc
...
 
       if (SNMALLOC_LIKELY(entry.get_remote() == public_state()))
       {
-        dealloc_local_object(p, entry);
-        return;
+        auto unreturned =
+          dealloc_local_objects_fast(entry, msg, entropy, domesticate);
+        if (SNMALLOC_UNLIKELY(unreturned.needs_work()))
+        {
+          dealloc_local_object_slow(msg.as_void(), entry);
```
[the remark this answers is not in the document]
nwf: `dealloc_local_object_slow` might actually release a large object, but is otherwise a call to `dealloc_local_object_slower`, which, as you say, just updates metadata.  Maybe `dealloc_local_object_slow_meta` for the latter instead?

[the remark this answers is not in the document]
mjp41: That makes sense. It just felt in first inspection that the batch was added back in parts, rather than is all added back in one go, and then we might have to do several updates to get the meta data in the correct state.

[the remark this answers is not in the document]
nwf: Yeah, you're right.  I've renamed the function (in a new commit at the bottom of the series) and will expand commentary on the later-introduced call site as part of the refactoring for your other comments. :)

src/snmalloc/mem/metadata.h
```diff
@@ -500,6 +500,55 @@ namespace snmalloc
...
+      if (n >= needed())
+      {
+        n -= needed();
+        needed() = 0;
+        return ReturnObjectsResult(n);
+      }
+
+      needed() -= n;
+      return ReturnObjectsResult();
+    }
```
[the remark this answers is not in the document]
nwf: It's a little more messy than that, for reasons to be explained in comments. :)

nwf:
OK, I think that's all the review comments.

Also, huh, I guess my time at MS using this account (what a mistake) reset the CLA bot, which I appeased all the way back in https://github.com/microsoft/snmalloc/pull/41 (how time _flies_!). 
 I _think_ the right thing to say here is:

nwf:
I'm unsure what I could have done to make Windows unhappy, but that suggest I've broken something subtle. :(

src/snmalloc/mem/metadata.h
```diff
@@ -500,6 +500,58 @@ namespace snmalloc
...
+        }
+
+        /*
+         * If we've already taken a step and there are no more unreturned
+         * objects, then we don't need to take another step.  It can happen
+         * that we get here because _meta->needed() is still 0, tho', so
+         * bail.
+         */
+        if (!first && _batch == 0)
+          return false;
```
mjp41, line 532: > I'm unsure what I could have done to make Windows unhappy, but that suggest I've broken something subtle. :(
> 
> I no longer have a Windows machine on which to test things; can I ask you to have a look, @mjp41?

I haven't tried on Windows yet.  But just inspecting the CI, I am wondering if it is valid to read needed after we have performed the last meta data update.  Could another thread have reused the meta-data by this point or turned the page behind the meta-data off?  Perhaps, we need to use the batch size first in the cases where we can.  Which is effectively what your original code did as it would only look another time if it was non-zero.
```suggestion
        SNMALLOC_ASSERT(first && (batch != 0));

        /*
         * If we've already taken a step and there are no more unreturned
         * objects, then we don't need to take another step.  It can happen
         * that we get here because _meta->needed() is still 0, tho', so
         * bail.
         */
        if (!first && _batch == 0)
          return false;
        
        if (SNMALLOC_LIKELY(_batch < _meta->needed()))
        {
          // Will not hit threshold for state transition
          _meta->needed() -= _batch;
          return false;
        }
```

mjp41: FYI, updated the suggestion above to fix my failure to do basic logic.

nwf: Ah, that's a good point.  Lemme apply that fix.

nwf: I don't think that `SNMALLOC_ASSERT` is right, though.  The thing the new condition is guarding against is precisely the case that we have taken the first step (and enough subsequent steps) to have returned all objects.

nwf:
Looks like that was it!

src/snmalloc/mem/metadata.h
mjp41: Sorry, I'm a bit confused why the first one would be empty?  Perhaps, applying de-Morgan to the assert would be helpful:
```C++
   SNMALLOC_ASSERT(!(first && batch == 0));
```

nwf: Indeed, I just couldn't read.  I'll add the assert.

mjp41 (approved):

mjp41:
@nwf want to click the merge button?

PR #678, opened 2024-09-21
Further gcc -Werror=array-bounds fix

nwf:
In test/perf/startup, gcc (Debian 14.2.0-3) seems to get confused about the size of the counters vector as the code was written.  Rewrite the code to pass the same value (`std::thread::hardware_concurrency()`, but in a local) to both `counters.resize()` and the `ParallelTest` ctor.

### DIFF
--- a/src/test/perf/startup/startup.cc
+++ b/src/test/perf/startup/startup.cc
@@ -70,7 +70,8 @@ public:
 
 int main()
 {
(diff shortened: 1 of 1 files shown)


mjp41 (approved):
Not sure how that fixes GCC warning, but the code looks the same, so looks fine to me.

PR #680, opened 2024-09-24
Remove some unneeded headers

mjp41:
Towards #670.

### DIFF
--- a/src/snmalloc/ds_core/redblacktree.h
+++ b/src/snmalloc/ds_core/redblacktree.h
@@ -3,7 +3,6 @@
 #include <array>
 #include <cstddef>
 #include <cstdint>
-#include <string>
 
 namespace snmalloc
 {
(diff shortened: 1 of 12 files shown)


src/snmalloc/ds_core/redblacktree.h
```diff
@@ -439,9 +438,19 @@ namespace snmalloc
           depth);
         if (!(get_dir(true, curr).is_null() && get_dir(false, curr).is_null()))
         {
-          auto s_indent = std::string(indent);
-          print(get_dir(true, curr), (s_indent + "|").c_str(), depth + 1);
-          print(get_dir(false, curr), (s_indent + " ").c_str(), depth + 1);
+          char s_indent[128];
```
SamuelMarks, line 441: Why 128? - Also you might want to use a scoped constant, like `enum { size = 128 }; char s_indent[size]`

mjp41: @samuelMarks thanks for your code review.  I've added a comment to explain the bound, and made a `constexpr` for the constant.

PR #684, opened 2024-10-19
PAL::haiku finally supports getentropy.

devnexen:
Also fixing the build with PAL::posix assuming the intent is to avoid using std namespace for max but rather using the in-house implementation.


mjp41 (approved):
LGTM

mjp41:
Sorry, clangformat is complaining.

PR #691, opened 2024-11-19
clean up Result and make Result move-only

SchrodingerZhu:
Sorry, I failed to catch this in my last patch. The random walk can easily run into OOM due this leaked memory resource.


mjp41 (approved):
LGTM

PR #695, opened 2024-11-24
avoid chrono inclusion when clock_gettime is available

SchrodingerZhu:
Avoid pulling in non freestanding chrono dependency if we can use `clock_gettime`.


SchrodingerZhu:
FYI, I did not use `timespec_get` because we do not have the control of monotonicity.

mjp41 (approved):
LGTM

PR #699, opened 2024-12-01
cmake: look for rdynamic instead of hardcoding it

jcelerier:
e.g. it's not available on MinGW's GCC


devnexen (approved):
LCTM

mjp41 (approved):

mjp41:
Thanks for doing this. We have had problems with MingW in the past, so I'm not sure if you will encounter other problems. @SchrodingerZhu might be able to say more as he previously investigated MingW.

SchrodingerZhu:
For MinGW, clang UCRT is now officially a part of it. I think we can push it forward if anyone have bandwidth working on it

PR #703, opened 2024-12-11
remove string_view STL dependency

SchrodingerZhu:


SchrodingerZhu:
The CI failures are due to macos-12 being obsolete. We should probably update the images

mjp41:
I've just pushed a PR to move the CI forward to newer versions.  Hopefully will be smooth.

mjp41:
Rebasing should fix the CI.

src/snmalloc/pal/pal.h
```diff
@@ -159,9 +159,7 @@ namespace snmalloc
    * The following are supported as arguments:
    *
    *  - Characters (`char`), printed verbatim.
-   *  - Strings (anything convertible to `std::string_view`), typically string
```
mjp41, line 162: As we've lost the other types that can flow into this, I am left wondering if we actually use this append function?

mjp41: As I guess I'm wondering if we should inline the functionality into append?

SchrodingerZhu: Sorry, was busy on other stuffs. Inling should work except the two definitions will run into ambiguity.

mjp41 (approved):
LGTM. Thanks for stressing the earlier review.

Issue #701, opened 2024-12-11
Sets `_HAS_EXCEPTIONS=0` globally


Kyuuhachi:
snmalloc has `target_compile_definitions(snmalloc INTERFACE -D_HAS_EXCEPTIONS=0)`, which seems to set that flag on all downstream consumers. This causes some rather strange issues in programs that use exceptions. Is this intentional? 

https://github.com/microsoft/snmalloc/blob/d6ad197d3c8452cd4d2548813e0b88a333ce11d6/CMakeLists.txt#L297

mjp41:
I think that should be fixed.  I think it was so the various libraries we build do not have to support exceptions, and thus can be more efficiently compiled.  We should move that to just apply to the built libraries.

Probably moving it here would make sense:
https://github.com/microsoft/snmalloc/blob/d6ad197d3c8452cd4d2548813e0b88a333ce11d6/CMakeLists.txt#L431


@davidchisnall sounds sensible to you?

mjp41:
Closed by #707.

PR #707, opened 2024-12-11
Fixes: Sets `_HAS_EXCEPTIONS=0` globally #701

mjp41:
This makes the `_HAS_EXCEPTIONS` flag on windows only apply to libraries, and not if the library is used header only.


mjp41:
@kyuuhachi I think this should address #701.  Could you let me know if this PR works for your use case.  Thanks

Kyuuhachi:
Yep, seems to work :+1:

PR #712, opened 2024-12-15
remove custom ModArray type attempt.

devnexen:


mjp41:
I'm not sure about this.  The ModArray means that if we end up with a corrupted sizeclass being passed in we won't get an out-of-bounds read.  I am not sure hpw dangerous that is, but it only leads to a single mask on the fast path.

@devnexen is this for the general simplification of the code base?  Or for performance of that mask on the fast path?

devnexen:
I admit much more the former.

PR #713, opened 2024-12-15
[ds_core] simplify MessageBuilder construction

SchrodingerZhu:


src/snmalloc/ds_core/helpers.h
```diff
@@ -361,28 +339,46 @@ namespace snmalloc
...
-  public:
     /**
-     * Constructor.  Takes a format string and the arguments to output.
+     * Append with format string and arguments. Compiler
+     * is able to optimize the recursion into loops.
      */
-    template<typename... Args>
-    SNMALLOC_FAST_PATH MessageBuilder(const char* fmt, Args... args)
+    template<typename Head, typename... Tail>
+    void append_args(const char* fmt, Head&& head, Tail&&... tail)
```
mjp41, line 347: If you added 
```C++
    void append_args(const char* fmt)
    {
      append(fmt);
    }
```

Then you could use overloading to remove the
```C++
        if constexpr (sizeof...(Tail) == 0)
        {
          append(fmt + 2);
        }
        else
        {
          append_args(fmt + 2, std::forward<Tail>(tail)...);
        }
```
and just replace it with
```C++
        append_args(fmt + 2, std::forward<Tail>(tail)...);
```

src/snmalloc/ds_core/helpers.h
```diff
@@ -361,28 +339,46 @@ namespace snmalloc
...
     {
-      buffer[SafeLength] = 0;
-      size_t arg = 0;
-      auto args_tuple = std::forward_as_tuple(args...);
-      for (const char* s = fmt; *s != 0; ++s)
+      if (*fmt == 0)
       {
-        if (s[0] == '{' && s[1] == '}')
+        return;
```
mjp41, line 351: Is this really an error?  We have unused arguments passed to the message builder.

src/snmalloc/ds_core/helpers.h
mjp41: I also wonder if you could drop the special case:
```C++
    SNMALLOC_FAST_PATH MessageBuilder(const char* fmt)
    {
      buffer[SafeLength] = 0;
      for (const char* s = fmt; *s != 0; ++s)
      {
        append_char(*s);
      }
      append_char('\0');
    }
```
with this change?

SchrodingerZhu:
It is strange that the linker is broken on macos-14. Do you have any insights?

src/snmalloc/ds_core/helpers.h
```diff
@@ -361,43 +339,33 @@ namespace snmalloc
...
-    template<typename... Args>
-    SNMALLOC_FAST_PATH MessageBuilder(const char* fmt, Args... args)
+    template<typename Head, typename... Tail>
+    void append(const char* fmt, Head&& head, Tail&&... tail)
     {
-      buffer[SafeLength] = 0;
-      size_t arg = 0;
-      auto args_tuple = std::forward_as_tuple(args...);
-      for (const char* s = fmt; *s != 0; ++s)
+      if (fmt[0] == '{' && fmt[1] == '}')
```
mjp41, line 349: I think we still need the case of

```C++
if (fmt[0] == 0)
  error("Internal error: format string missing {}!");
```

Otherwise the code will OOB read if the {} is missing.

mjp41 (changes requested):
This looks like a nice simplification

mjp41:
> It is strange that the linker is broken on macos-14. Do you have any insights?

Sorry no idea.

mjp41 (approved):

mjp41:
Thanks for fixing the linker crash, and addressing my comments.  Looks great.

PR #715, opened 2024-12-18
[proxy](1/n) start self-vendored STL with atomic

SchrodingerZhu:
This patch starts to vendor functionalities from STL using our own code.
It adds a `SNMALLOC_USE_SELF_VENDORED_STL` option which means `snmalloc` will try to build its core header library without `c++` STL.

The way we do this is to introduce a `proxy` which dispatches underlying implementations based on `SNMALLOC_USE_SELF_VENDORED_STL`. It is allowed to use STL in tests even when `SNMALLOC_USE_SELF_VENDORED_STL=On`.

This patch only includes changes related to `std::atomic`. The `snmalloc::Atomic` is simple wrapper of compiler intrinsics that only support trivial types with unique object representations.

### DIFF
--- a/docs/combininglock.md
+++ b/docs/combininglock.md
@@ -94,14 +94,14 @@ Let us provide an MCS queue lock for our API given above:
 ```C++
 struct LockNode
 {
-  std::atomic<LockNode*> next{nullptr};
-  std::atomic<bool> available{false};
+  proxy::Atomic<LockNode*> next{nullptr};
+  proxy::Atomic<bool> available{false};
 };
 
 struct MCSLock
 {
   // End of queue
-  std::atomic<LockNode*> last;
+  proxy::Atomic<LockNode*> last;
 };
 
 template<typename F>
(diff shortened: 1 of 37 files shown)


SchrodingerZhu:
The NetBSD's unsecure mirror is not working. not sure how to fix it

docs/combininglock.md
```diff
@@ -94,14 +94,14 @@ Let us provide an MCS queue lock for our API given above:
 ```C++
 struct LockNode
 {
-  std::atomic<LockNode*> next{nullptr};
```
mjp41, line 97: I would suggest we leave this file untouched as it is for understanding purposes, and the proxy will only serve to complicate the presentation.

docs/release/0.7/README.md
```diff
@@ -112,7 +112,7 @@ This allows developers to build new security features on top of snmalloc.
 For instance, building snmalloc with the following definition of `Alloc` will allow you to store a 64-bit counter for each allocation:
 ```cpp
   using Alloc = snmalloc::LocalAllocator<snmalloc::StandardConfigClientMeta<
-    ArrayClientMetaDataProvider<std::atomic<size_t>>>>;
+    ArrayClientMetaDataProvider<proxy::Atomic<size_t>>>>;
```
mjp41, line 115: Similarly in this file.

mjp41 (changes requested):
Thanks for doing this.  I am unsure on the name `proxy` it is revealing implementation details rather than saying what it is.  I wonder if 
```C++
snmalloc::std::Atomic
```
would be better.  It is the standard library that we should use from snmalloc.


I also wonder if it would be better to have a structure roughly like

```
snmalloc/
  std/
    README.md
    atomic.h
    ...
    cxx/
      README.md
      atomic.h
      ...
    llvm/
      README.md
      atomic.h
      ...
```
where 
```
snmalloc/std/atomic.h
```
Contains an `ifdef` to the include the correct `atomic.h`.

Sorry for not engaging sooner I took some time off.

SchrodingerZhu:
@mjp47 All reviews are addressed in the latest commit, with some changes:

1. I used `stl` instead of `std` to avoid conflict in name look up. If `std` exists inside `snmalloc`, C++ cannot resolve the namespace to `::std` by default.
2. I changed `llvm` to `gnu` since the language extensions are from GCC, although Clang also defines all the builtin functions.

mjp41 (approved):
LGTM thanks for addressing the feedback

PR #717, opened 2024-12-19
[proxy](2/n) provide proxy layers for type traits

SchrodingerZhu:
On top of #715, proxy layer for type traits

PR #719, opened 2024-12-21
[proxy](4/n) use customised placement-new

SchrodingerZhu:
Stacked on

- #715 
- #717 
- #718


src/snmalloc/proxy/new.h
```diff
@@ -0,0 +1,35 @@
...
+#include "snmalloc/ds_core/defines.h"
+
+namespace snmalloc
+{
+
+  // This is used in both vendored and non-vendored mode.
+  struct PlacementToken
+  {};
+
+  inline constexpr PlacementToken placement_token{};
```
mjp41, line 12: Why is this needed?  I guess it would be nice to just define
```C++
SNMALLOC_FAST_PATH_INLINE void*
operator new(size_t, void* ptr) noexcept
{
  return ptr;
}
```
But I assume that fails?

SchrodingerZhu: Placement new operator has to appear at global namespace. Consider using `snmalloc` as a header library with other `c++` programs, if we provide such definitions, then it is likely to conflict with `<new>` or other's definitions.

mjp41 (approved):

PR #720, opened 2024-12-21
[proxy](5/n) use c headers instead of c++ headers

SchrodingerZhu:
Stacked on:

- #715 
- #717 
- #718 
- #719

PR #721, opened 2024-12-21
[proxy](6/n) self-vendor utility headers

SchrodingerZhu:
Stacked on #720

### DIFF
--- a/src/snmalloc/backend/fixedglobalconfig.h
+++ b/src/snmalloc/backend/fixedglobalconfig.h
@@ -1,7 +1,10 @@
 #pragma once
 
 #include "../backend_helpers/backend_helpers.h"
+#include "snmalloc/stl/type_traits.h"
 #include "standard_range.h"
+
+#include <cstddef>
 
 namespace snmalloc
 {
--- a/src/snmalloc/backend/fixedglobalconfig.h
+++ b/src/snmalloc/backend/fixedglobalconfig.h
@@ -4,7 +4,7 @@
 #include "snmalloc/stl/type_traits.h"
 #include "standard_range.h"
 
-#include <cstddef>
+#include <stddef.h>
 
 namespace snmalloc
 {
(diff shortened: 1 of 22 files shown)


src/snmalloc/backend/fixedglobalconfig.h
```diff
@@ -117,8 +120,16 @@ namespace snmalloc
     {
       static_assert(B::wildness == capptr::dimension::Wildness::Wild);
 
-      static const size_t sz = sizeof(
-        std::conditional<std::is_same_v<std::remove_cv<T>, void>, void*, T>);
```
SchrodingerZhu, line 121: I guess this was written wrongly.

mjp41: I am not sure my spot the difference skills are good enough.  What is wrong?

SchrodingerZhu: conditional is an empty type I suppose?

src/snmalloc/backend/fixedglobalconfig.h
```diff
@@ -117,8 +120,16 @@ namespace snmalloc
...
+      constexpr size_t sz = []() constexpr {
+        if constexpr (proxy::is_same_v<proxy::remove_cv_t<T>, void>)
+        {
+          return sizeof(void*);
+        }
+        else
+        {
+          return sizeof(T);
+        }
+      }();
```
mjp41, line 133:
```suggestion
      static const size_t sz = sizeof(
        proxy::conditional_t<proxy::is_same_v<proxy::remove_cv<T>, void>, void*, T>);
```
Ah, I see, so shouldn't this really be a `conditional_t`, and then the rewrite is much smaller?

SchrodingerZhu: I can have a try but I remember void cannot appear at the given position.

mjp41: Thanks

SchrodingerZhu: It actually works. It did not work previously because it was using `remove_cv` instead of `remove_cv_t`.

I have promoted this fix to the second patch of the serious as it is more related to type traits.

mjp41 (approved):
LGTM

PR #722, opened 2024-12-21
[proxy](7/n) self vendor array and algorithm headers

SchrodingerZhu:
This patch clears all STL dependency in the core header library.
Stacked on #721


src/snmalloc/proxy/algorithm.h
```diff
@@ -0,0 +1,53 @@
...
+    constexpr SNMALLOC_FAST_PATH const T& max(const T& a, const T& b)
+    {
+      return a < b ? b : a;
+    }
+
+    template<typename T>
+    constexpr SNMALLOC_FAST_PATH const T& min(const T& a, const T& b)
+    {
+      return a < b ? a : b;
+    }
```
mjp41, line 29: Theses are also defined in `bits.h` could we simplify now?

src/snmalloc/global/memcpy.h
```diff
@@ -156,7 +157,7 @@ namespace snmalloc
      */
     SNMALLOC_UNUSED_FUNCTION
     static constexpr size_t LargestRegisterSize =
-      std::max(sizeof(uint64_t), sizeof(void*));
+      proxy::max(sizeof(uint64_t), sizeof(void*));
```
mjp41, line 160: Could this use, then we wouldn't need to vendor std::max/std::min?
```suggestion
      bits::max(sizeof(uint64_t), sizeof(void*));
```

CMakeLists.txt
```diff
@@ -205,6 +206,12 @@ else()
   target_compile_definitions(snmalloc INTERFACE SNMALLOC_USE_WAIT_ON_ADDRESS=0)
 endif()
 
+if(SNMALLOC_USE_SELF_VENDORED_STL)
+  target_compile_definitions(snmalloc INTERFACE SNMALLOC_USE_SELF_VENDORED_STL=1)
+else()
+  target_compile_definitions(snmalloc INTERFACE SNMALLOC_USE_SELF_VENDORED_STL=0)
+endif()
+
```
mjp41, line 214: I've noticed as a style you tend to use 1/0 to signify a boolean, but you can do his with defined and not defined.  The advantage of defined and not defined is that it doesn't require the build to setup as many defines as it will default to not defined.

In this instance, not self vendoring seems like a good idea.  If you are using this for allowing multiple self vendorings, then we could do undefined as not, and define it as "gcc" or other things for the relevant directories.

src/snmalloc/global/memcpy.h
SchrodingerZhu: done

CMakeLists.txt
SchrodingerZhu: Good idea. Changing now.

mjp41 (approved):
LGTM.  Thanks for doing this.

PR #725, opened 2025-01-03
Remove dependencies from Singleton and FlagLock

mjp41:
The singleton and flag lock cannot be used inside the Pal.  As it was defined in `ds` rather than `ds_core`.  This alters the structure to add a ds_aal that can be used from the Pal, but can also take a dependence on the aal.


mjp41:
FYI, I want this to be able to use the Singleton inside the Pal.

mjp41:
@davidchisnall tagging you because you originally design the layering of directories.  I think this respects the original design but enables a couple of pieces to be used in the Pal.  This is important for implementing #727 without having to duplicate code.

SchrodingerZhu (approved):
the change looks good to me

PR #727, opened 2025-01-06
Add Windows Support for Lazy Commit of page map

mjp41:
The Pagemap can potentially be accessed where there is no direct
knowledge of the location.  On platforms with lazy commit this can
be handled by the OS faulting in a zero page.  Here we modify the
Windows PAL to add an exception handler that will commit pages
in the read only range, i.e. the page map.  We still require explicit
notify_using to be able to modify the pagemap.

Stacked on top of #725


SchrodingerZhu (approved):
LGTM.

PR #729, opened 2025-01-08
Changing DEBUG constexpr to SNMALLOC_DEBUG

NeilMonday:
Defining the constexpr DEBUG caused problems when building SNMALLOC into our project. This change simply changes the constexpr's name to SNMALLOC_DEBUG.

### DIFF
--- a/src/snmalloc/ds_core/defines.h
+++ b/src/snmalloc/ds_core/defines.h
@@ -118,9 +118,9 @@
 namespace snmalloc
 {
 #ifdef NDEBUG
-  static constexpr bool DEBUG = false;
+  static constexpr bool Debug = false;
 #else
-  static constexpr bool DEBUG = true;
+  static constexpr bool Debug = true;
 #endif
 
   // Forwards reference so that the platform can define how to handle errors.
--- a/src/snmalloc/ds_core/mitigations.h
+++ b/src/snmalloc/ds_core/mitigations.h
@@ -19,7 +19,7 @@ namespace snmalloc
   {
     if (SNMALLOC_UNLIKELY(!test))
     {
-      if constexpr (!DEBUG)
+      if constexpr (!Debug)
       {
         UNUSED(str, args...);
         SNMALLOC_FAST_FAIL();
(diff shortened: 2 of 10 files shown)


NeilMonday:
> @microsoft-github-policy-service agree

NeilMonday:
> @NeilMonday, Could you elaborate a bit more on this. The `constexpr` is already in a `namespace`:
> 
> ```c++
> snmalloc::DEBUG
> ```
> 
> Are you including the whole namespace, and then getting conflicts?

In my project, I have a preprocessor definition "DEBUG=1" for debug builds. So, I think this is conflicting with the snmalloc::DEBUG even though it is in its own namespace?

SchrodingerZhu:
is it possible to use DDEBUG or other prefixed DEBUG macro in your project? Or use #pragma macro_push to guard around the inclusion of snmalloc header?

SchrodingerZhu:
See https://gcc.gnu.org/onlinedocs/gcc/Push_002fPop-Macro-Pragmas.html

mjp41:
So at one point I tried to use the convention that
* `constexpr` were capitalised, and 
* macros were all caps and preceeded with SNMALLOC_.
 
So if we did

```C++
snmalloc::Debug
```

Then I would be happy, and I think that would no longer conflict with the pre-processor expansion.  That would hopefully fix your issue.  It would tidy the code to the unwritten convention.

NeilMonday:
> I think `snmalloc::Debug` would be a great solution. Let me try that now.
> 
> While looking online it seems like #define DEBUG is a common thing in some generic C++ examples (even though it isn't a default from msvc or gcc). So seems likely it could happen to someone else in the future.

mjp41 (approved):
LGTM.  Thanks for the contribution @NeilMonday.

mjp41:
@NeilMonday if you're happy can you mark as "Ready for review" and we'll merge this.

mjp41:
@SchrodingerZhu thanks for the suggestions.  I think it is probably best to avoid `DEBUG`.  If one person has an issue, others will.

PR #733, opened 2025-01-19
pal_windows.h: missing <cerrno> for ENOMEM on MinGW

jcelerier:


mjp41:
@schrodingerzhu, is this an area where we are missing stuff for freestanding?

jcelerier:
> However, it is a bit inconsistent that we include cxx header for mingw targets.

hmm why ?

mjp41:
> > However, it is a bit inconsistent that we include cxx header for mingw targets.
> 
> hmm why ? it's C++ code, the valid header to include is the <c....> cxx version, not the C .h version and both GCC and Clang as shipped by mingw become more strict with this over time

Just for some context.  @Schrodingerzhu has been working on getting snmalloc as a built-in option in the LLVM libc.  To do that, they have removed a bunch of dependencies on the C++ header files.  This is mostly a refactoring.  My comment was mostly "@schrodingerzhu I think you missed a bit."  rather than a comment about your change.

src/snmalloc/pal/pal_windows.h
```diff
@@ -7,6 +7,7 @@
 #ifdef _WIN32
 #  ifndef _MSC_VER
 #    include <cstdio>
+#    include <cerrno>
```
mjp41, line 10: I think the following would help with the snmalloc inside libc changes.  Also, the clangformat rules want them alphabetically. 

```suggestion
#    include <errno.h>
#    include <stdio.h>
```

mjp41 (changes requested):
Thanks for doing this.

jcelerier:
> > > However, it is a bit inconsistent that we include cxx header for mingw targets.
> > 
> > 
> > hmm why ? it's C++ code, the valid header to include is the <c....> cxx version, not the C .h version and both GCC and Clang as shipped by mingw become more strict with this over time
> 
> Just for some context. @SchrodingerZhu has been working on getting snmalloc as a built-in option in the LLVM libc. To do that, they have removed a bunch of dependencies on the C++ header files. This is mostly a refactoring. My comment was mostly "@SchrodingerZhu I think you missed a bit." rather than a comment about your change.


ahh thanks for the context, makes more sense!

SchrodingerZhu (approved):
LGTM

mjp41 (approved):

Issue #630, opened 2023-08-30
Consistent state across forks


davidchisnall:
I am increasingly unconvinced that snmalloc is safe to use in programs that call `fork`.  I had originally thought that the mostly lockless design meant that we're safe, but I think there are three problems:

 - If a thread holds any of the range locks, we can deadlock on `fork`.
 - Any allocators owned by threads other than the one calling `fork` will leak.
 - Any allocators owned by threads other than the one calling `fork` will be in an inconsistent state.

This is all bad.  

The first problem can be solved by acquiring the lock(s?) in the prepare callback for `pthread_atfork` and then releasing it in both the parent and child callbacks.

The second can be solved by putting all allocated allocators in a linked list and walking that list in the child to return them all to the pool (and probably coalescing them at that point).

I'm not sure about the third.  Do we atomically pop things off free lists on the fast path?  On anything that's doing read-modify-write operations on an allocator, we'd need an asymmetric barrier.  

I'm not especially worried about this because any program that calls `malloc` from a multithreaded program that calls `fork` and not `execve` is going to leak memory, but ideally `malloc` between `fork` and `execve` should work (which requires fixing the first one).

mjp41:
Two other places I can think of in addition 

* the allocator pool has a lock as well. So needs fixing in the same way. 
* the singleton pattern uses a lock too.
* message queue is non-linearisable. This means you can begin an enqueue, another thread forks, and then the message queue will be in a state where it will never become reconnected to the `back`. The child post fork handler should process the message queue and reconnect the `back` if required.

mjp41:
So I finally got around to thinking about this.  I had an idea for a C++ class to encapsulate  the logic for preventing deadlocks, but not leaks:

```C++
// This is a simple implementation of a class that can be
// used to prevent a process from forking. Holding a lock
// in the allocator while forking can lead to deadlocks.
// This causes the fork to wait out any other threads inside
// the allocators locks.
//
// The use is
// ```
//  {
//     PreventFork pf;
//     // Code that should not be running during a fork.
//  }
// ```
class PreventFork
{
  // Global atomic counter of twice the number of threads currently preventing the system from forking.
  // The bottom bit is used to signal that a thread is wanting to fork.
  static std::atomic<size_t> threads_preventing_fork{0}

  // The depth of the current thread's prevention of forking.
  // This is used to enable reentrant prevention of forking.
  static thread_local size_t depth_of_prevention{0};

[...]

PR #735, opened 2025-01-24
Protection against fork

mjp41:
If a thread forks, while another thread is holding an snmalloc lock, then the allocator could stop working.

This patch attempts to protect against the cases of this. There is one case that is not covered. If a fork occurs during the very first allocation. This can result in the installation of the fork handler racing with the fork, and all bets are off.

Address #630

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -120,6 +120,19 @@ int main()
  SNMALLOC_LINKER_SUPPORT_NOSTDLIBXX)
 set(CMAKE_REQUIRED_LINK_OPTIONS "")
 
+# Detect if pthread_atfork works
+CHECK_CXX_SOURCE_COMPILES("
+#include <pthread.h>
+void prepare() {}
+void parent() {}
+void child() {}
+int main() {
+  pthread_atfork(prepare, parent, child);
+  return 0;
+}
+" SNMALLOC_PTHREAD_ATFORK_WORKS)
+
+
 if (NOT MSVC AND NOT (SNMALLOC_CLEANUP STREQUAL CXX11_DESTRUCTORS))
   # If the target compiler doesn't support -nostdlib++ then we must enable C at
   # the global scope for the fallbacks to work.
@@ -320,6 +333,7 @@ add_as_define(SNMALLOC_QEMU_WORKAROUND)
 add_as_define(SNMALLOC_TRACING)
 add_as_define(SNMALLOC_CI_BUILD)
 add_as_define(SNMALLOC_PLATFORM_HAS_GETENTROPY)
+add_as_define(SNMALLOC_PTHREAD_ATFORK_WORKS)
 add_as_define(SNMALLOC_HAS_LINUX_RANDOM_H)
 add_as_define(SNMALLOC_HAS_LINUX_FUTEX_H)
 if (SNMALLOC_NO_REALLOCARRAY)
--- a/src/snmalloc/ds/combininglock.h
+++ b/src/snmalloc/ds/combininglock.h
@@ -274,26 +274,32 @@ namespace snmalloc
   template<typename F>
   inline void with(CombiningLock& lock, F&& f)
   {
-    // Test if no one is waiting
-    if (SNMALLOC_LIKELY(lock.last.load(stl::memory_order_relaxed) == nullptr))
+    // A unix fork while holding a lock can lead to deadlock. Protect against
+    // this by not allowing a fork while holding a lock.
+    PreventFork pf;
+    snmalloc::UNUSED(pf);
     {
-      // No one was waiting so low contention. Attempt to acquire the flag
-      // lock.
-      if (SNMALLOC_LIKELY(
-            lock.flag.exchange(true, stl::memory_order_acquire) == false))
+      // Test if no one is waiting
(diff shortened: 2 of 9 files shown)


SchrodingerZhu:
Looking into it today

SchrodingerZhu:
Need to adjust header order.

The change looks good to me but I may need to double check whether snmalloc is calling any non-async-signal-safe function (which would be UB if used after fork).

davidchisnall:
Calling malloc after fork is undefined behaviour, so I’m not too worried about this as a problem (which is why I didn’t fix it three years ago). Using the atfork hooks may be problematic because the exact time at which they’re called is a bit loosely defined and we’re unordered with respect to other users of these APIs. I am not sure what happens in your code if another user calls malloc or free in their pre/post fork code.

FreeBSD libc provides two hooks: `_malloc_prefork` and `_malloc_postfork`, which are guaranteed to be called only in multithreaded code and called after other prefork before other post-fork hooks. Acquiring the lock in the prefork and releasing it in postfork should be sufficient there. 

These hooks exist because jemalloc determined that malloc being fork safe was not possible with the standard pthread APIs.

We should probably expose this as a PAL hook as we do with some of the other libc-specific things.

mjp41:
@davidchisnall, I just want to check.  The original issue that you raised said: 

> ideally malloc between fork and execve should work

But I think from your more recent message on this PR, this is UB, and not something we should address.

> I am not sure what happens in your code if another user calls malloc or free in their pre/post fork code.

This is a great observation, I hadn't consider multiple handlers.  This code could deadlock if an allocation occurs in a later prefork, or earlier postfork than the ones supplied by snmalloc.  So this code makes things worse, than doing nothing.

I am going to drop this PR, and close the underlying issue as I believe we are as safe as the standard requires.

SchrodingerZhu:
FYI, according to [POSIX.1-2024](https://pubs.opengroup.org/onlinepubs/9799919799/), 

> A process shall be created with a single thread. If a multi-threaded process calls fork(), the new process shall contain a replica of the calling thread and its entire address space, possibly including the states of mutexes and other resources. Consequently, the application shall ensure that the child process only executes async-signal-safe operations until such time as one of the [exec](https://pubs.opengroup.org/onlinepubs/9799919799/functions/exec.html) functions is successful.

So, only [a very limited list of posix functions](https://man7.org/linux/man-pages/man7/signal-safety.7.html) could be invoked post fork. ( I think direct syscall wrappers that are not in POSIX are also safe to call).

davidchisnall:
Note, in particular, that `malloc` and `free` are *not* on the list of async signal safe functions, so it’s fine according to POSIX for a multithreaded program’s child to deadlock if it calls snmalloc between fork and exec. 

That said, FreeBSD libc *does* provide hooks that allow the bundled jemalloc to make this work. Snmalloc would have to acquire all of the locks for chunk allocators in the pre hook and release them in the post hook. I don’t think it’s worth trying to make this work on platforms that don’t expose equivalent hooks (I think most libcs do, perhaps with different spellings).

mjp41:
I have fixed the issue with other prefork and postfork calls.

I am still concerned about the first call to allocation being in parallel with another thread call ing `fork`.  I think there is an interesting race condition here. I couldn't find out what the semantics of `fork` in parallel with `pthread_atfork`.  One hopes that the effects are either `pthread_atfork` occurs entirely before `fork`, or the `fork` completes before the `pthread_atfork`.  That is a call to `fork` blocks `pthread_atfork`, and a call to `pthread_atfork` blocks the starting of a call to `fork`.

If this has a sensible semantics, then I think this code could work as is.  Looking at glibc there is a lock guarding both addition of a handler and running of all the handlers, so this would give the sensible semantics.

mjp41:
Sadly, looking at the more recent code that locks was made more complex, and they broke this property.

https://sourceware.org/git/?p=glibc.git;a=commit;h=52a103e237329b9f88a28513fe7506ffc3bd8ced

This drops the lock during each call to a handler.  This was required to handle re-entrancy.  But this means a parallel call can no longer work correctly.

mjp41:
I feel I have descended down a rabbit hole on this one. So the following code on my 22.04 installation terminates, and on a 24.04 deadlocks:

https://github.com/mjp41/pthread_atfork/blob/main/main_reentrant.cpp 

@schrodingerzhu @davidchisnall  does this make sense to you?   Releasing the lock during fork pre/post process alters the semantics.

mjp41:
Submitted a bug report to glibc
https://sourceware.org/bugzilla/show_bug.cgi?id=32615

SchrodingerZhu:
> Submitted a bug report to glibc https://sourceware.org/bugzilla/show_bug.cgi?id=32615

you can also send the bug to [libc-coord](https://www.openwall.com/lists/libc-coord/2020/01/30/1). That mailing list should be quite active.

mjp41:
> > Submitted a bug report to glibc https://sourceware.org/bugzilla/show_bug.cgi?id=32615
> 
> you can also send the bug to [libc-coord](https://www.openwall.com/lists/libc-coord/2020/01/30/1) if you consider it important to discuss about. That mailing list should be quite active.

Sent a message CCing you.

PR #737, opened 2025-01-27
Enable a seconary allocator support (e.g. GWP-Asan)

SchrodingerZhu:


src/snmalloc/mem/localalloc.h
```diff
@@ -767,6 +768,9 @@ namespace snmalloc
 #ifdef SNMALLOC_PASS_THROUGH
       return external_alloc::malloc_usable_size(const_cast<void*>(p_raw));
 #else
+
+      if (SecondaryAllocator::has_secondary_ownership(p_raw))
```
mjp41, line 772: I would prefer this to check that the allocation is not owned by snmalloc, rather than rely on the secondary allocator having this.  For GWP Asan, it is fine.

But I would also been keen to use this to build an approach on Windows to replace the allocator, and pass any original frees back to the Windows Heap.

src/snmalloc/mem/secondary/gwp_asan.h
```diff
@@ -0,0 +1,72 @@
...
+      };
+#endif
+      singleton.init(opt);
+      max_allocation_size =
+        singleton.getAllocatorState()->maximumAllocationSize();
+    }
+
+    SNMALLOC_FAST_PATH static void* allocate(size_t size)
+    {
+      if (SNMALLOC_UNLIKELY(singleton.shouldSample()))
```
mjp41, line 36: As this is only getting called when a free list is exhausted should we be doing something else here.  I.e. should we be passing how many requests of this size have been handled since we last called this allocator?

SchrodingerZhu: `shouldSample` is implemented as
```
    // NextSampleCounter == 0 means we "should regenerate the counter".
    //                   == 1 means we "should sample this allocation".
    // AdjustedSampleRatePlusOne is designed to intentionally underflow. This
    // class must be valid when zero-initialised, and we wish to sample as
    // infrequently as possible when this is the case, hence we underflow to
    // UINT32_MAX.
    if (GWP_ASAN_UNLIKELY(getThreadLocals()->NextSampleCounter == 0))
      getThreadLocals()->NextSampleCounter =
          ((getRandomUnsigned32() % (AdjustedSampleRatePlusOne - 1)) + 1) &
          ThreadLocalPackedVariables::NextSampleCounterMask;

    return GWP_ASAN_UNLIKELY(--getThreadLocals()->NextSampleCounter == 0);
```

src/snmalloc/mem/secondary/gwp_asan.h
```diff
@@ -0,0 +1,72 @@
...
+      snmalloc_check_client(
+        mitigations(sanity_checks),
+        singleton.pointerIsMine(pointer),
+        "Not allocated by snmalloc or secondary allocator");
+
+      singleton.deallocate(pointer);
+    }
+
+    SNMALLOC_FAST_PATH
+    static bool has_secondary_ownership([[maybe_unused]] const void* pointer)
```
devnexen, line 61: nit: do not think [maybe_unused] is necessary here ?

src/snmalloc/mem/secondary/default.h
```diff
@@ -0,0 +1,49 @@
...
+namespace snmalloc
+{
+  class DefaultSecondaryAllocator
+  {
+  public:
+    SNMALLOC_FAST_PATH
+    static void initialize() {}
+
+    SNMALLOC_FAST_PATH
+    static void* allocate([[maybe_unused]] size_t size)
```
mjp41, line 17:
```suggestion
    static void* allocate(size_t)
```

src/snmalloc/mem/secondary/default.h
```diff
@@ -0,0 +1,49 @@
...
+      // it shouldn't be called on.
+      // TODO: Should this be tested even in the !CHECK_CLIENT case?
+      snmalloc_check_client(
+        mitigations(sanity_checks),
+        pointer == nullptr,
+        "Not allocated by snmalloc.");
+    }
+
+    SNMALLOC_FAST_PATH
+    static bool has_secondary_ownership([[maybe_unused]] const void* pointer)
```
mjp41, line 35:
```suggestion
    static bool has_secondary_ownership(const void*)
```

src/snmalloc/mem/secondary/default.h
```diff
@@ -0,0 +1,49 @@
...
+    {
+      return false;
+    }
+
+    SNMALLOC_FAST_PATH
+    static size_t alloc_size([[maybe_unused]] const void* pointer)
```
mjp41, line 41:
```suggestion
    static size_t alloc_size(const void*)
```

src/snmalloc/mem/secondary/gwp_asan.h
mjp41: Sorry, I meant.  Normally usage of GWP Asan, every call to `malloc` will call `shouldSample`.  But this usage has moved it from the fast path, so we will call it every new free list for that size.  This in the case of 16 byte allocations worst case could be 1024 allocations between calls.  This will alter the likelihood of it being called, and might bias away from small allocations.  I think for GWP Asan on we would want randomisation too, so that the free list that gets exhausted.

I think this is probably the same information that would be required to unbias profiling, so should probably be left to a subsequent PR, but we should perhaps add either an issue or a comment about this.

mjp41 (changes requested):

src/snmalloc/mem/localalloc.h
SchrodingerZhu:
```
    SNMALLOC_FAST_PATH size_t alloc_size(const void* p_raw)
    {
#ifdef SNMALLOC_PASS_THROUGH
      return external_alloc::malloc_usable_size(const_cast<void*>(p_raw));
#else

      if (!check_domestication(p_raw))
        return SecondaryAllocator::alloc_size(p_raw);

  }
```

Could you illustrate a better way to do this? If do it the way above, we will always have the overhead for domestication test.

mjp41:
```C++
      const PagemapEntry& entry =
        Config::Backend::get_metaentry(address_cast(p_raw));

      // Check is managed by snmalloc
      if (SNMALLOC_LIKELY(entry.get_remote() != nullptr)
        return sizeclass_full_to_size(entry.get_sizeclass());

      // If not managed by snmalloc, pass to secondary allocator.
      return SecondaryAllocator::alloc_size(p_raw);
```

src/snmalloc/mem/localalloc.h
```diff
@@ -186,6 +187,10 @@ namespace snmalloc
         return small_alloc<NoZero>(1);
       }
 
+      void* result = SecondaryAllocator::allocate(size);
+      if (result != nullptr)
+        return capptr::Alloc<void>::unsafe_from(result);
+
```
mjp41, line 193: I think we should probably move this inside the `check_init` call.  I.e. just before `// Grab slab of correct size`, I think we might have initialisation issues if we haven't fully initialised snmalloc as the pagemap won't exist when we look up a GWP Asan allocation.

src/snmalloc/mem/corealloc.h
```diff
@@ -885,6 +886,10 @@ namespace snmalloc
     {
       size_t rsize = sizeclass_to_size(sizeclass);
 
+      void* result = SecondaryAllocator::allocate(rsize);
+      if (result != nullptr)
+        return capptr::Alloc<void>::unsafe_from(result);
+
```
mjp41, line 892: Thinking about this a bit more.  I wonder if this should be at the start of `small_alloc`.  I think there are allocation patterns that would never hit this path. It is possible to work completely without having to grab a new chunk from the backend.

Sorry, I think I had this wrong in my original issue.

SchrodingerZhu:
@mjp41 could you help rerun the failed test in the latest action submission

mjp41 (commented):
Thanks for all the changes.  Just one small bit left.

@nwf would you be able to take a look at the comment below?

src/snmalloc/mem/localalloc.h
```diff
@@ -637,10 +639,29 @@ namespace snmalloc
...
+      auto p_tame =
+        capptr_domesticate<Config>(core_alloc->backend_state_ptr(), p_wild);
+      const PagemapEntry& entry =
+        Config::Backend::get_metaentry(address_cast(p_tame));
+      return {p_tame, entry};
+    }
 
-      capptr::AllocWild<void> p_wild = capptr_from_client(p_raw);
+    // Check if a pointer is domestic to SnMalloc
+    SNMALLOC_FAST_PATH bool check_domestication(const void* p_raw)
```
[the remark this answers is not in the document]
mjp41: @schrodingerzhu, if you could just perform the rename, and then I will merge this PR. Thanks so much.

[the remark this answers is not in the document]
SchrodingerZhu: Doing it now.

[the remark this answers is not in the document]
SchrodingerZhu: done

mjp41 (approved):
Awesome work. Thanks

PR #746, opened 2025-02-11
Adding a function to deinitialise the singleton so we can remove our exception handler.

NeilMonday:
When using snmalloc in a DLL, the exception handler is not removed when the DLL is unloaded. This will cause a crash if the EXE later throws an exception.

Example:
1. An EXE loads a DLL which uses snmalloc
2. AddVectoredExceptionHandler gets called to add HandleReadonlyLazyCommit as an exception handler
3. The EXE unloads the DLL (the memory at the address where the function HandleReadonlyLazyCommit existed has been freed)
4. The EXE throws an exception, and is expecting its own exception handler to handle it
5. While throwing the EXE's exception, MSVC tries to call HandleReadonlyLazyCommit and hits an access violation since that no longer exists

We resolve this, I'm adding code to call RemoveVectoredExceptionHandler when the Singleton is destroyed. That way MSVC won't have a dangling function pointer after the DLL is unloaded.

Note: This adds an extra template param to Singleton class, which probably will affect other platforms, which I have not updated.

### DIFF
--- a/src/snmalloc/pal/pal_windows.h
+++ b/src/snmalloc/pal/pal_windows.h
@@ -101,15 +101,31 @@ namespace snmalloc
       return EXCEPTION_CONTINUE_EXECUTION;
     }
 
+    // Keep a handle for the exception handler, so we can remove it later
+    // when needed.
+    inline static PVOID g_Handler{};
+
     static void initialise_for_singleton(size_t*) noexcept
     {
-      AddVectoredExceptionHandler(1, HandleReadonlyLazyCommit);
+      if (!g_Handler)
+      {
+        g_Handler = AddVectoredExceptionHandler(1, HandleReadonlyLazyCommit);
+      }
+    }
+
+    static void deinitialise_for_singleton(size_t*) noexcept
+    {
+      if (g_Handler)
+      {
+        RemoveVectoredExceptionHandler(g_Handler);
+        g_Handler = NULL; // Prevent dangling pointer
+      }
     }
 
     // Ensure the exception handler is registered.
     static void initialise_readonly_av() noexcept
     {
-      static Singleton<size_t, &initialise_for_singleton> init;
+      static Singleton<size_t, &initialise_for_singleton, &deinitialise_for_singleton> init;
       init.get();
     }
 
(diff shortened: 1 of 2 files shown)


src/snmalloc/pal/pal_windows.h
```diff
@@ -101,15 +101,31 @@ namespace snmalloc
...
+    inline static PVOID g_Handler{};
+
     static void initialise_for_singleton(size_t*) noexcept
     {
-      AddVectoredExceptionHandler(1, HandleReadonlyLazyCommit);
+      if (!g_Handler)
+      {
+        g_Handler = AddVectoredExceptionHandler(1, HandleReadonlyLazyCommit);
+      }
+    }
```
mjp41, line 114: How about something like this:

```suggestion
      // Keep a handle for the exception handler, so we can remove it later
      // when needed.
      static PVOID g_Handler{};

      // Destructor for removing exception handler.
      static OnDestruct tidy([](){
        if (g_Handler)
        {
          RemoveVectoredExceptionHandler(g_Handler);
          g_Handler = NULL; // Prevent dangling pointer
        }
      });
      // Add exception handler for lazy commit.
      if (!g_Handler)
      {
        g_Handler = AddVectoredExceptionHandler(1, HandleReadonlyLazyCommit);
      }
    }
```

It wouldn't need the Singleton to change, and I believe should get called in the right place to remove the exception handler?

mjp41: This is using the 
 
https://github.com/microsoft/snmalloc/blob/a3fe420af250f4662aced78b167e45166d202898/src/snmalloc/ds_core/helpers.h#L71-L86

which is how I typically get things to get destructed.

NeilMonday: That seems like it was made exactly for this situation 👍 

I'll just try it out today and update the PR.

mjp41 (approved):
LGTM, thanks for addressing the feedback.  This is not currently tested by CI.  Can you confirm that it addresses your issue, then I can merge.

NeilMonday:
> LGTM, thanks for addressing the feedback. This is not currently tested by CI. Can you confirm that it addresses your issue, then I can merge.

Everything looks good on my end. It is now able to remove the exception handler. Thanks.

Issue #747, opened 2025-02-12
stl vendor and new override


mjp41:
Currently 

https://github.com/microsoft/snmalloc/blob/a3fe420af250f4662aced78b167e45166d202898/src/snmalloc/override/new.cc#L36-L39

uses std::nothrow. I think this should really include `#include <new>` to get this.  I don't think this affects vendoring for a libc as this is very much at the level of libstdc++/libc++, so we should always have the actual header available?

@SchrodingerZhu do you have thoughts on this?

PR #749, opened 2025-02-13
build: Fix nothrow include.

mjp41:
With STL no longer being included by the primary headers the `new.cc` override needs to include `<new>` to get `nothrow`


mjp41:
Addesses #747

PR #750, opened 2025-02-14
Refactor: Remove unused features and functions, and move most allocator operations to a global namespace.

mjp41:
Three main changes:
* Removing pass through, as VeronaRT can do this more cleanly now
* Adding a Config into the snmalloc namespace rather than it being implicit on Alloc::Config.  This also makes Alloc a derived concept.
* Reducing the number of functions that are on the allocator, and moving them to the global snmalloc namespace. E.g. the instance method: `snmalloc::Alloc::alloc`, becomes `snmalloc::alloc`.

A lot of this refactoring is to make changing the structure stored in the TLS easier, by moving a lot of the logic out of `localalloc.h`, which will go away in a subsequent refactor for the TLS.


mjp41:
Best to review using individual commits as each one is a fairly self contained change, but the combination is quite large. @nwf I have learnt from you ;-)

PR #751, opened 2025-02-20
Refactor representation of thread local state.

mjp41:
Change thread local state to be a single pointer to a backend allocated structure.  Previously, the TLS contained a complex structure, but that didn't interact well with loading numerous copies at once.


src/snmalloc/global/globalalloc.h
```diff
@@ -4,7 +4,7 @@
 
 namespace snmalloc
 {
-  template<SNMALLOC_CONCEPT(IsConfig) Config>
+  template<SNMALLOC_CONCEPT(IsConfig) Config = Config>
```
nwf, line 8: 1. I'm a little surprised these aren't triggering shadowing warnings?
2. Are these functions ever invoked with non-default `Config`s?  If not, maybe convert the template parameter and concept annotation to a suitable `static_assert` `requires`?

CMakeLists.txt
```diff
@@ -574,22 +574,9 @@ endif()
...
-      set(FLAVOURS fast;check)
-    else()
-      set(FLAVOURS fast;check;malloc)
-    endif()
+    set(FLAVOURS fast;check)
 
     foreach(FLAVOUR ${FLAVOURS})
-      if (${FLAVOUR} STREQUAL "malloc")
-        set(DEFINES SNMALLOC_PASS_THROUGH)
-      endif()
```
nwf, line 592: :tada:

src/snmalloc/global/libc.h
```diff
@@ -71,11 +70,11 @@ namespace snmalloc::libc
         SNMALLOC_ASSUME(ptr != nullptr);
         ::memcpy(p, ptr, sz);
       }
-      a.dealloc(ptr);
+      dealloc(ptr);
     }
     else if (SNMALLOC_LIKELY(size == 0))
     {
-      a.dealloc(ptr);
+      dealloc(ptr);
```
nwf, line 77: Is the inliner sufficiently clever &| informed to keep this to a single TLS lookup?

nwf:
Only found time to make it into "Restrict ThreadAlloc usage to globalalloc", but so far so great.  Thanks for nwfifying the commit stack. :D

src/snmalloc/global/libc.h
mjp41: Good point, I should check that.  I normally only look at malloc and free, and not the disasm for realloc, which generally is not that nice.

src/snmalloc/mem/check_init.h
```diff
@@ -0,0 +1,33 @@
...
+
+namespace snmalloc
+{
+  /**
+   * This class provides a default implementation for checking for
+   * initialisation. It does nothing. This is used to allow the thread local
+   * state to inject code for correctly initialising the thread local state with
+   * an allocator, but this check is performed off the fast path.
+   */
+  class CheckInitDefault
```
nwf, line 11: Maybe `UnCheckedInit` or `CheckInitNoOp`?

src/snmalloc/mem/localalloc.h
```diff
@@ -212,39 +134,48 @@ namespace snmalloc
...
-      });
+      };
+
+      return CheckInit::check_init(
+        fast_path,
+        [](size_t size, LocalAllocator* self) {
+          return self->alloc_not_small<zero_mem, CheckInitDefault>(size);
+        },
+        size,
+        this);
```
nwf: It's a bit of a pity that something so formulaic can't, AFAIK, be templated away.

src/snmalloc/mem/freelist_queue.h
```diff
@@ -145,11 +135,15 @@ namespace snmalloc
       Cb cb)
     {
       invariant();
-      SNMALLOC_ASSERT(front.load() != nullptr);
+
+      freelist::HeadPtr curr = domesticate_head(front.load());
+      if (curr == nullptr)
+        // First entry is still in progress of being added.
+        // Nothing to do.
+        return;
```
nwf, line 143: `{}` please. :)

Also more seriously, argh, what a frustrating branch!

nwf (commented):
This continues to look really nice.  (Read up through "WIP", exclusive.)

src/snmalloc/mem/freelist_queue.h
mjp41: Yeah, but I needed to not have encoded pointers needed in the representation for empty.  This was the simplist solution I could come up with.

I am a bit unsure that we should encode nullptr, and instead have a representation that doesn't encode null.  I would like to do this, but as a subsequent PR.

src/snmalloc/global/libc.h
mjp41: It accesses the TLS a few times, but there is register pressure as the slow paths cause calls to occur, so it is not clear the code generated is any worse.  The perf in mimalloc-bench is about the same.

src/snmalloc/global/globalalloc.h
mjp41: I did `Config_` in #750.

src/snmalloc/global/libc.h
mjp41: I'm going to look at optimising realloc after this PR.  So going to leave this for now.

src/snmalloc/mem/corealloc.h
```diff
@@ -562,195 +422,561 @@ namespace snmalloc
...
+     *             - Check if the allocator has been initialised, if it hasn't
+     *               restart with a fresh allocator.
+     *             - Otherwise, call small_refill_slow(size_t) which allocates
+     *               a new slab and fills it with a free list using
+     *               alloc_new_list.
+     *    - static alloc_not_small(size_t, Allocator*)
+     *      - If size is zero, call small_alloc(1) - this deals with the
+     *        annoying case of zero off the fast path.
+     *      - Otherwise, call backend to allocate a large allocation.
+     *************************************************************************/
```
nwf, line 554: I love it. :heart:

mjp41:
I have been throught this carefully today, and am going to merge.  There is one outstanding issue that I am not sure how we should modify the domestication test.  There is complexity with how BatchIt interacts.  I will raise an issue to fix this down the line, but this shouldn't block other work.

PR #752, opened 2025-02-25
Bug fixes

mjp41:
* Address overflow in alignment calculation not failing an allocation.
* Fix heuristic for how long to process a batch for

### DIFF
--- a/src/snmalloc/mem/sizeclasstable.h
+++ b/src/snmalloc/mem/sizeclasstable.h
@@ -129,17 +129,6 @@ namespace snmalloc
 
   using sizeclass_compress_t = uint8_t;
 
-  constexpr SNMALLOC_FAST_PATH static size_t
-  aligned_size(size_t alignment, size_t size)
-  {
-    // Client responsible for checking alignment is not zero
-    SNMALLOC_ASSERT(alignment != 0);
-    // Client responsible for checking alignment is a power of two
-    SNMALLOC_ASSERT(bits::is_pow2(alignment));
-
-    return ((alignment - 1) | (size - 1)) + 1;
-  }
-
   /**
    * This structure contains the fields required for fast paths for sizeclasses.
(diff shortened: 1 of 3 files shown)


src/snmalloc/mem/sizeclasstable.h
```diff
@@ -137,7 +137,9 @@ namespace snmalloc
     // Client responsible for checking alignment is a power of two
     SNMALLOC_ASSERT(bits::is_pow2(alignment));
 
-    return ((alignment - 1) | (size - 1)) + 1;
+    size_t result = ((alignment - 1) | (size - 1)) + 1;
+    // Check for overflow and return the maximum size.
+    return result < size ? result - 1 : result;
```
nwf, line 142: I _think_ (and I think z3 agrees, see below) that `result < size` implies `result == 0`, but it'd be good to write that down or perhaps use the other spelling?

```
(define-fun result ( (size (_ BitVec 32)) (align (_ BitVec 32)) ) (_ BitVec 32)
  (bvadd #x00000001 (bvor (bvsub align #x00000001) (bvsub size #x00000001))))

(declare-const align (_ BitVec 32))
(declare-const size (_ BitVec 32))

(define-fun conjecture () Bool
  (=> (bvult (result size align) size) (= (result size align) #x00000000)))
(assert (not conjecture))
(check-sat)
```

nwf (commented):
LGTM, other than a request for more documentation.

src/snmalloc/mem/sizeclasstable.h
mjp41: Agreed.  I used result-1 as it was a quick way to give -1 of the right size.  I have moved to SIZE_MAX now, and also I rewrote it to generate better code with the sizeclass calc.

nwf (approved):

PR #757, opened 2025-03-11
Fixed pagemap reservations for a fixed-region use-case using the default PAL

Trithek:
When using a fixed region configuration, the pagemap needs the address to be a relative offset to the base address in order to calculate the correct range for the reservation. Without this change it will try to reserve outside of the given region.

We could just always subtract the base address without the additional if checks since base is 0 when no fixed region is configured and thus would result in no change. However, the additional checks make it clearer to the reader that this is for the bounded use-case. It is also consistent with the other functions that need similar adjustments.

### DIFF
--- a/src/snmalloc/ds/pagemap.h
+++ b/src/snmalloc/ds/pagemap.h
@@ -61,10 +61,11 @@ namespace snmalloc
 
       if constexpr (has_bounds)
       {
-        if ((p - base >= size) || (p + length - base >= size))
+        if ((p - base >= size) || (p + length - base > size))
         {
-          p = p - base;
+          PAL::error("Internal error: Pagemap registering out of range.");
         }
+        p = p - base;
       }
 
       // Calculate range in pagemap that is associated to this space.
--- a/src/snmalloc/ds/pagemap.h
+++ b/src/snmalloc/ds/pagemap.h
@@ -59,6 +59,14 @@ namespace snmalloc
     {
       SNMALLOC_ASSERT(is_initialised());
 
+      if constexpr (has_bounds)
+      {
+        if ((p - base >= size) || (p + length - base >= size))
+        {
+          p = p - base;
+        }
+      }
+
       // Calculate range in pagemap that is associated to this space.
       auto first = &body[p >> SHIFT];
       auto last = &body[(p + length + bits::one_at_bit(SHIFT) - 1) >> SHIFT];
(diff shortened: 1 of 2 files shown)


Trithek:
@microsoft-github-policy-service agree company="Remedy Entertainment"

src/snmalloc/ds/pagemap.h
```diff
@@ -59,6 +59,14 @@ namespace snmalloc
...
       SNMALLOC_ASSERT(is_initialised());
 
+      if constexpr (has_bounds)
+      {
+        if ((p - base >= size) || (p + length - base >= size))
+        {
+          p = p - base;
+        }
+      }
+
```
mjp41, line 69: I think it would be good to have an error in the case where we try to register a range that is not part of the fixed range.

mjp41 (changes requested):

src/snmalloc/ds/pagemap.h
Trithek: In the case of a fixed range, it is currently guaranteed to be out of range.
If that is not supposed to happen, then yes, I would agree.
But then there's an upstream fix to be made somewhere :)

Trithek: Alright, you were absolutely correct :)

I didn't test properly, and the check was a classic off-by-one.

The end-address should be checked with > and not with >=, as it's of course perfectly legal to end on the last byte of the arena.

Trithek: One thing that is still weird is that although it now works, when using fixed range our application uses a lot more memory.

I suspect though that's unrelated to this specific thing, so I'll dig a bit and open a new PR when I have something

Trithek: Also, the crashes on some platforms are... odd
I don't have access to those platforms though, so they're a bit hard to reproduce :(

mjp41 (approved):
LGTM, thanks for your first contribution to snmalloc.

src/snmalloc/ds/pagemap.h
mjp41: > Also, the crashes on some platforms are... odd I don't have access to those platforms though, so they're a bit hard to reproduce :(

Which platforms are crashing?  I have got a clean CI run, so don't know what needs investigating. Could you let me know what they are?

Trithek: Ah, it was in the previous runs it seems. They were crashing on Mac and some older Linux, but not the newer.
I have no idea why :)

But yes, they're all green now. Great!

PR #763, opened 2025-03-28
[Rust] add support for optionally compiling libc functions

wpbrown:
- Include malloc.cc in the same compilation unit as the rust stub and mangle the functions with sn_

This is to address this issue: SchrodingerZhu/snmalloc-rs#197 

Rust PR to consume this change: SchrodingerZhu/snmalloc-rs#198


SchrodingerZhu (approved):
Looks good from my perspective.

src/snmalloc/override/rust.cc
```diff
@@ -1,5 +1,11 @@
 #define SNMALLOC_NAME_MANGLE(a) sn_##a
-#include "snmalloc/snmalloc.h"
+
+// The libc API provided by malloc.cc will always be mangled per above.
+#ifdef SNMALLOC_RUST_LIBC_API
+#  include "malloc.cc"
```
mjp41, line 5: Could we move the `SNMALLOC_NAME_MANGLE` inside the `ifdef`?
```suggestion
#  define SNMALLOC_NAME_MANGLE(a) sn_##a
#  include "malloc.cc"
```

wpbrown: Sorry the comment said "move" but I see in your suggested edit it is "copied". Shall I remove the comment then too?

wpbrown: reverted

mjp41:
@wpbrown thanks for this.

PR #764, opened 2025-04-01
Header-only build fix for WIN32_LEAN_AND_MEAN redefinition

sooknarine:
When using snmalloc as header-only library in a project that globally defines `WIN32_LEAN_AND_MEAN`, there is a compile error for macro redefinition.

Guard the pal_windows.h define.


mjp41 (approved):
LGTM, thanks.

PR #768, opened 2025-04-28
Remove 20.04 from CI

mjp41:
Remove 20.04 from CI as it is no longer available.


nwf:
I guess I should get my actual PowerPC box set up as a snmalloc CI agent.

mjp41:
That would be nice.  But not sure you want an internet facing runner on your dev box.  I'll add an issue to fix CI for PowerPC.

PR #770, opened 2025-05-07
Remove tests for experimental jemalloc ABIs

markjdb:
This is needed in order to build snmalloc's tests on recent FreeBSD main.

I verified that all of the remaining tests pass there.


mjp41:
@davidchisnall you added these bits are you okay with them going away now?

mjp41 (approved):
LGTM

PR #771, opened 2025-05-08
Adding code to release the body of the pagemap when destructing.

NeilMonday:
This alleviates the primary memory leak when unloading a DLL with snmalloc.

### DIFF
--- a/src/snmalloc/ds/pagemap.h
+++ b/src/snmalloc/ds/pagemap.h
@@ -104,7 +104,7 @@ namespace snmalloc
 
       if constexpr (pal_supports<Release, PAL>)
       {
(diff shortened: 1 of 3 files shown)


NeilMonday:
I guess I would have to implement notify_release for all PAL sub classes. Maybe as an empty function?

NeilMonday:
Converting to draft while I work on remaining issues.

NeilMonday:
@mjp41 This is what I am thinking for cleaning up the pagemap reservation. Do you have any concerns with this?

NeilMonday:
Moving back to draft. I feel like there might not be a way to restore the default operators so that the exe could continue after unloading the DLL.

PR #773, opened 2025-05-22
Release all pagemap reservations at the very end of the program or DLL

NeilMonday:
Continuation of: https://github.com/microsoft/snmalloc/pull/771. I made a new PR since it was a somewhat different concept from the original.

This seems to be a more correct way to cleanup at the end of the program or DLL.

1. Reservations are tracked in a static vector
2. This vector is a custom one that use VirtualAlloc and VirtualFree to avoid recursive calls
3. When it's time to exit, free the reservations we have created
4. #pragma init_seg(".CRT$XCB") is used to set this translation unit to be very early in the initialization phase. This means that our cleanup function passed to atexit() gets called at the very end, and our cleanup code can work properly.

### DIFF
--- a/src/snmalloc/pal/pal_windows.h
+++ b/src/snmalloc/pal/pal_windows.h
@@ -231,57 +231,57 @@ public:
 };
 
 /**
-* atexit takes only function pointers that have no arguments
-* and return void. So we create this reservations vector with
-* static storage duration and internal linkage so it is visible
-* only to this translation unit and can be deleted by the
-* ReservationsCleanup
-*/
+ * atexit takes only function pointers that have no arguments
+ * and return void. So we create this reservations vector with
+ * static storage duration and internal linkage so it is visible
+ * only to this translation unit and can be deleted by the
+ * ReservationsCleanup
+ */
 static inline VirtualVector<void*> reservations;
 
 /**
-* This is the function that gets provided to atexit, and
-* calls VirtualFree on all the reservations.
-*/
+ * This is the function that gets provided to atexit, and
+ * calls VirtualFree on all the reservations.
+ */
 static inline void ReservationsCleanup()
 {
-for (size_t i = reservations.get_size(); i > 0; i--)
-{
+  for (size_t i = reservations.get_size(); i > 0; i--)
+  {
     volatile size_t index = i - 1;
     if (reservations[index] == nullptr)
-    continue;
-
+      continue;
+    
     BOOL ok = VirtualFree(reservations[index], 0, MEM_RELEASE);
-
+    
     reservations[index] = nullptr;
-
+    
     if (!ok)
     {
-        fputs("VirtualFree failed\n", stderr);
-        fflush(stderr);
-        abort();
+      fputs("VirtualFree failed\n", stderr);
+      fflush(stderr);
+      abort();
     }
-}
+  }
 }
 
 /**
-* The ReservationsCleanupHandler's constructor registers the
-* cleanup function with atexit.
-*/
+ * The ReservationsCleanupHandler's constructor registers the
+ * cleanup function with atexit.
+ */
 struct ReservationsCleanupHandler
 {
-ReservationsCleanupHandler()
-{
+  ReservationsCleanupHandler()
+  {
     atexit(ReservationsCleanup);
-}
+  }
 };
 
 /**
-* We create a static instance of the cleanup handler,
-* which will call the constructor and register the
-* ReservationsCleanup with atexit(). This is done before
-* all other statics because of init_seg(".CRT$XCB").
-*/
+ * We create a static instance of the cleanup handler,
+ * which will call the constructor and register the
+ * ReservationsCleanup with atexit(). This is done before
+ * all other statics because of init_seg(".CRT$XCB").
+ */
 static inline ReservationsCleanupHandler cleanup_handler;
 
 namespace snmalloc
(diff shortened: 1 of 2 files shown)


NeilMonday:
@mjp41 Sorry to bug you again. What do you think of this change?

Also, I can't see how to re-run the failed job. I wanted to re-run since the equivalent jobs for 2022 and 2025 both seem to pass.

src/snmalloc/pal/pal_windows.h
```diff
@@ -27,6 +27,263 @@
...
+  ~VirtualVector()
+  {
+    if (data)
+    {
+      for (size_t i = 0; i < size; ++i)
+        data[i].~T();
+
+      VirtualFree(data, 0, MEM_RELEASE);
+    }
+  }
```
mjp41, line 88: I think it is nice that you have built a generic class, but I wonder if specialising it would make more sense.  Then the destructor of the Vector could call Virtual Free on all the things.  I am currently worried about the `VirtualVector` destructor running before the `ReservationsCleanup`.  I guess that can't happen, but I wasn't sure.

src/snmalloc/pal/pal_windows.h
```diff
@@ -27,6 +27,263 @@
...
+template<typename T>
+class VirtualVector
+{
+  T* data = nullptr;
+  size_t size = 0;
+  size_t committed_elements = 0;
+  size_t reserved_elements = 0;
+
+public:
+  VirtualVector(size_t reserve_elems = 64, size_t initial_commit = 32)
```
mjp41, line 74: Should the initial sizes be such that it is a Page?  
  (PAGE_SIZE / sizeof(T))
Perhaps, I guess there are issues if sizeof(T) is larger than a page. So would probably want a max in there to stop it becoming 0.

src/snmalloc/pal/pal_windows.h
```diff
@@ -27,6 +27,263 @@
...
+    }
+  }
+
+  void reserve_and_commit(size_t reserve_elems, size_t commit_elems)
+  {
+    size_t reserve_bytes = reserve_elems * sizeof(T);
+    T* new_block =
+      (T*)VirtualAlloc(nullptr, reserve_bytes, MEM_RESERVE, PAGE_READWRITE);
+    if (!new_block)
+      throw std::bad_alloc();
```
mjp41, line 125: Could we perhaps call error() here, I don't think anything else in the codebase uses exceptions.

src/snmalloc/pal/pal_windows.h
```diff
@@ -27,6 +27,263 @@
...
+/**
+ * The ReservationsCleanupHandler's constructor registers the
+ * cleanup function with atexit.
+ */
+struct ReservationsCleanupHandler
+{
+  ReservationsCleanupHandler()
+  {
+    atexit(ReservationsCleanup);
+  }
```
mjp41, line 276: Could other code call `atexit` and end up with the reservations going away?

mjp41 (changes requested):
I think this makes sense.  I will review again next week in more detail. Here are a few minor comments.

src/snmalloc/pal/pal_windows.h
```diff
@@ -27,6 +27,263 @@
...
+    new (&data[size]) T(value);
+    ++size;
+  }
+
+  void push_back(T&& value)
+  {
+    ensure_capacity();
+    new (&data[size]) T((T &&) value);
+    ++size;
+  }
```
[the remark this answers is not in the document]
NeilMonday: Done.

src/snmalloc/pal/pal_windows.h
```diff
@@ -27,6 +27,263 @@
...
+ * during exit in LIFO order. To make sure that we are
+ * the first call in the atexit chain, we use the pragma
+ * init_seg(".CRT$XCB")  which ensures that statics in
+ * this translation unit get initialized first and that
+ * atexit() in the constructor gets called first. Then,
+ * on program or DLL exit, we know that our function will
+ * be called last.
+ */
+#  pragma warning(disable : 4075)
+#  pragma init_seg(".CRT$XCB")
```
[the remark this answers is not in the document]
NeilMonday: This applies to the whole translation unit.

src/snmalloc/pal/pal_windows.h
NeilMonday: I refactored it so that VirtualFree happens in the destructor. It was just a gap in my understanding of atexit and the whole init_seg pragma. This next commit should much cleaner. No more function pointer sent to atexit. The static objects themselves get destroyed in FILO order and the init_seg by itself is all that is needed.

src/snmalloc/pal/pal_windows.h
NeilMonday: Done. Had to re-arrange code a bit.

mjp41:
> I’m installing VS 2019 to test that 32-bit test failure.

That failing test is a little flaky.  It has an unpredictable memory usage.  However, it does cause interesting contention that discovered a bug.  So I am not entirely happy disabling it, but I am guessing it is just out of memory.  I'll run it again.

NeilMonday:
> Small change. I think the min sizes should be a bit more pervasive. What do you think the following suggestions?

Yes, all of this makes more sense. I'll just double check locally.

NeilMonday:
@mjp41 Do you know what is causing these failures? They don't seem relevant to my change.

mjp41:
> @mjp41 Do you know what is causing these failures? They don't seem relevant to my change.

I'm investigating today.  It is blocking other PRs too.

Issue #740, opened 2025-02-05
Extend secondary allocation support


mjp41:
Here are some thoughts for things we can do

* [ ] Add support for refilling the free list to a particular size, e.g. refill(LocalCache&, sizeclass, size_t& request). `refill` will add at most `request` entries to the freelist, removes the count of added entries from `request`.
* [ ] Using modified `refill` adapt GWP-Asan sampling to accurately reflect the sampling request from the user.
* [ ] Add documentation for building with GWP-Asan.
* [ ] Automate downloading GWP-Asan. 
* [ ] Provide support for Windows using detours, and a secondary allocator of the original Windows Heap.
* [ ] Build sampling profiler using secondary allocator.
* [ ] Add C++ Concept
* [x] Enable out of tree override of secondary allocator.

This is a bit of a brain dump of things I think this feature enables.  My thought is to try to get a reasonable number of these together, and release as 0.8.0.

@SchrodingerZhu, or anyone else, please feel free to add your own ideas on use cases.

PR #774, opened 2025-06-27
Make Secondary Allocator a template parameter

mjp41:
This refactors the use of the Secondary Allocator, so that out of tree implementations can be used.


mjp41:
#740

PR #788, opened 2025-07-03
Pass continuations for success and failure cases

mjp41:
This PR provides a templated parameter to the allocation routines. This can be used to add special behaviour in both the successful allocation behaviour, and in the failing to allocate cases.

The intent of this is to enable two future features
* set_new_handler - so that the failure case doesn't just set ENOMEM, and return nullptr.  But can handle both the Windows and C++ versions of (_)set_new_handler.
* The success handler can be used to add checking, zeroing and in the future storing precise size information in metadata for each allocation.


SchrodingerZhu (approved):
LGTM

Issue #789, opened 2025-07-03
Fallible Pal::notify_using


mjp41:
On Windows, `Pal::notify_using` can fail.  This currently raises a `Pal::error`, but this is not what the standard says it should do.  Instead, it should fail the allocation that required the pages to be committed.

We should make `Pal::notify_using` return a boolean, and then for the calls to `Pal::notify_using` we should suitably propagate the error up to the user.

mjp41:
Closed by #790

PR #790, opened 2025-07-04
Fallible notify using

mjp41:
On Windows, the most common way to out-of-memory is to fail to commit a page.  The current design of the Pal assumes that notify_using always succeeds, thus we have to raise an error if it fails.  This changes the specification of notify_using to return a bool, indicating whether the notify succeeded or not.

This allows Windows to fail the notification, and then the surrounding code can handle the failure appropriately, such as by throwing an exception or returning the nullptr for the allocation.


mjp41:
Address #789

SchrodingerZhu:
LGTM

PR #792, opened 2025-07-09
Code Quality - inlining

mjp41:
This PR adds some inlining attributes to lambdas on the fast path.  Some versions of g++ and clang++ were not inlining as much as expected, and leading to poor codegen.  This PR fixes that.

Additionally, it disables the fork prevention during remote allocator posting.  This was visible in some traces as causing a bottleneck.


SchrodingerZhu (approved):

mjp41:
This was running on an Azure Linux 3 instance with clang++ (18.1.8) and g++ (13.2.0)
image

PR #799, opened 2025-10-17
Remove interface linking for PRIVATE linkage

maxtropets:
Kind of a follow up to #797 

Even `PRIVATE` linkage retains
```
# Create imported target snmalloc::snmallocshim-static
add_library(snmalloc::snmallocshim-static STATIC IMPORTED)

set_target_properties(snmalloc::snmallocshim-static PROPERTIES
  INTERFACE_LINK_LIBRARIES "\$<LINK_ONLY:snmalloc::snmalloc>"
)
```
in the resulting cmake config.

https://cmake.org/cmake/help/latest/command/target_link_libraries.html#libraries-for-both-a-target-and-its-dependents says that 
```
"Library dependencies are transitive by default with this signature. When this target is linked into another target then the libraries linked to this target will appear on the link line for the other target too. This transitive "link interface" is stored in the [INTERFACE_LINK_LIBRARIES](https://cmake.org/cmake/help/latest/prop_tgt/INTERFACE_LINK_LIBRARIES.html#prop_tgt:INTERFACE_LINK_LIBRARIES) target property and may be overridden by setting the property directly."
```

Therefore we unset the variable, although it doesn't look right, maybe.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -10,11 +10,13 @@ concurrency:
 on:
   schedule:
     - cron: "0 0 * * 1"  # Runs every Monday at midnight UTC
-  # Triggers the workflow on push or pull request events but only for the master branch
+  # Triggers the workflow on push to main or any branch starting bench.
+  # The bench/** branches can be used for test before merge any PR that
+  # might regress performance.
   push:
-    branches: [ main ]
-  pull_request:
-    branches: [ main ]
+    branches:
+      - 'main'
+      - 'bench/**'
 
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:


CMakeLists.txt
```diff
@@ -509,6 +509,10 @@ if(NOT SNMALLOC_HEADER_ONLY_LIBRARY)
   function(compile name TYPE ${ARGN})
     add_library(${name} ${TYPE} ${ARGN})
     target_link_libraries(${name} PRIVATE snmalloc)
+
+    # PRIVATE linking still touches INTERFACE level linking for unknown reasons.
```
mjp41: Could you expand this comment to include the link you mention in the PR:
https://cmake.org/cmake/help/latest/command/target_link_libraries.html#libraries-for-both-a-target-and-its-dependents

Then it is easier for someone (future me) to find out what you are referring.

PR #744, opened 2025-02-07
upstream snmalloc-rs as part of snmalloc

SchrodingerZhu:

Issue #742, opened 2025-02-05
Add Workflow to check if we break snmalloc-rs


mjp41:
Currently the downstream consumer snmalloc-rs is very dependent on the internals of snmalloc.  This strong coupling makes release a little awkward. We either need

* Release process for snmalloc, should have a PR ready to go on snmalloc-rs before we hit complete.
* Add some workflow jobs that replicate snmalloc-rs CI, and check each PR doesn't break it. (Allow failure, but need to track)

SchrodingerZhu:
If needed, as the owner of `snmalloc-rs`, I am happy to donate the whole repo to upstream and I am willing to do the most work to merge it into snmalloc.

The caveat is that we will lose the native FreeBSD CI for rust.

mjp41:
Thanks for doing this. Are you sure about this. You've put a lot of work into snmalloc-rs, and putting it under the Microsoft org, would give you less personal ownership.

I think we should think carefully about the pros and cons of merging it in. I'll try to get the CI functioning, but I propose we leave it a week or two to think about it.

SchrodingerZhu:
> but I propose we leave it a week or two to think about it.

Good idea. I would also like to collect others' idea

> You've put a lot of work into snmalloc-rs, and putting it under the Microsoft org, would give you less personal ownership.

Personally, I think it is okay as long as doing this reduces "unsync" between `snmalloc` and `snmalloc-rs`. There is a large amount of downloads of `snmalloc-rs` so I think improve its maintainability is of great importance now.

Anyway, let's take time to think about the best way to handle this. Also, another thing I want to do is to sync the version and release cycle between upstream and downstream.

mjp41:
Closed by #801

PR #801, opened 2026-01-27
snmalloc-rs

mjp41:
This pulls in the entirety of snmalloc-rs and its history, and places it under snmalloc.  This will allow for less breakages of the snmalloc-rs packages, and ensure we update them with all future releases.

See for discussion
https://github.com/microsoft/snmalloc/issues/742

This PR is based on 
https://github.com/microsoft/snmalloc/pull/744
but keeps the history intact.


SchrodingerZhu:
Thanks for doing this. I will take a look of the porting process!

SchrodingerZhu:
On windows, zero-sized array potentially requires specialization.

SchrodingerZhu:
use `TerminateProcess(GetCurrentProcess(), STATUS_ABORTED);` or `RaiseFailFastException`?

mjp41:
@schrodingerzhu, do you have any ideas why it keeps building Release static lib when it is a debug build.  I don't know really know how to debug this part of the build system.

SchrodingerZhu:
cmake-rs have some auto detection logic to inject cmake build profile

mjp41:
@schrodingerzhu I think this fixes the issues, the refactor of Array, I'll put in another PR as it is not essential to land that for the merge.

I just want to double check you are happy with this approach.  I believe we will have all the original history, and even be able to apply patches from snmalloc-rs to here.

SchrodingerZhu:
I am happy with the upstreaming. This also reduces the budern of updating the rust wrapper separately.

PR #806, opened 2026-01-29
Refactor CI

mjp41:
This restructures the CI to make it clearer what is different between each build by using reusable workflows.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -252,10 +252,13 @@ jobs:
     steps:
     - uses: actions/checkout@v4
     - name: Install cross-compile toolchain and QEMU
-      run: |
-        sudo apt update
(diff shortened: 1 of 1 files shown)


SchrodingerZhu:
If it is easier and helpful, you may want to directly absorb #808 into this.

mjp41:
> If it is easier and helpful, you may want to directly absorb #808 into this.

I'll leave it as a separate PR.  Then this one is just a refactor, and then we have a separate PR for the addition.

SchrodingerZhu:
> > If it is easier and helpful, you may want to directly absorb #808 into this.
> 
> I'll leave it as a separate PR. Then this one is just a refactor, and then we have a separate PR for the addition.

sounds good

PR #812, opened 2026-02-22
Reapplying the Decay Range and other opts

mjp41:
This reapplies the DecayRange PR to the latest version of snmalloc.  It also 
* removes some unrequired initialisation that occurs in a fast path for large allocations. Finally, it adds a LargeObjectCache to improve performance of large allocations.

This is a very draft PR with a lot of reviewing of AI refactorings required.

This might increase footprint considerably, and care is needed to configure who it operates precisely.


mjp41:
I think this is ready to review.  But we should address #814 before we commit this, as it may either hide or worsen this behaviour.

PR #818, opened 2026-02-26
Fix performance issue in Red Black Tree init

mjp41:
The paths were being unnecessarily initialised this caused a performance issue in some paths.


SchrodingerZhu (approved):
LGTM.

PR #821, opened 2026-02-27
weakavl manual

SchrodingerZhu:
- **abstract out tree policy**
- **add weak avl policy**
- **remove extra copy**
- **adjust the way of policy being specified**


SchrodingerZhu:
interesting. just got the error report. so the slow down may be due to bugs in my code.

SchrodingerZhu:
Closing this but I will keep the branch.

PR #822, opened 2026-03-02
Add SNMALLOC_EXPORT annotation to c++ allocator functions.

akrieger:
https://github.com/CleverRaven/Cataclysm-DDA/issues/85620 showed that when compiling snmalloc statically into a binary the new/delete overrides are stripped by `strip` by default because they appear unused. Add the existing SNMALLOC_EXPORT annotation to these functions so snmalloc can be used in combination with `-fvisibility=hidden` when overriding the c++ allocators, so that shared libraries use the overrides instead of loading the system default implementations.

### DIFF
--- a/src/snmalloc/override/new.cc
+++ b/src/snmalloc/override/new.cc
@@ -1,3 +1,4 @@
+#include "override.h"
 #include "snmalloc/snmalloc.h"
 
 #include <new>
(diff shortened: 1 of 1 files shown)


mjp41:
@akrieger could you clang-format.

akrieger:
formatted

mjp41 (approved):
LGTM, I think the CI failures look like packaging issues with Mac.  Nothing related to this PR.

PR #823, opened 2026-03-03
Add CHERI-safe memmove implementation

devnexen:
New attempt at custom memmove, succeeding PR #593 which was reverted due to fuzzer-found bugs (off-by-one in reverse copy loop with size_t underflow) and CHERI incompatibility (byte-at-a-time reverse copy destroys capability tags).

Key changes from the original PR #593:

- Per-Arch move() and forward_move() methods following the existing copy() pattern, instead of a single generic byte-by-byte reverse.

- Three-way overlap detection: non-overlapping uses optimized Arch::copy(), dst > src overlap uses Arch::move() (reverse), and dst < src overlap uses Arch::forward_move() (forward without the copy_end trick that re-reads already-overwritten bytes).

- block_reverse_copy<Size> operates at register width (16 bytes on x86-64/PPC64/CHERI) instead of byte-by-byte, with byte-by-byte only for the sub-register remainder.

- GenericStrictProvenance (CHERI) move() and forward_move() preserve capability tags by using pointer-pair (Ptr2) operations on aligned regions, with byte-by-byte only for sub-pointer head/tail padding where no aligned capabilities can exist.

- Comprehensive tests: overlapping copies at various sizes (1-2048) and overlap amounts, exhaustive offset x length testing for small buffers (2-64), alignment boundary tests, bounds checking, and direct snmalloc::memmove<false> unchecked path tests.

- Re-enables memmove fuzz tests (simple_memmove, forward_memmove, backward_memmove) in snmalloc-fuzzer.cpp.


devnexen:
So to summarize, Claude filled the gap in regard of CHERI here.

mjp41:
@nwf would you be able to review this?

src/snmalloc/global/memcpy.h
```diff
@@ -29,6 +29,29 @@ namespace snmalloc
...
+#if __has_builtin(__builtin_memmove)
+    __builtin_memmove(dst, src, Size);
+#else
+    // Fallback: byte-by-byte copy through a temporary buffer to avoid
+    // the compiler generating a memcpy call for struct assignment.
+    char tmp[Size];
+    for (size_t i = 0; i < Size; ++i)
+      tmp[i] = static_cast<const char*>(src)[i];
+    for (size_t i = 0; i < Size; ++i)
+      static_cast<char*>(dst)[i] = tmp[i];
```
mjp41: This doesn't feel safe on CHERI?  I guess that CHERI platforms will always have __builtin_memove?

src/snmalloc/global/memcpy.h
```diff
@@ -29,6 +29,29 @@ namespace snmalloc
...
+   * Uses __builtin_memmove which the compiler can optimize to register-width
+   * loads/stores while correctly handling overlap.  We cannot use
+   * __builtin_memcpy_inline (ASan treats it as memcpy and flags overlap)
+   * or struct copy (compiler lowers *d = *s to a memcpy call, same problem).
+   */
+  template<size_t Size>
+  SNMALLOC_FAST_PATH_INLINE void copy_one_move(void* dst, const void* src)
+  {
+#if __has_builtin(__builtin_memmove)
+    __builtin_memmove(dst, src, Size);
```
mjp41: Is it possible that __builtin_memmove can map back onto memmove?  Would lead to infinite recursion, though that is not the worst failure as it would be easy to see what is happening wrong.

src/snmalloc/global/memcpy.h
```diff
@@ -374,6 +678,38 @@ namespace snmalloc
...
+      }
+      return orig_dst;
+    }
+
+    /**
+     * Reverse copy for overlapping memmove where dst > src.
+     * No rep movsb in reverse (ERMS doesn't support DF=1), so use
+     * SSE-width block_reverse_copy.
+     */
+    static void* move(void* dst, const void* src, size_t len)
```
mjp41: For symmetry should this be called reverse_move?  Same comment for all `::move`s as they all appear to be reverse.

mjp41 (commented):
It would be great to get some eyes from CHERI experts on this (@nwf, @davidchisnall)

PR #824, opened 2026-03-04
WIP Experiment with different kind of allocator for backend.

mjp41:
The buddy allocator for large objects and chunks leads to good fragmentation behaviour, but can result in a lot of internal fragmentation as the blocks it returns can be twice the size requested.  This PR makes the blocks have more sizeclasses using a DougLea/TLSF like consolidation strategy, but with some modification to handle the alignment requires for snmalloc.

### DIFF
--- a/src/snmalloc/backend/standard_range.h
+++ b/src/snmalloc/backend/standard_range.h
@@ -40,7 +40,7 @@ namespace snmalloc
     // Decay range caches deallocated memory and gradually releases it
     // back to the parent, avoiding expensive repeated decommit/recommit
     // cycles for transient allocation patterns.
-#ifdef SNMALLOC_NO_DECAY
+#ifdef SNMALLOC_ENABLE_DECAY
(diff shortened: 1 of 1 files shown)


akrieger:
Tag me when you think it would be helpful for me to can kick the tires on this in an ad hoc fashion (or maybe I just hop on it anyway ;) ).

mjp41:
> Tag me when you think it would be helpful for me to can kick the tires on this in an ad hoc fashion (or maybe I just hop on it anyway ;) ).

It's a little premature yet.  I just wanted to get a read on CI to see how bad it was, also pushed it to get my benchmarking infrastructure to take a look.

The code is mostly me experimenting with CoPilot to see how well we can develop for snmalloc with it.  There are good and bad bits.  There are a few underlying concepts that will survive, but I will probably end up reimplementing a lot to be confident in the code.  It has a habit of taking shortcuts way more than it should on a project like this.

PR #825, opened 2026-03-10
Adding vcpkg integration

matajoh:
This PR proposes a vcpkg integration for snmalloc in both header-only library mode and static-shim configurations. Once accepted, it will be possible to add snmalloc to vcpkg repositories.

### DIFF
--- a/src/snmalloc/pal/pal_windows.h
+++ b/src/snmalloc/pal/pal_windows.h
@@ -24,6 +24,8 @@
       (WINVER >= _WIN32_WINNT_WIN10) && !defined(USE_SYSTEMATIC_TESTING)
 #      define PLATFORM_HAS_VIRTUALALLOC2
 #      define PLATFORM_HAS_WAITONADDRESS
+#      pragma comment(lib, "mincore.lib")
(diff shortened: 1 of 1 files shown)


ports/snmalloc/portfile.cmake
```diff
@@ -0,0 +1,42 @@
...
+
+vcpkg_cmake_install()
+
+vcpkg_cmake_config_fixup(
+    PACKAGE_NAME snmalloc
+    CONFIG_PATH share/snmalloc
+)
+
+file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/debug/include")
+file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/debug/share")
```
mjp41: I don't understand what these lines are doing.

mjp41 (commented):
Is there a way this can be tested in CI?  I just want to add something so this can't be broken in the future?

matajoh:
> Is there a way this can be tested in CI? I just want to add something so this can't be broken in the future?

Great suggestion, I've added CI tests for this which simulate using vcpkg to install the library.

PR #831, opened 2026-03-24
hotfix for cargo publish

SchrodingerZhu:
A hotfix for the blocker that stop us from publishing crate.
SnMalloc cpp portions were not included in the publish mixture.


SchrodingerZhu:
image

@mjp41 we should consider adding the dry run publish in a pipeline so that we don't run into this hotfix situation again

PR #832, opened 2026-03-25
expose c23 free_sized/free_aligned_sized

SchrodingerZhu:
See https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3220.pdf `7.24.3.4` and `7.24.3.5` for newly added free functions.

A peculiar requirement is that:
- the result of an malloc, calloc, or realloc call may not be passed to free_aligned_sized
- the result of an aligned_alloc call may not be passed to free_sized

In SCUDO, this is tracked by attaching tag in alloc metadata. not sure if we want to check this or not.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -373,7 +373,10 @@ jobs:
     # Run the clang-format check and error if it generates a diff
     - name: Run clang-format
       working-directory: ${{github.workspace}}/build
-      run: make clangformat-check
+      run: |
+        set -eo pipefail
+        make clangformat
+        git diff --exit-code
     - name: Run clang-tidy
       run: |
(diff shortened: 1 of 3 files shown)


SchrodingerZhu:
I may have mixed the order of align/size. I will fix that tomorrow.

mjp41 (approved):
LGTM

mjp41:
@schrodingerzhu could you fix the clangformat issue please.

mjp41:
Clangformat is still unhappy

SchrodingerZhu:
@mjp41 I don't have clang-format-15 on my side. Is it possible to generate a patch as before?

mjp41:
If you revert the change from 826 to the yml file then it will do a patch. 

We can change to that

      run: | 
         set -eo pipefail 
         make clangformat 
         git diff --exit-code 
       run: make clangformat-check

mjp41:
I think we should change it back to make it easier. ( On my phone, so can't actually do that currently)

mjp41 (approved):

PR #835, opened 2026-03-28
Test build perf v2

mjp41:


SchrodingerZhu:
We can also try modules if speed is a concern.

mjp41:
> We can also try modules if speed is a concern.

Mostly there is a lot of redundant work in the build, and I wanted to experiment with improving that.

mjp41:
@schrodingerzhu there is a lot of changes here.  The individual commits are all pretty self-contained.

If you don't have time, just let me know, and I'll take a second pass on the commits myself.

src/snmalloc/mem/corealloc.h
```diff
@@ -1400,7 +1422,7 @@ namespace snmalloc
 
       auto& key = freelist::Object::key_root;
 
-      for (size_t i = 0; i < NUM_SMALL_SIZECLASSES; i++)
+      for (smallsizeclass_t i; i < NUM_SMALL_SIZECLASSES; i++)
```
SchrodingerZhu: this looks a bit strange. maybe just use `smallsizeclass_t i{}` for explicit default initialization.

SchrodingerZhu:
LGTM in general, the change duplicates the symbols on slow paths:
```
schrodingerzy@promaxgb10-edf0:~/Documents/snmalloc/build$ llvm-nm-21 libsnmallocshim.so --demangle  | grep __malloc_end_pointer
000000000001e9d8 t snmalloc::libc::__malloc_end_pointer(void*)
000000000001e9d8 T __malloc_end_pointer
```

Is this expected?

mjp41:
> LGTM in general, the change duplicates the symbols on slow paths:
> 
> ```
> schrodingerzy@promaxgb10-edf0:~/Documents/snmalloc/build$ llvm-nm-21 libsnmallocshim.so --demangle  | grep __malloc_end_pointer
> 000000000001e9d8 t snmalloc::libc::__malloc_end_pointer(void*)
> 000000000001e9d8 T __malloc_end_pointer
> ```
> 
> Is this expected?

This is a good question.  I have not been precise on what symbols should be exposed by each library. 

What do you think is the right thing?   Should the libsnmallocshim.so just provide a libc interface or also the snmalloc::libc interface?

SchrodingerZhu:
> What do you think is the right thing? Should the libsnmallocshim.so just provide a libc interface or also the snmalloc::libc interface?

I think it is fine to have the current version. I initially thought the refactor may add a layer of indirection but the compiler aliased the symbol directly.

SchrodingerZhu:
Ccache does not work well with cl. You can try the ccache action with sccache variant. Do not use the sccache action as GHA cache object service is pretty bad in terms of ratelimit.

SchrodingerZhu:
See also https://github.com/llvm/llvm-project/blob/6591086b6ed41cfac31fd33b3c3e34708763a352/.github/workflows/libc-overlay-tests.yml#L60

Line 87 may also be useful:

```
    # Use MinSizeRel to reduce the size of the build.
    # Notice that CMP0141=NEW and MSVC_DEBUG_INFORMATION_FORMAT=Embedded are required
    # by the sccache tool
```

mjp41:
Made it use sccache, but that doesn't get picked up by the MSBuild builds.  I would like to test msbuild as that should be a supported way to build snmalloc.  So I think this is done.

I think the next step to improving CI time, would be for tests to have a `--smoke-test` or something setting, so that they run the perf tests for shorter durations.

I also wonder if some of the longer running tests are providing value.

mjp41:
@schrodingerzhu are you happy with this level of caching?

SchrodingerZhu (approved):
LGTM.

PR #850, opened 2026-05-22
WIP Hcarter/port to cheri linux

harrisoncarter-capablehub:
This also contains MORE patches to get workflow to behave. I'll figure it out later.

### DIFF
--- a/src/snmalloc/pal/pal_linux.h
+++ b/src/snmalloc/pal/pal_linux.h
@@ -19,6 +19,13 @@
 #    include <linux/futex.h>
 #  endif
 
+#  if defined(__CHERI_PURE_CAPABILITY__) //&& defined(__MORELLO__)
+#    include <sys/cheri.h>
+#    if !defined(CHERI_PERM_SW_VMEM)
+#      define CHERI_PERM_SW_VMEM CHERI_PERM_CHERIABI_VMMAP
+#    endif
+#  endif
+
 #  include <stdio.h>
 
 namespace snmalloc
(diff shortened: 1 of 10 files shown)


src/snmalloc/pal/pal_linux.h
```diff
@@ -161,6 +168,33 @@ namespace snmalloc
...
+    /**
+     * On CheriBSD, exporting a pointer means stripping it of the authority to
+     * manage the address space it references by clearing the SW_VMEM
+     * permission bit (Which sounds appropriate on Linux, too).
+     */
+    template<typename T, SNMALLOC_CONCEPT(capptr::IsBound) B>
+    static SNMALLOC_FAST_PATH CapPtr<T, capptr::user_address_control_type<B>>
+    capptr_to_user_address_control(CapPtr<T, B> p)
+    {
+      if constexpr (Aal::aal_cheri_features & Aal::AndPermsTrapsUntagged)
```
davidchisnall, line 185: I think this is now dead.  The corresponding bit in the AAL looks like it's based on a very old RISC-V spec where we were trapping on some capability manipulation instructions.  I think these two flags can probably both be removed now.

harrisoncarter-capablehub: Sorry I created this PR. I forgot to change what I was merging _into_. I was hoping it would just sorta vanish into the aether.

That is good to know, actually, though. I've not properly tested it, yet. The rest of the PR is/will be getting it to use our workflow so I've been focussing on that. The main thing I've done is just get it to build in yocto. Thanks, though. At least I'll know where to look when it doesn't malloc.

PR #855, opened 2026-06-09
New Arena Range

mjp41:
This PR replaces them with a new backend that handles snmalloc's full
(exponent, mantissa) size class sequence end-to-end, and still maintains
the snmalloc invariant that any allocation is aligned by the largest power
of two that divides its size.

Free blocks are kept in a set of bins, with a few extra bins to handle the
alignment subtleties: a 5-unit block at the wrong address cannot serve a
4-unit request (because of the higher alignment), whereas a 6-unit block
can serve every smaller size. A small precomputed mask per request hides
exactly those bins that cannot serve it.

Consolidation is Doug Leaâstyle: on free we look left and right and merge
maximally. Blocks live in two layers of red-black trees â one tree per
non-empty bin (for selection within a bin) and one address-keyed tree over
all free blocks in the allocator (for left/right neighbour lookup on
free). Because the trees scale, the same backend stacks at multiple levels
of the range pipeline, the way buddy allocators used to.


mjp41:
So running this on CCF (thanks @achamayou), this seems to show a few consistent wins on some benchmarks, and non-of the benchmarks consistently lose.  So this seems like a good perf win for CCF.

On mimalloc bench there appear to be a couple of regressions, but nothing particularly bad. And the issues appear to be confined to the least interesting micro-benchmarks.

mjp41:
@nwf this might be of interest to you.  It contains the alignment based bucketing we discussed.

mjp41:
I think this PR should move the version number to 0.8.0, it is a substantial change to the backend of snmalloc.

SchrodingerZhu:
I will take a look. I should be able to resume my open source work in this repo now.

docs/Arena.md
```diff
@@ -0,0 +1,152 @@
...
+A buddy allocator only stores power-of-two blocks. A request for 5 chunks
+must be served from an 8-chunk buddy block, wasting 3 chunks. We wanted a
+range that
+
+* stores blocks at their actual size,
+* uses snmalloc's full `(exponent, mantissa)` size-class sequence at the
+  range level, and
+* still answers "find a block that can serve this request" in O(1).
+
+## The core idea: search upward, mask out exceptions
```
SchrodingerZhu, line 22: This sounds a lot like the TLSF idea. See https://github.com/llvm/llvm-project/blob/9e3a7335e77e92cf4d739d78592c6c31900c4e97/libc/src/__support/freestore.h#L240

docs/Arena.md
```diff
@@ -0,0 +1,152 @@
...
+Most requests need no exceptions at all. Only size classes whose alignment
+exceeds the expected alignment for their position in the sequence have any
+bits to mask. The whole structure is constant-folded into a few small tables.
+
+## The two-tree structure
+
+A bitmap alone is not enough — when a bin is non-empty, the arena still has
+to *retrieve* and *coalesce* blocks. Each `Arena` therefore maintains:
+
+* **One red-black tree per non-empty bin** (the "bin trees"), keyed by
```
SchrodingerZhu, line 76: why is the bin tree indexed by address? if we generally request oversized bin, we can potentially do O(1) popping. The range tree then maps the address to the doubly linked list node. I am writing down this while reading the doc first, so I may change my mind when reviewing the code :D.

mjp41: We have used a strategy of trying to allocate low address's in preference.  I am sure I read somewhere that focussing on an area can help with fragmentation. So it focusses on smallest x lowest as a lex product effectively.

docs/Arena.md
mjp41: It is very much like TLSF, but it handles alignment too.  That is something that the TLSF algorithm doesn't handle.  I think it over-allocates to ensure alignment.

In fact my starting point was can I do a TLSF/DL like backend to allow for more intermediate sizes, and still respect all the alignment guarantees I wanted.

docs/Arena.md
SchrodingerZhu: I see.

PR #856, opened 2026-06-10
Move test_memory_stats to an integration test

mjp41:
test_memory_stats reads a process-global counter that every allocation in the process feeds into. Placing it under tests/ runs it in its own binary, isolating its snapshots from any other test thread.


mjp41:
FYI @jayakasadev

PR #858, opened 2026-06-17
Fix MSVC C4864 warning and testlib message_impl link error

matajoh:
corealloc.h: `small_refill` and `small_alloc` are member templates invoked through an `Allocator*` of dependent type. MSVC's conformance mode requires the `template` disambiguator, otherwise it emits C4864 ("expected 'template' keyword before dependent template name"). Add the keyword to the three affected call sites.

snmalloc_testlib.cc: `message_impl` is declared non-inline in defines.h but defined inline in pal.h. On MSVC its COMDAT was never emitted into the test library (nothing referenced it), so testlib-only tests that see only the declaration (redblack, seqset) failed to link with LNK2019. Force-emit the COMDAT via a non-elidable odr-use, keeping the symbol inline everywhere so it stays ODR-safe with the copies other tests compile from snmalloc.h.

### DIFF
--- a/src/snmalloc/mem/corealloc.h
+++ b/src/snmalloc/mem/corealloc.h
@@ -656,7 +656,8 @@ namespace snmalloc
           smallsizeclass_t sizeclass,
           freelist::Iter<>* fl,
           size_t size) SNMALLOC_FAST_PATH_LAMBDA {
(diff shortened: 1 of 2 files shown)


mjp41 (approved):
Lgtm

PR #859, opened 2026-06-20
Implement tick() for RISC-V

Xeonacid:
RISC-V builds currently fail when code instantiates Aal::tick() if the compiler does not provide __builtin_readcyclecounter(). The generic AAL falls back to Arch::tick(), but AAL_RISCV has no implementation.


mjp41 (approved):
LGTM, would you be able to address the clangformat issue please. Thanks

Xeonacid:
> LGTM, would you be able to address the clangformat issue please. Thanks

Oops, I didn't notice that. Updated. Thanks for the review!

PR #863, opened 2026-06-26
reallocarr overflow check ordering fix.

devnexen:
check the overflow flag before the zero-size early return, otherwise an overflow truncating to zero is wrongly reported as success.


mjp41 (approved):
LGTM. Good catch.

PR #864, opened 2026-07-10
build: bump rules_rust to 0.71.3, mark crate_universe extension dev-only

jayakasadev:
## What
Bumps `rules_rust` 0.70.0 -> 0.71.3 and marks the `crate_universe` extension (`flate2`/`backtrace` for the profiling/symbolicate features) `dev_dependency = True`.

## Why
`rules_rust` 0.71.0 removed the deprecated `crate_universe/extension.bzl` alias this repo's `MODULE.bazel` still loaded, and 0.71.1 additionally requires every non-root module's `crate.from_specs()`/`from_cargo()` call to ship an explicit lockfile. Neither applied while this repo sat at 0.70.0, but both blocked the capitalintent monorepo (a downstream `bazel_dep` consumer) from bumping its own `rules_rust` past 0.70.0.

`dev_dependency = True` is the actual fix, not a workaround: bzlmod skips dev-only extensions entirely when a module is a non-root dependency, so a downstream consumer's `bazel mod deps` never evaluates `crate.from_specs()` here and never needs a lockfile for it. `rust.toolchain()` was already `dev_dependency = True` and never hit either issue, which is what pointed at this fix.

## Cost
Downstream consumers can no longer resolve `@crates`/`@snmalloc_crates` directly, so `:snmalloc_rs_profiling` / `:snmalloc_rs_profiling_symbolicated` aren't buildable outside this repo's own tests/CI. `:snmalloc_rs_profile_compat` (already existed in `snmalloc-rs/BUILD.bazel`) is the documented escape hatch — it gets the `SNMALLOC_PROFILE=ON` C archive without the flate2/backtrace Rust deps. The `MODULE.bazel` comment block had drifted out of sync with this tradeoff, which `snmalloc-rs/BUILD.bazel` already documented correctly; updated it to match.

## Test plan
- [x] `bazel mod deps --lockfile_mode=update` resolves clean
- [x] `bazel build //snmalloc-rs/...` (excluding the pre-existing, unrelated `-mcx16`-on-arm64 fuzzer failure) — all 16 targets build, including the profiling/symbolicate variants
- [x] `bazel test //snmalloc-rs/...` — 10/10 tests pass


jayakasadev:
Opened against the wrong repo by mistake (meant to target my fork, jayakasadev/snmalloc) — closing. Apologies for the noise.

PR #865, opened 2026-07-25
The constructor left it indeterminate, so a non-static timer is linked

devnexen:
into PalTimer's list with a garbage next pointer that apply_all then dereferences. PalNotificationObject already initialises its equivalent.

### DIFF
--- a/src/snmalloc/pal/pal_ds.h
+++ b/src/snmalloc/pal/pal_ds.h
@@ -103,7 +103,7 @@ namespace snmalloc
     template<typename T>
     friend class PalList;
 
-    stl::Atomic<PalTimerObject*> pal_next;
+    stl::Atomic<PalTimerObject*> pal_next{nullptr};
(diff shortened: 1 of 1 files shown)


src/snmalloc/pal/pal_ds.h
```diff
@@ -103,7 +103,7 @@ namespace snmalloc
     template<typename T>
     friend class PalList;
 
-    stl::Atomic<PalTimerObject*> pal_next;
+    stl::Atomic<PalTimerObject*> pal_next = nullptr;
```
mjp41: I have a slight preference for:
```suggestion
    stl::Atomic<PalTimerObject*> pal_next{nullptr};
```

I found sometimes the `= nullptr` can prevent `static constexpr` working properly.

mjp41 (approved):
LGTM, minor comment

PR #878, opened 2026-09-07
Fix the header-only integration recipe in BUILDING.md

lenamonj:
The header-only section of docs/BUILDING.md links `snmalloc_lib`, a target that left CMakeLists.txt in the CMake cleanup (#384), and includes `src/snmalloc/override/malloc.cc`, a path that resolves neither relative to the caller's file nor on any include path. A project following the recipe verbatim does not build.

This links the `snmalloc` interface target instead, which puts `src` on the include path, and includes the two override files through that path.

Verified on a fresh clone of main by extracting the fenced blocks of the section into a consumer project with `[lib_name]` replaced by a target name: before the change the build fails at the first include, after it the binary builds, defines `malloc_good_size` and exits 0. Documentation only.


mjp41 (approved):
Thanks for fixing.

